PR #1, opened 2016-04-09
[JENKINS-26132] Created FlowNodeSerialWalker

jglick:
Downstream of https://github.com/jenkinsci/workflow-cps-plugin/pull/5.

@reviewbybees esp. @svanoort with whom I have discussed this concept.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNodeSerialWalker.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNodeSerialWalker.java
@@ -27,49 +27,99 @@ package org.jenkinsci.plugins.workflow.graph;
 import java.util.Iterator;
 import java.util.List;
 import java.util.NoSuchElementException;
+import javax.annotation.CheckForNull;
+import org.jenkinsci.plugins.workflow.actions.LabelAction;
+import org.jenkinsci.plugins.workflow.actions.StageAction;
 
 /**
  * Unlike {@link FlowGraphWalker}, this iterator does not traverse the entire graph, but only walks backward and outward from one node.
  * If the start node is inside some nested blocks, it will step up through the {@link BlockStartNode}s.
  * But when moving backwards in time, when encountering blocks that the starting node was not inside,
- * it will visit the {@link BlockEndNode} and then the {@link BlockStartNode} without visiting any of the internal nodes.
- * The {@link FlowStartNode} will be last.
+ * it will visit the {@link BlockStartNode} without visiting any of the internal nodes or {@link BlockEndNode}.
+ * The {@link FlowStartNode} will not be visited.
  */
-public class FlowNodeSerialWalker implements Iterable<FlowNode> {
+public final class FlowNodeSerialWalker implements Iterable<FlowNode> {
 
     private final FlowNode start;
 
     /**
      * Initializes the walker.
-     * @param start the node which will be the first to be returned in the walk
+     * @param start the node which will be the first to be returned in the walk (unless it was a {@link BlockEndNode}, in which case the corresponding {@link BlockStartNode} will be first)
      */
     public FlowNodeSerialWalker(FlowNode start) {
-        this.start = start;
+        this.start = start instanceof BlockEndNode ? ((BlockEndNode) start).getStartNode() : start;
     }
 
-    @Override public Iterator<FlowNode> iterator() {
-        return new Iterator<FlowNode>() {
-            FlowNode n = start;
-            @Override public boolean hasNext() {
-                return !n.getParents().isEmpty();
+    @Override public EnhancedIterator iterator() {
+        return new EnhancedIterator();
+    }
+    
+    /**
+     * Special iterator offering additional functionality.
+     */
+    public final class EnhancedIterator implements Iterator<FlowNode> {
+
+        private FlowNode next;
+        private FlowNode curr;
+        boolean ancestor;
+
+        EnhancedIterator() {
+            next = start;
+        }
+
+        @Override public boolean hasNext() {
+            return !next.getParents().isEmpty();
+        }
+
+        @Override public FlowNode next() {
+            if (next instanceof BlockEndNode) {
+                next = ((BlockEndNode) next).getStartNode();
+                ancestor = false;
+            } else {
+                ancestor = next == start || next instanceof BlockStartNode;
             }
-            @Override public FlowNode next() {
-                FlowNode prev = n;
-                if (n instanceof BlockEndNode) {
-                    n = ((BlockEndNode) n).getStartNode();
-                } else {
-                    List<FlowNode> parents = n.getParents();
-                    if (parents.size() != 1) {
-                        throw new NoSuchElementException("unexpected " + n + " with parents " + parents);
-                    }
-                    n = parents.get(0);
+            curr = next;
+            List<FlowNode> parents = next.getParents();
+            if (parents.size() != 1) {
+                throw new NoSuchElementException("unexpected " + next + " with parents " + parents);
+            }
+            next = parents.get(0);
+            return curr;
+        }
+        
+        @Override public void remove() {
+            throw new UnsupportedOperationException();
+        }
+
+        /**
+         * Checks whether the current node is an ancestor of the start node.
+         * May only be called after {@link #next}.
+         * @return true if the current node is the start node, or is a {@link BlockStartNode} which enclosed the start node
+         */
+        public boolean isAncestor() {
+            return ancestor;
+        }
+
+        /**
+         * Checks if there is a display label shown on the current node which should apply to the start node.
+         * Applicable labels include {@link LabelAction#getDisplayName} if {@link #isAncestor}, or {@link StageAction#getStageName} if encountered anywhere.
+         * May only be called after {@link #next}.
+         * @return a label, or null if inapplicable
+         */
+        public @CheckForNull String currentLabel() {
+            if (ancestor) {
+                LabelAction a = curr.getAction(LabelAction.class);
+                if (a != null) {
+                    return a.getDisplayName();
                 }
-                return prev;
             }
-            @Override public void remove() {
-                throw new UnsupportedOperationException();
+            StageAction a = curr.getAction(StageAction.class);
+            if (a != null) {
+                return a.getStageName();
             }
-        };
+            return null;
+        }
+
     }
 
 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNodeSerialWalker.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNodeSerialWalker.java
@@ -65,6 +65,9 @@ public class FlowNodeSerialWalker implements Iterable<FlowNode> {
                 }
                 return prev;
             }
+            @Override public void remove() {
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNodeSerialWalker.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.List;
+import java.util.NoSuchElementException;
+import javax.annotation.CheckForNull;
+import org.jenkinsci.plugins.workflow.actions.LabelAction;
+import org.jenkinsci.plugins.workflow.actions.StageAction;
+
+/**
+ * Unlike {@link FlowGraphWalker}, this iterator does not traverse the entire graph, but only walks backward and outward from one node.
+ * If the start node is inside some nested blocks, it will step up through the {@link BlockStartNode}s.
+ * But when moving backwards in time, when encountering blocks that the starting node was not inside,
```
amuniz: What's "move backwards _in time_", is that possible?

jglick: Well for

```groovy
stage 'dev' // 1
node { // 2
  sh 'echo hello' // 3
}
```

if 3 is the start node, we would visit 3, then 2, then 1, where 2 is moving up and 1 is moving back.

Evidently needs to be documented more clearly.

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeSerialWalkerTest.java
```diff
@@ -0,0 +1,114 @@
...
+            "lE 'end'", true));
+        WorkflowRun b = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
+        r.assertLogContains("start: null", b);
+        r.assertLogContains("in-one: Branch: one", b);
+        r.assertLogContains("in-two: Branch: two", b);
+        r.assertLogContains("middle: null", b);
+        r.assertLogContains("in-dev: dev", b);
+        r.assertLogContains("in-test: test", b);
+        r.assertLogContains("in-three: Branch: three", b);
+        r.assertLogContains("after-load: test", b);
```
amuniz: Could you point me to the code preventing this to be `after-load: f1` please? Is not `load 'f1'` parent of `lE 'after-load'`?

amuniz:
:bee: (but I'm not sure I understand how that iterator traverses the flow graph)

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeSerialWalkerTest.java
jglick: Yes it is a parent according to the graph, but it does not enclose the `lE` step. Specifically `currentLabel()` will be null on that iteration since `isAncestor()` will be false: the iterator would first have hit the `BlockEndNode` of the `load` step, omitting it in favor of the corresponding `BlockStartNode` (`StepStartNode` to be precise), and setting `ancestor = false`.

svanoort:
@jglick Yes, this is fairly intuitive for someone familiar with the flow graph (going backwards in time, and jumping from the end of an irrelevant block to its start).   Clever in how it is exposed as an iterator.   Overlaps with what's planned for the new FlowAnalyzer, but more limited in scope.  

I see a couple problems at the architectural/algorithmic level that would better be addressed with a different approach:

1. We're offering one solution to one problem where we really need a general API now, because we have many similar cases.  This is why I am working towards a more generalized API based on a SAX-like visitor pattern with the new FlowAnalyzer in [JENS-34038](https://issues.jenkins-ci.org/browse/JENKINS-34038).  I think the better approach is to finish that and tie it into this, as this is a subset of the problem. 
2. The iterator approach here is clever, why not let the walker do the filtering rather than requiring every consumer to do it?  Rather than exposing an iterator, we can expose something like:  `public FlowNode searchFlowForFirst(FlowNode startNode, FlowNodePredicate searchCondition, FlowNode searchType)` where the FlowNodePredicate returns true when it has found a node matching the condition.  Here searchType would be something like FlowSearch.BLOCKS (as opposed to FlowSearch.PARENTS, or FlowSearch.ALLNODES).  We could also offer a search API that exposes and iterator too for cases where we want it, but to pass in a Visitor object that is called internally upon seeing a node meeting the predicate.
3. Limited to operating on one FlowNode at a time.  Often we will be scanning a whole FlowExecution for all labelled blocks or nodes fitting a specific criteria; it is both simpler and more efficient to iterate through all nodes just once in those cases (see the SAX-like approach above). We can track the current block hierarchy in a Stack, pushing and popping to this as we enter or exit blocks. 

I feel that this is :bug:  for these reasons.  Not because the algorithm is wrong (it looks correct, though I haven't checked all the edge cases), but because we can do this even better, and I already am well on the way to doing so!  (and would like to roll this in as a subset of that).

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNodeSerialWalker.java
amuniz: Yeah, actually adding that example to the javadoc would be great, or even a more complex case and how the iterator order would be.

jglick:
Both points 1 and 3 deal with the fact that there are other kinds of clients with somewhat overlapping needs. That is true, but I would like to differentiate between those which are rendering UI (possibly rapidly and repeatedly), vs. those which are performing a one-time calculation as part of the build itself. The currently filed downstream https://github.com/jenkinsci/workflow-durable-task-step-plugin/issues/2 falls into the first class; the fix of [JENKINS-28119](https://issues.jenkins-ci.org/browse/JENKINS-28119) which I am working on (can file a werk-in-progress soon) falls into the second class. For the second class, we do not want any caching (since the result must be precise), and really any optimization is wasted.

The proposed `searchFlowForFirst` would not work for https://github.com/jenkinsci/workflow-durable-task-step-plugin/issues/2 I think. That code is not finding a single node matching a predicate.

svanoort:
@jglick 

> Both points 1 and 3 deal with the fact that there are other kinds of clients with somewhat overlapping needs.

Very true, but better if we can solve many challenges at once.

> I would like to differentiate between those which are rendering UI (possibly rapidly and repeatedly), vs. those which are performing a one-time calculation as part of the build itself

It is definitely true there are a couple kinds of analysis, I think even better than UI vs. while-building, it is better to split the use cases like so:

* Scan for information about one node (often a head)
* Scan a whole run for statistics
* Scan for information about the current block or stage only

From the analyzer/walker perspective, there is no difference between a result that gets cached for reuse and one that does not.  That is best handled by the function that wants the information, which is best informed about how to handle it (and when it is appropriate to reuse results).

Really the only difference between JENKINS-28119 and  jenkinsci/workflow-durable-task-step-plugin#2 is in how we scan (parallel-aware vs. DFS with block skipping) and what we look for (ErrorAction vs. instanceof ExecutorStepNode). 

I've got some comments about how to use this in conjunction with caching/memoization, and how to build upon performance gains from block-skipping, but on reflection they better belong in the consuming PRs.

svanoort:
> The proposed searchFlowForFirst would not work for jenkinsci/workflow-durable-task-step-plugin#2 I think. That code is not finding a single node matching a predicate.

Well, you could do two operations, the first searches for the first node starting an enclosing executor or label block (returning null if not found), then if one is found, the appropriate value is store.  Then the second search starts from that block and looks for the other block type, returns the node, and uses its value.

Technically it works, but that's terribly clumsy.

A better way is to sketch the stack of enclosing blocks, using a visitor object that internally stores the state we're tracking inside the iterator here (and which ends analysis when done, by returning false on one of the handler method calls).

jglick:
> A better way is to sketch the stack of enclosing blocks, using a visitor object…

How is that an improvement over the straightforward iterator idiom being used in the downstream PR?

jglick:
Unproven sketch of alternate API, for my records:

```java
public class FlowMetadataCache<T> {
    
    private final FlowExecution execution;
    private final Analyzer<T> analyzer;
    private final Cache<String,T> cache = CacheBuilder.newBuilder().build();

    private FlowMetadataCache(FlowExecution execution, Analyzer<T> analyzer) {
        this.execution = execution;
        this.analyzer = analyzer;
    }

    public T get(FlowNode n) {
        return null; // TODO
    }

    public static <T> FlowMetadataCache<T> of(FlowExecution execution, Analyzer<T> analyzer) {
        return new FlowMetadataCache<T>(execution, analyzer); // TODO cache them
    }

    public static Analyzer<String> ENCLOSING_LABEL = new Analyzer<String>() {
        @Override protected String analyze(FlowNode node) {
            LabelAction la = node.getAction(LabelAction.class);
            if (la != null) {
                return la.getDisplayName();
            }
            FlowNode n = findEnclosingNode(node);
            if (n != null) {
                return get(n);
            } else {
                return STAGE.get(node);
            }
        }
    };

    public static Analyzer<String> STAGE = new Analyzer<String>() {
        @Override protected String analyze(FlowNode node) {
            StageAction sa = node.getAction(StageAction.class);
            if (sa != null) {
                return sa.getStageName();
            }
            FlowNode n = findPrecedingNode(node);
            if (n != null) {
                return get(n);
            } else {
                return null;
            }
        }
    };

    public static abstract class Analyzer<T> {

        protected abstract T analyze(FlowNode node);

        /**
         * Load cached metadata from another node so as to build an incremental cache.
         */
        // TODO prone to stack overflows, redesign
        protected final T get(FlowNode node) {
            return null; // TOOD
        }

        /**
         * Find the node enclosing this one.
         */
        protected final @Nonnull FlowNode findEnclosingNode(FlowNode node) {
            return null; // TODO
        }

        /**
         * Find the node preceding this one in series, or enclosing this one.
         */
        protected final @CheckForNull FlowNode findPrecedingNode(FlowNode node) {
            return null; // TODO
        }

    }

}
```

svanoort:
Worth mentioning that we have a basic incremental caching implementation tying to the newer API set here:
https://github.com/jenkinsci/workflow-api-plugin/pull/2/files#diff-adbf1cb9136f7bb5edff0dcc77372b0cR19

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeSerialWalkerTest.java
```diff
@@ -0,0 +1,114 @@
...
+public class FlowNodeSerialWalkerTest {
+
+    @ClassRule public static BuildWatcher buildWatcher = new BuildWatcher();
+    @Rule public JenkinsRule r = new JenkinsRule();
+
+    @Test public void labelEnclosing() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition(
+            "lE 'start'; parallel one: {lE 'in-one'}, two: {lE 'in-two'}; lE 'middle'\n" +
+            "stage 'dev'; lE 'in-dev'; stage 'test'; lE 'in-test'\n" +
```
jglick: @svanoort noted that this is failing to test the case which will not work, when you have

```groovy
node {
  stage 'Dev'
  // …
}
echo 'search from here'
```

jglick:
Expected to be replaced by something TBD in #2 + #10; will have to see how to adapt downstream PRs.

PR #2, opened 2016-04-14
Generified Flow Search Algorithms

svanoort:
Provides a bundle of search and iteration APIs to simplify working with pipeline flow graphs (DAGs). 

These are intended to be extensible for custom use cases, but provide a general API that covers what is needed. 

For a rundown of the core APIs and how to work with them, see: https://github.com/jenkinsci/workflow-api-plugin/pull/2/files#diff-f8537b8acb35b350992eb264d4f9f207R43

In gist, we're offering some core primitives for working with the flow graph that are needed to do more involved work with the flow graph:

* The visitor pattern is the starting point for SAX-like analysis of a flowgraph (and may be extended to provide a block-aware or marker-aware, i.e. stage/milestone-aware visitor)
* ForkScanner is needed to effectively handle blocks in the flowgraph (since blocks are seen in a logical linear order).  It is the primitive needed for block-based use in a DOM-like tree analysis, or block-aware SAX-like analysis
* LinearBlockHoppingFlowScanner is the primitive that allows for quickly ignoring sibling blocks (helpful if you only care about preceding or enclosing nodes, similar to the FlowNodeSerialWalker). 

Notes:

* Parallel blocks have their own LabelAction for each branch, a ParallelLabelAction on the StepStartNode beginning each branch, with displayname branch + displayname, and getThreadName that returns the label given to the branch, StepEndNode gives the timing too (and final state)

General use cases:
* ErrorInfo - https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/2 = https://issues.jenkins-ci.org/browse/JENKINS-28119
   - Searching back for node emitting failure (ErrorAction), skips block end node because the blockendnode gets the final error action.
  - Then does the walk back via blockjumping to get the nearest log action, and the nearest label too
* [`WorkflowRun.getLogPrefix`](https://github.com/jenkinsci/workflow-job-plugin/blob/master/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L402-L426).
* [Milestone step - JENKINS-27039](https://github.com/jenkinsci/pipeline-milestone-step-plugin/pull/1)
  - Needs something like FlowGraphWalker but that bombs out on parallel -> optimal case for recursive breadth-first algorithm (planned but not implemented yet).  Caching not allowable probably.
* Durable task step: search label/stage which uses a executor node.
   - Good use case for caching or incremental caching. 
* Log search: nearest enclose start node with label for parallel

Pipeline Stage View:
[...]

PR #3, opened 2016-04-26
Incremental flow analysis

svanoort:
Splits out the incremental flow analysis/caching from #2 so that it can be reviewed & worked on separately.  Depends on #2 though. 

TODO:
- [x] Basic incremental flow analysis
- [ ] Pluggable flowscanner implementation
- [ ] Basic test for incremental flow analysis
- [ ] Test for incremental flow analysis with parallel flows that may be in progress


svanoort:
Closing this one out because it doesn't seem like this analysis on in-progress pipelines is the bottleneck it used to be now (thanks to internal improvements done)

PR #5, opened 2016-06-09
[JENKINS-26130] Way to print progress from pickles

jglick:
[JENKINS-26130](https://issues.jenkins-ci.org/browse/JENKINS-26130)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
@@ -183,9 +183,9 @@ public class FlowExecutionList implements Iterable<FlowExecution> {
                     @Override
                     public void onFailure(Throwable t) {
                         if (t instanceof CancellationException) {
-                            LOGGER.log(Level.FINE, "cancelled load of " + e, t);
+                            LOGGER.log(Level.FINE, "Cancelled load of " + e, t);
                         } else {
-                            LOGGER.log(WARNING, "failed to load " + e, t);
+                            LOGGER.log(WARNING, "Failed to load " + e, t);
                         }
                     }
                 });
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
@@ -19,9 +19,11 @@ import org.jenkinsci.plugins.workflow.steps.StepExecutionIterator;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.channels.CancelledKeyException;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
@@ -167,7 +169,7 @@ public class FlowExecutionList implements Iterable<FlowExecution> {
 
         @Override
         public void onLoaded() {
-            for (FlowExecution e : list) {
+            for (final FlowExecution e : list) {
                 LOGGER.log(FINE, "Eager loading {0}", e);
                 Futures.addCallback(e.getCurrentExecutions(false), new FutureCallback<List<StepExecution>>() {
                     @Override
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -180,7 +182,11 @@ public void onSuccess(List<StepExecution> result) {
 
                     @Override
                     public void onFailure(Throwable t) {
-                        LOGGER.log(WARNING, null, t);
+                        if (t instanceof CancellationException) {
+                            LOGGER.log(Level.FINE, "cancelled load of " + e, t);
```
oleg-nenashev: NIT: start entries from the uppercase letter?

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionOwner.java
```diff
@@ -109,6 +111,14 @@ public String getUrlOfExecution() throws IOException {
     public abstract int hashCode();
 
     /**
+     * Gets a listener to which we may print general messages.
+     * Normally {@link StepContext#get} should be used, but in some cases there is no associated step.
+     */
```
oleg-nenashev: <code>@since</code>
Maybe also getTaskListener()

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionOwner.java
```diff
@@ -122,4 +132,33 @@ public String getUrlOfExecution() throws IOException {
 
     }
 
+    /**
+     * A placeholder implementation for use in compatibility stubs.
+     */
+    public static FlowExecutionOwner dummyOwner() {
```
oleg-nenashev: <code>@since</code>.

I would also use DummyFlowExecutionOwner#getInstance()

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionOwner.java
```diff
@@ -122,4 +132,33 @@ public String getUrlOfExecution() throws IOException {
...
+        return new DummyOwner();
+    }
+
+    private static class DummyOwner extends FlowExecutionOwner {
```
oleg-nenashev: NIT, maybe just DefaultImpl

oleg-nenashev:
🐝 for the implementation, 🐜 for annotations and naming. Mostly NIT

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
jglick: Not sure if there is any particular convention.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionOwner.java
jglick: Generally have not bothered with `@since` on these plugin APIs. If you are already depending on the version that has it, there is little reason to care when it was introduced. If you are not, then you will not know it exists, unless you are browsing sources on GitHub or something. When in doubt, use the latest version.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionOwner.java
jglick: Well I wanted to be clear that it is not a “default”, it is really a stub.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionOwner.java
jglick: > I would also use DummyFlowExecutionOwner#getInstance()

Could; just more verbose for no gain.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionOwner.java
oleg-nenashev: ok

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
oleg-nenashev: But maybe makes sense to keep in consistent within Pipeline plugin at least

oleg-nenashev:
re-:bee:

abayer:
:bee: as well

PR #6, opened 2016-06-10
Pipeline graph timing & structural analysis APIs

svanoort:
Not ready for review, please respect the WIP label.

Builds on top of https://github.com/jenkinsci/workflow-api-plugin/pull/2 with a set of data structures to split into chunks (runs of nodes with a defined start & end -- they may be the same node though). 

Couples with the https://github.com/jenkinsci/pipeline-graph-analysis-plugin which will then provide status/timing on the chunks. 

Consumers:
* Implement SimpleBlockVisitor to consume SAX-like notifications of blocks & parallels info, and use the ForkScanner.visitSimpleChunks method to invoke it.
* Maybe use a standardized callback-based API that builds on the simple visitors to couple with a storage system for chunks (or maybe even return a full representation of the flow).


svanoort:
Replaced by #10

PR #8, opened 2016-07-11
Adding Intellij files to .gitignore.

abayer:
So that I don't end up doing this as part of another PR in the future, here's a PR for .gitignore now. =)

cc @reviewbybees, esp @jglick


svanoort:
🐝  a billion.

jglick:
:bee:

PR #10, opened 2016-08-12
Block scanning APIs

svanoort:
APIs:

Builds on top of #2 with a set of data structures to split into chunks (runs of nodes with a defined start & end -- they may be the same node though).

Couples with the https://github.com/jenkinsci/pipeline-graph-analysis-plugin which will then provide status/timing on the chunks.

Includes:

* Representations of chunks of flows and interfaces for use (providing a common representation)
* Implement SimpleBlockVisitor invocation API to consume SAX-like notifications of blocks & parallels info, and use the ForkScanner.visitSimpleChunks method to invoke it.

@reviewbybees


pom.xml
```diff
@@ -33,7 +33,7 @@
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-api</artifactId>
-    <version>2.1-SNAPSHOT</version>
+    <version>2.2.blockapis-SNAPSHOT</version>
```
svanoort: Temporary versioning to allow use in stage view/pipeline graph analysis.  This will be changed before release.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -149,11 +151,11 @@ private void save(List<FlowExecutionOwner> copy) {
     private static final Logger LOGGER = Logger.getLogger(FlowExecutionList.class.getName());
 
     public static FlowExecutionList get() {
-        Jenkins j = Jenkins.getInstance();
-        if (j == null) { // might be called during shutdown
-            return new FlowExecutionList();
+        FlowExecutionList l = ExtensionList.lookup(FlowExecutionList.class).get(FlowExecutionList.class);
```
svanoort: Review updates from generic flow analysis PR.

pom.xml
```diff
@@ -33,7 +33,7 @@
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-api</artifactId>
-    <version>2.2-SNAPSHOT</version>
+    <version>2.2.blockapis-SNAPSHOT</version>
```
abayer: FYI, the right way to do this kind of versioning would be `2.2-blockapis-SNAPSHOT` - `2.2.blockapis-SNAPSHOT` will actually be treated as a higher version than `2.2-SNAPSHOT`.

abayer:
:bee: - I'm not entirely up to speed on the code enough to *really* judge what it's doing well enough, but nothing jumps out at me as bad, so hey. =)

svanoort:
Especially a CC for @vivek and @jglick  here

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/BlockChunkFinder.java
```diff
@@ -0,0 +1,30 @@
...
+import org.jenkinsci.plugins.workflow.graph.BlockEndNode;
+import org.jenkinsci.plugins.workflow.graph.BlockStartNode;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+
+/**
+ * Matches start and end of a block.  Any block
+ * @author Sam Van Oort
```
oleg-nenashev: All API methods have no <code>@since</code> :(

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -73,6 +93,21 @@
 
     private boolean walkingFromFinish = false;
 
+    protected NodeType currentType;
+    protected NodeType nextType;
```
oleg-nenashev: 🐛 Uninitialized fields, which are being used in child classes. At least assign null

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -42,27 +45,44 @@
...
  *     <li>Branch information is available for use here</li>
  * </ul>
  *
- * @author <samvanoort@gmail.com>Sam Van Oort</samvanoort@gmail.com>
+ * @author Sam Van Oort
  */
 public class ForkScanner extends AbstractFlowScanner {
 
+    public NodeType getCurrentType() {
+        return currentType;
```
oleg-nenashev: CheckForNull?

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -406,6 +470,13 @@ protected FlowNode hitParallelStart() {
     }
 
     @Override
+    public FlowNode next() {
```
oleg-nenashev: Makes sense to add NotThreadSafe annotation to this class

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -419,16 +490,25 @@ protected FlowNode next(@Nonnull FlowNode current, @Nonnull Collection<FlowNode>
...
                     return temp;
                 }
             } else if (!blackList.contains(p)) {
+                if (p instanceof BlockStartNode && p.getAction(ThreadNameAction.class) != null) {
+                    nextType = NodeType.PARALLEL_BRANCH_START;
+                } else if (ForkScanner.isParallelEnd(p)) {
+                    nextType = NodeType.PARALLEL_END;
+                } else {
+                    nextType = NodeType.NORMAL;
+                }
```
oleg-nenashev: dark magic on multiple ifs.
Would it be possible to replace it by a state machine?

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/MemoryFlowChunk.java
```diff
@@ -0,0 +1,99 @@
...
+
+    public MemoryFlowChunk(@CheckForNull FlowNode before, @Nonnull FlowNode firstNode, @Nonnull FlowNode lastNode, @CheckForNull FlowNode nodeAfter) {
+        this.setNodeBefore(before);
+        this.setFirstNode(firstNode);
+        this.setLastNode(lastNode);
+        this.setNodeAfter(lastNode);
+    }
+
+    public MemoryFlowChunk() {
+
```
oleg-nenashev: It leaves fields uninitialized 🐛

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/SimpleChunkVisitor.java
```diff
@@ -0,0 +1,114 @@
...
+ * </ul>
+ *
+ * <em>Important implementation note: multiple callbacks can be invoked for a single node depending on its type.</em>
+ * <p></p>For example, we may capture parallels as chunks.
+ *
+ * <p></p><h3>Callbacks Reporting on chunk/parallel information:</h3>
+ * <ul>
+ *     <li>{@link #chunkStart(FlowNode, FlowNode, ForkScanner)} is called on the current node when we hit start of a boundary (inclusive) </li>
+ *     <li>{@link #chunkEnd(FlowNode, FlowNode, ForkScanner)} is called when we hit end of a boundary (inclusive)</li>
+ *     <li>{@link #atomNode(FlowNode, FlowNode, FlowNode, ForkScanner)} called when a node is neither start nor end.</li>
```
oleg-nenashev: atomicNode?

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/MemoryFlowChunk.java
svanoort: @oleg-nenashev Java does implicit initialization -- http://docs.oracle.com/javase/specs/jls/se7/html/jls-4.html#jls-4.12.5 -- these are null unless set.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
svanoort: @oleg-nenashev Second time: java does implicit initialization -- http://docs.oracle.com/javase/specs/jls/se7/html/jls-4.html#jls-4.12.5 -- these are null unless set.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/BlockChunkFinder.java
svanoort: True.   Once I know when they're actually going to get merged I can add one 😞

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/SimpleChunkVisitor.java
svanoort: @oleg-nenashev Hey now, leave my past work in nuclear physics out of it, I promise I don't glow green at night. ;) 

In all seriousness, AtomNode is a convention in pipeline itself (see: StepAtomNode).  This is following the standards there.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/MemoryFlowChunk.java
oleg-nenashev: I know. But you will be receiving legitimate warnings from static analyzers

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
svanoort: Yes, good point (probably for all of them, and worth a mention in AbstractFlowScanner Javadocs).   Waiting for the review batch to be in to batch up changes though.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
svanoort: It's a good point: it is a sort of state machine, but the transition rules are rather complex (of necessity, unfortunately).   The currentType/nextType *might* be usable to feed some of the internal logic -- it might simplify things, it might not.  

Currently the discovery of a parallelStartNode for the type fields is a bit hack though, so it's not as trustworthy yet for that purpose. 

May be worth revisiting this in the future if ForkScanner needs to be touched again after this.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/MemoryFlowChunk.java
svanoort: Fair.  I don't think it's worth making a bug, but I'll change any anyway since it's trivial to do.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/BlockChunkFinder.java
svanoort: Looks like per @jglick convention in this plugin is not to use @since so I'm off the hook... this time.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/MemoryFlowChunk.java
oleg-nenashev: ty

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
oleg-nenashev: OK, agreed

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/SimpleChunkVisitor.java
oleg-nenashev: OK

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowChunkWithContext.java
```diff
@@ -0,0 +1,17 @@
+package org.jenkinsci.plugins.workflow.graphanalysis;
+
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+
+import javax.annotation.CheckForNull;
+
+/** FlowChunk with information about what comes before/after */
+public interface FlowChunkWithContext extends FlowChunk {
+
+    /** Return the node before this chunk, or null if it is the end */
```
oleg-nenashev: Copy-pasta? I'd expect "the beginning"

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -42,27 +46,47 @@
...
+    }
+
+    @CheckForNull
+    public NodeType getNextType() {
+        return nextType;
+    }
+
+    /** Used to recognize special nodes */
+    public enum NodeType {
+        NORMAL,
```
oleg-nenashev: TYPE_WE_DO_NOT_CARE_ABOUT would fit the context better, but oh well

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -73,6 +97,21 @@
 
     private boolean walkingFromFinish = false;
 
+    NodeType currentType = null;
+    NodeType nextType = null;
+
+    public ForkScanner() {
+
```
oleg-nenashev: no setup call at all? I would just invoke setup(null) in order to get the updated logic if it changes

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -437,8 +513,71 @@ protected FlowNode next(@Nonnull FlowNode current, @Nonnull Collection<FlowNode>
...
+                    break;
+                case PARALLEL_BRANCH_END:
+                    visitor.parallelBranchEnd(this.currentParallelStartNode, myCurrent, this);
+                    break;
+                case PARALLEL_BRANCH_START:
+                    // Needed because once we hit the start of the last branch, the next node is our currentParallelStart
+                    FlowNode parallelStart = (nextType == NodeType.PARALLEL_START) ? myNext : this.currentParallelStartNode;
+                    visitor.parallelBranchStart(parallelStart, myCurrent, this);
+                    break;
+                default:
```
oleg-nenashev: Good practice: break on NORMAL and throw IllegalStateException on default. Just to avoid missing logic processing if the node types get extended

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/MemoryFlowChunk.java
oleg-nenashev: Yep, bug was likely an overkill

oleg-nenashev:
🐝 with minor comments left

vivek:
@svanoort LGTM 🐝

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
svanoort: It's safe to run without initialization here -- you just get no nodes.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
svanoort: It lacks punch ;)

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -437,8 +513,71 @@ protected FlowNode next(@Nonnull FlowNode current, @Nonnull Collection<FlowNode>
...
+                    throw new IllegalStateException("Unhandled type for current node");
```
oleg-nenashev: Ideally it should mention the type you've got

oleg-nenashev:
re-:bee:

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/StandardChunkVisitor.java
```diff
@@ -0,0 +1,64 @@
+package org.jenkinsci.plugins.workflow.graphanalysis;
+
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+
+/**
+ * Simple handler for linear chunks (basic stages, etc), designed to be extended
```
jglick: :ant: Javadoc summary sentences must end with a period.

pom.xml
```diff
@@ -63,6 +63,7 @@
     </pluginRepositories>
     <properties>
         <jenkins.version>1.642.3</jenkins.version>
+        <additionalparam>-Xdoclint:none</additionalparam>
```
jglick: :ant: delete and fix your Javadoc `error`s

src/main/java/org/jenkinsci/plugins/workflow/actions/FlowNodeAction.java
```diff
@@ -42,7 +42,7 @@
      * Called by {@link FlowExecution#loadActions(FlowNode)} when
      * actions get loaded from persistent storage.
      *
-     * <p>
+     * <p></p>
```
jglick: :ant: this is wrong, revert

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/AbstractFlowScanner.java
```diff
@@ -43,9 +43,9 @@
 /**
  * Core APIs and base logic for FlowScanners that extract information from a pipeline execution.
  *
- * <p/>These iterate through the directed acyclic graph (DAG) or "flow graph" of {@link FlowNode}s produced when a pipeline runs.
+ * <p></p>These iterate through the directed acyclic graph (DAG) or "flow graph" of {@link FlowNode}s produced when a pipeline runs.
```
jglick: :ant: ditto, just use `<p>`

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ChunkFinder.java
```diff
@@ -0,0 +1,34 @@
+package org.jenkinsci.plugins.workflow.graphanalysis;
+
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+
+/**
+ * Matches the start and end of a chunk
```
jglick: So, what is a “chunk” and why would you want to find one?

jglick: I am guessing this type is just a helper that you pass to something else that is better documented—so link to that Javadoc in case someone is unlucky enough to start browsing here.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowChunk.java
```diff
@@ -0,0 +1,53 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.jenkinsci.plugins.workflow.graphanalysis;
```
jglick: Does the stuff pertaining to “chunks” belong in a nested package for clarity?

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -42,27 +46,47 @@
...
+    public NodeType getNextType() {
+        return nextType;
+    }
+
+    /** Used to recognize special nodes */
+    public enum NodeType {
+        NORMAL,
+        PARALLEL_START,
+        PARALLEL_END,
+        PARALLEL_BRANCH_START,
```
jglick: I _think_ I know what you are referring to here, but it would be best to add Javadoc for everyone else.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/LabelledChunkFinder.java
```diff
@@ -0,0 +1,42 @@
...
+        return true;
+    }
+
+    @Override
+    public boolean isChunkStart(@Nonnull FlowNode current, @CheckForNull FlowNode previous) {
+        LabelAction la = current.getAction(LabelAction.class);
+        return la != null;
+    }
+
+    /** End is where you have a label marker before it... or  */
```
jglick: Or…?

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/MemoryFlowChunk.java
```diff
@@ -0,0 +1,99 @@
...
+package org.jenkinsci.plugins.workflow.graphanalysis;
+
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+
+/**
+ * FlowChunk that holds direct references to the {@link FlowNode} instances and context info
```
jglick: :ant: end summary sentence with `.`

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/SimpleChunkVisitor.java
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Called when hitting the start of a chunk
+     * @param startNode First node in chunk (marker), included in node
+     * @param beforeBlock First node before chunk
+     * @param scanner Forkscanner used (for state tracking)
+     */
+    void chunkStart(@Nonnull FlowNode startNode, @CheckForNull FlowNode beforeBlock, @Nonnull ForkScanner scanner);
+
+    /** Called when hitting the end of a block */
```
jglick: documentation of parameters would be helpful

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/SimpleChunkVisitor.java
```diff
@@ -0,0 +1,114 @@
...
+     * @param parallelStartNode The {@link org.jenkinsci.plugins.workflow.graph.BlockStartNode} beginning it, next will be branches
+     * @param branchNode {@link org.jenkinsci.plugins.workflow.graph.BlockStartNode} for one of the branches (it will be labelled)
+     * @param scanner ForkScanner used
+     */
+    void parallelStart(@Nonnull FlowNode parallelStartNode, @Nonnull FlowNode branchNode, @Nonnull ForkScanner scanner);
+
+    /**
+     * Notifies that we've seen the end of a parallel block
+     * @param parallelStartNode First node of parallel (BlockStartNode before the branches)
+     * @param parallelEndNode Last node of parallel (BlockEndNode)
```
jglick: Use `@link` when referring to particular types.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/StandardChunkVisitor.java
```diff
@@ -0,0 +1,64 @@
...
+ * Note: does not handle parallels or nesting
```
jglick: Can you expand on what concretely it does not handle, and how it behaves if given graphs with such structures?

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/TestVisitor.java
```diff
@@ -0,0 +1,110 @@
...
+    public static class CallEntry {
+        CallType type;
+        int[] ids = {-1, -1, -1, -1};
+
+        public void setIds(FlowNode... nodes) {
+            for (int i=0; i<nodes.length; i++) {
+                if (nodes[i] == null) {
+                    ids[i] = -1;
+                } else {
+                    ids[i] = Integer.parseInt(nodes[i].getId());
```
jglick: :ant: best not to do this even in tests.

jglick:
:bee: without much comprehension

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/StandardChunkVisitor.java
svanoort: You realize how silly this is, right?

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/LabelledChunkFinder.java
svanoort: Life is better with a little mystery in it.

src/main/java/org/jenkinsci/plugins/workflow/pickles/Pickle.java
```diff
@@ -50,7 +50,7 @@
      * An implementation should return quickly and avoid acquiring locks in this method itself (as opposed to the future).
      * {@link ListenableFuture#cancel} should be implemented if possible.
      * @param owner an owner handle on which you may, for example, call {@link FlowExecutionOwner#getListener}
-     * @return a future on which {@link ListenableFuture#cancel} might be called; also polite to override {@link ListenableFuture#toString} for diagnostics
+     * @return a future on which {@link ListenableFuture#cancel(boolean)} might be called; also polite to override the toString method for diagnostics
```
jglick: BTW this will be a merge conflict with #11.

jglick: (prefer if you copy my change)

jglick:
re:bee:

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowChunk.java
jglick: Specifically if you look at [the current package](https://github.com/jenkinsci/workflow-api-plugin/tree/block-scanning-APIs/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis) it is not immediately apparent that `ForkScanner` plus all the types mentioning `Chunk` form a logical unit of usage.

svanoort: Docs gap, yes.  Not a bad point, just a bit too late to reasonably do a restructuring like this and propagate to dependent codebases.

vivek:
This is going to be very useful as lightweight mechanism to add key/value pairs. Saves from creating marker interface.  However there is value in adding Action objects, for example an action object added by Kyoto like plugin as Java API of its own. There is also issue of key collisions.

svanoort:
@vivek Wrong PR ;-)

vivek:
Oh sorry @svanoort, doing PR comment on mobile devices not a great idea:)

PR #12, opened 2016-09-13
Create a new TagsAction [JENKINS-26522]

svanoort:
Used for a variety of annotations for [JENKINS-26522](https://issues.jenkins-ci.org/browse/JENKINS-26522)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/TagsAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/TagsAction.java
@@ -85,12 +85,12 @@ public class TagsAction implements Action{
     }
 
     /**
-     * Get the set of tag-value mappings
+     * Get the tag-value mappings
      * @return Unmodifiable view of tag-value mappings
      */
     @Nonnull
-    public Set<Map.Entry<String,String>> getTagMappings() {
-        return Collections.unmodifiableSet(tags.entrySet());
+    public Map<String,String> getTags() {
+        return Collections.unmodifiableMap(tags);
     }
 
     // Static convenience methods
@@ -100,9 +100,9 @@ public class TagsAction implements Action{
      * @return Unmodifiable view of tag-value mappings
      */
     @Nonnull
-    public static Set<Map.Entry<String,String>> getTagMappings(@Nonnull  FlowNode node) {
+    public static Map<String,String> getTags(@Nonnull  FlowNode node) {
         TagsAction tagAction = node.getAction(TagsAction.class);
-        return (tagAction == null) ? (Set)(Collections.emptySet()) : tagAction.getTagMappings();
+        return (tagAction == null) ? (Map)(Collections.emptyMap()) : tagAction.getTags();
     }
 
     /**
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/TagsAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/TagsAction.java
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2013-2014, CloudBees, Inc.
+ * Copyright (c) 2016, CloudBees, Inc.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


svanoort:
First peek for @jglick and @abayer  and @vivek  -- not fully fleshed out but you can get the idea.

abayer:
Woo. That's a start. =) I still want gobs more, of course - JUnit results attached to a node! Status! Monkeys for some reason! But I'll be patient. =)

svanoort:
@abayer I think JUnit results probably will want either an identifier tag (JUNIT_RESULT_12345) and/or a JUNIT Tag.   Still undecided how much to attach to the FlowNode itself vs. the WorkflowRun but there need to be some more actions anyway.

Monkeys?  Nonono, I was doing Llamas. Let's talk about a refactor.

abayer:
Monkeys or GTFO. 🐒 🐒 🐒

abayer:
But yes, :bee:. This rates 1.5 🐒s.

abayer:
Also 1/3 of a 🍌 

Oh god, I've gotten hooked on emojis.

svanoort:
@abayer Might want to wait until it's marked for review to bee it, but I appreciate your enthusiasm and, ah initiative ;)

Did a quick bounce of the concept off @jglick and he favors a tag-value approach over simple marker tags.  That would be a fairly trivial change too (in that case you'd have annotations 'SYNTHETIC' => pre and 'SYNTHETIC' => post ).   Slightly more complex but not much.  Either way could finalize for merge very fast. 

Thoughts?  Also cc @vivek

abayer:
Ok, de-:bee:, one extra 🐒

abayer:
And yes, key/value a la AWS tags is definitely superior to raw tag now that I think about it. Makes things much more flexible.

vivek:
This is going to be very useful as lightweight mechanism to add key/value pairs. Saves from creating marker interface. However there is value in adding Action objects, for example an action object added by Kyoto like plugin as Java API of its own. There is also issue of key collisions.

svanoort:
@vivek That's a valid point -- but why would you want to put complex logic in the Action itself? 

Tag strings can be namespaced if desired, i.e. org.jenkinsci.foo.bar.BarStatusLookup. 

We're only aiming for a lightweight solution to cover the 90% case so we don't end up with 9999 individual micro-actions that do subsets of this. 

Also worth mentioning: storing and retrieving individual actions is actually shockingly expensive (on its own, about 25% of generating the run digest in stage view ).  Using a key-value store lets you do one fetch and then extract multiple kinds of information.

abayer:
@svanoort I think it's less complex logic in `Action`s than it is complex data. But yeah, bridge we can cross later.

vivek:
@svanoort its good for what it is, enforcing namespace is not easy. I think Action as Java object with its API and data it exposes is valuable. It gives you a contract to work with vs deal with properties. Both have valid use case is what I am saying.

svanoort:
@reviewbybees Especially @abayer and @vivek

src/main/java/org/jenkinsci/plugins/workflow/actions/TagsAction.java
```diff
@@ -0,0 +1,138 @@
...
+
+    private LinkedHashMap<String, String> tags = new LinkedHashMap<String, String>();
+
+    /**
+     * Add a tag key:value pair to this FlowNode, null or empty values are ignored
+     * Inputs are CheckForNull so you can directly pass in values without nullchecks upfront.
+     * @param tag Tag to add to, null or empty values are no-ops
+     * @param value Tag to add to, null or empty values are no-ops
+     */
+    public void addTag(@CheckForNull String tag, @@CheckForNull String value) {
```
abayer: `@@`? =)

svanoort:
@vivek Oh, agreed -- this isn't intended to replace general Actions, it's just intended to avoid reserve new action creation for when it's adding something. 

Also because we have use cases this covers well.

abayer:
:bee:

jglick:
Is there only one demonstrated use case for this so far? If so, it is premature to commit to an API. For the particular use case as I understand it (very dimly), a dedicated `Action` and a `Step` to attach it would perhaps make more sense. The JIRA RFE requires quite different things.

svanoort:
@jglick Specifically for https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/13 at the moment. 

Of course once, we have it in, there's visualization options too...

vivek:
@svanoort 🐝 LGTM

PR #14, opened 2016-09-19
Revert "Create a new TagsAction [JENKINS-26522]"

svanoort:
Reverts jenkinsci/workflow-api-plugin#12


svanoort:
At request of @jglick  because @abayer hasn't finished https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/13

jglick:
:bee: though you need no PR just to revert something I think.

PR #15, opened 2016-09-22
Make DepthFirstScanner obey last->first parallel branch ordering [JENKINS-38458]

svanoort:
Per [JENKINS-38458](https://issues.jenkins-ci.org/browse/JENKINS-38458) and it will also incidentally solve [JENKINS-38457](https://issues.jenkins-ci.org/browse/JENKINS-38457).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/DepthFirstScanner.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/DepthFirstScanner.java
@@ -24,6 +24,7 @@
 
 package org.jenkinsci.plugins.workflow.graphanalysis;
 
+import com.google.common.collect.Lists;
 import org.jenkinsci.plugins.workflow.graph.BlockStartNode;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 
@@ -31,6 +32,7 @@ import javax.annotation.Nonnull;
 import javax.annotation.concurrent.NotThreadSafe;
 import java.util.ArrayDeque;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
@@ -38,9 +40,11 @@ import java.util.List;
 /** Does a simple and somewhat efficient depth-first search of all FlowNodes in the DAG.
  *
  *  <p>Iteration order: depth-first search, revisiting parallel branches once done.
- *  With parallel branches, the first branch is explored, then remaining branches are explored in reverse order.
  *
  * <p> The behavior is analogous to {@link org.jenkinsci.plugins.workflow.graph.FlowGraphWalker} but faster.
+ *  With parallel branches, the last branch is explored, then remaining branches are explored in reverse order.
+ *  This keeps ordering compatibility with FlowGraphWalker.
+ *
  *  @author Sam Van Oort
  */
 @NotThreadSafe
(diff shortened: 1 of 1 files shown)


svanoort:
@reviewbybees especially @jglick

jglick:
```
org.junit.internal.ArrayComparisonFailure: Depth first: arrays first differed at element [3]; expected:<[9]> but was:<[12]>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.junit.internal.ExactComparisonCriteria.assertElementsEqual(ExactComparisonCriteria.java:8)
	at org.junit.internal.ComparisonCriteria.arrayEquals(ComparisonCriteria.java:53)
	at org.junit.Assert.internalArrayEquals(Assert.java:532)
	at org.junit.Assert.assertArrayEquals(Assert.java:283)
	at org.jenkinsci.plugins.workflow.graphanalysis.FlowTestUtils.assertNodeOrder(FlowTestUtils.java:84)
	at org.jenkinsci.plugins.workflow.graphanalysis.FlowTestUtils.assertNodeOrder(FlowTestUtils.java:93)
	at org.jenkinsci.plugins.workflow.graphanalysis.FlowScannerTest.testParallelScan(FlowScannerTest.java:357)
```

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/DepthFirstScanner.java
```diff
@@ -24,23 +24,27 @@
...
 import java.util.List;
 
 /** Does a simple and somewhat efficient depth-first search of all FlowNodes in the DAG.
  *
  *  <p>Iteration order: depth-first search, revisiting parallel branches once done.
- *  With parallel branches, the first branch is explored, then remaining branches are explored in reverse order.
  *
  * <p> The behavior is analogous to {@link org.jenkinsci.plugins.workflow.graph.FlowGraphWalker} but faster.
+ *  With parallel branches, the last branch is explored, then remaining branches are explored in reverse order.
+ *  This keeps ordering compatibility with FlowGraphWalker.
```
jglick: `@link`

jglick:
And does this really solve JENKINS-38457? Did you try the test which reproduced the problem? I suspect it would make it show C, B, A rather than A, B, C as expected.

jglick:
I was wrong, but it does still fail: it shows C, A, B.

jglick:
```diff
diff --git a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/DepthFirstScanner.java b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/DepthFirstScanner.java
index 7fd3e9a..f839c0b 100644
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/DepthFirstScanner.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/DepthFirstScanner.java
@@ -84,11 +84,7 @@ public class DepthFirstScanner extends AbstractFlowScanner {
             // Only ParallelStep nodes may be visited multiple times... but we can't just filter those
             // because that's in workflow-cps plugin which depends on this one.
             if (!blackList.contains(f) && !(f instanceof BlockStartNode && visited.contains(f))) {
-                if (output == null ) {
-                    output = f;  // Do direct assignment rather than needless push/pop
-                } else {
-                    queue.push(f);
-                }
+                queue.push(f);
             }
         }
```

at least makes `DepthFirstScanner` behave according to its revised Javadoc. `FlowGraphTable` is still broken, because it is getting branches in reverse order, where before it was getting them in forward order, so a patch is needed in `workflow-support` as well.

svanoort:
Will come back to this later - was throwing it out there for CI to run tests against.

jglick:
BTW if you are not really proposing something for immediate merge, you can add the `work–in–progress` label (intentionally using Unicode n-dashes here so it does not get picked up by tools which scan for the same text in comments). That hides it from certain dashboards and sets expectations for reviewers as well.

PR #17, opened 2016-09-23
[JEP-210] Log handling rewrite

jglick:
[JEP-210](https://github.com/jenkinsci/jep/blob/master/jep/210/README.adoc)

- [X] way to get a log

@reviewbybees esp. @oleg-nenashev


src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionOwner.java
```diff
@@ -121,11 +125,26 @@ public String getUrlOfExecution() throws IOException {
...
+    /**
+     * Gets the current log stream.
+     * May receive EOF even while a build is in progress.
+     * This is the raw log which may contain encoded {@link ConsoleNote}s.
+     * The encoding is assumed to be UTF-8.
+     * The caller should perform buffering if desired.
+     * @param start the start position to begin reading from (normally 0)
+     * @throws EOFException if the start position is larger than the log size (or you may simply return EOF immediately when read)
+     */
+    public @Nonnull InputStream getLog(long start) throws IOException {
```
carlossg: `@Beta` ?

jglick: Yes, ought to be.

src/main/java/org/jenkinsci/plugins/workflow/log/LogStorage.java
```diff
@@ -0,0 +1,129 @@
...
+        try {
+            for (LogStorageFactory factory : ExtensionList.lookup(LogStorageFactory.class)) {
+                LogStorage storage = factory.forBuild(b);
+                if (storage != null) {
+                    // TODO consider saving this decision
+                    return storage;
+                }
+            }
+            // Similar to Run.getLogFile, but not supporting gzip:
+            return StreamLogStorage.forFile(new File(b.getRootDir(), "log"), b);
```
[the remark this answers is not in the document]
svanoort: Yes, after discussion with @jglick and @dwnusbaum the best path is to keep something with similar performance scaling to the older file-per-step logging -- we get to keep good seek time so every single step where we view logs in BlueOcean and StageView, only results in reading *that* part of the log, rather than the entirety of the log (potentially hundreds of MB or more). 

It was *almost* a wash vs trying to fix the streaming implementation to support fast seeks without reading the entire log, but we opted to not go that path at this time to allow more time to work on that.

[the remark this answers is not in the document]
jglick: After thinking it over a bit more, I am optimistic that a single-file implementation with a rangeset side file may not be that complicated. I will take a shot at it, since that would be a better outcome.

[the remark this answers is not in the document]
svanoort: @jglick You're welcome to, but after thinking on it a bit myself I think we *really* should rely on having potentially >1 implementation on-master, in case we discover a performance failure mode with this (as noted, rangesets are weak with interleaved parallel branch content, and I can name at least one user who does that with massively parallel branches).

src/main/java/org/jenkinsci/plugins/workflow/log/StreamLogStorage.java
```diff
@@ -0,0 +1,331 @@
...
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.Beta;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+import org.kohsuke.stapler.framework.io.ByteBuffer;
+
+/**
+ * Simple implementation of log storage that intermixes node and general messages in a single log bytestream.
+ */
+@Restricted(Beta.class)
+public abstract class StreamLogStorage implements LogStorage {
```
svanoort: 🐛  For the record, this is what *currently* requires reading the entire log in order to fetch log information for *any* FlowNode viewed.  This will likely introduce serious performance problems on masters because users will be opening logs for one or more builds from the UI and logs often run 10 MB+, sometimes 100 MB to 1 GB.

If this behavior problem is fixed, this is viable though -- this can be solved bye storing ranges for logs contained in each FlowNode, in the LogAction added here.

jglick:
> read entire build log just to fetch data for one FlowNode in Blue Ocean

Should be addressed in new `FileLogStorage`: the I/O to serve the text for a `FlowNode` should be proportional to the served size, plus the number of node ID transitions (likely small except in pathologically heavy interleaving).

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
```diff
@@ -0,0 +1,256 @@
...
+        }
+
+        @Override public void flush() throws IOException {
+            os.flush();
+        }
+
+        @Override public void close() throws IOException {
+            if (id == null) {
+                os.close();
+                indexOs.close();
```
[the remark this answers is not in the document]
jglick: Yes, would be a bit safer. In practice AFAIK no `OutputStream.close` implementations actually _do_ throw `IOException` under any circumstances, but you never know.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
```diff
@@ -0,0 +1,256 @@
...
+                        long pos = start;
+                        boolean hadLastId;
+                        @Override public void write(int b) throws IOException {
+                            String line;
+                            while (lastTransition < pos && (line = indexBR.readLine()) != null) {
+                                int space = line.indexOf(' ');
+                                try {
+                                    lastTransition = Long.parseLong(space == -1 ? line : line.substring(0, space));
+                                } catch (NumberFormatException x) {
+                                    // corrupt index file; forget it
```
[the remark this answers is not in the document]
jglick: Yes.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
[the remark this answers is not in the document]
dwnusbaum: Hmm interesting, that's good to know.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
```diff
@@ -0,0 +1,271 @@
...
+                }
+            }
+        }
+
+    }
+
+    @Override public AnnotatedLargeText<FlowExecutionOwner.Executable> overallLog(FlowExecutionOwner.Executable build, boolean complete) {
+        return new AnnotatedLargeText<FlowExecutionOwner.Executable>(log, StandardCharsets.UTF_8, complete, build) {
+            @Override public long writeHtmlTo(long start, Writer w) throws IOException {
+                try (BufferedReader indexBR = index.isFile() ? Files.newBufferedReader(index.toPath(), StandardCharsets.UTF_8) : new BufferedReader(new NullReader(0))) {
```
jglick: Seems that `new NullReader(0)` is buggy.

jglick: …or not exactly. Apparently it throws an `IOException` (not `EOFException`) if you call `read()` after having already gotten -1 ~ EOF, unlike every other `Reader` implementation I can find which just continues to return -1. In other words, it is stricter, which is fine since it turned up a mistake in the caller here.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
```diff
@@ -0,0 +1,276 @@
...
+    private FileOutputStream os;
+    private Writer indexOs;
+    private String lastId;
+
+    private FileLogStorage(File log) {
+        this.log = log;
+        this.index = new File(log + "-index");
+    }
+
+    private synchronized void open() throws IOException {
```
svanoort: @jglick Would be great to get a comment to explain this -- especially curious if there's a specific reason why we're reading the entire file line by line in order to get the last line, rather than using a RandomAccessFile or similar and scanning in blocks from the end looking for a new line separator...

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
```diff
@@ -0,0 +1,276 @@
...
+                    String lastLine = null;
+                    while (true) {
+                        String line = r.readLine();
+                        if (line == null) {
+                            break;
+                        } else {
+                            lastLine = line;
+                        }
+                    }
+                    if (lastLine != null) {
```
svanoort: What if the content did not finish writing and the "lastId" has been truncated?  (i.e. the line has only partially written)

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
svanoort: Actually, scratch that, couldn't you just look for a space?

jglick: This was just the simplest code to write. If there is any specific reason to suspect a performance issue, it could of course be optimized to scan the file backwards for the penultimate `\n`, then convert the byte sequence from there to EOF to UTF-8 and set `lastId` accordingly.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
jglick: Then the line → step mapping for this build will not be perfect. For example, in classic UI it would mean that some text would be incorrectly not associated with a step block but just shown as part of general output. Too minor to worry about.

svanoort: Sounds like a bug to me, but I'll let you fix it when the first user reports it.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
```diff
@@ -0,0 +1,276 @@
...
+        }
+
+    }
+
+    @Override public AnnotatedLargeText<FlowExecutionOwner.Executable> overallLog(FlowExecutionOwner.Executable build, boolean complete) {
+        return new AnnotatedLargeText<FlowExecutionOwner.Executable>(log, StandardCharsets.UTF_8, complete, build) {
+            @Override public long writeHtmlTo(long start, Writer w) throws IOException {
+                try (BufferedReader indexBR = index.isFile() ? Files.newBufferedReader(index.toPath(), StandardCharsets.UTF_8) : new BufferedReader(new NullReader(0))) {
+                    ConsoleAnnotationOutputStream<FlowExecutionOwner.Executable> caos = new ConsoleAnnotationOutputStream<>(w, ConsoleAnnotators.createAnnotator(build), build, StandardCharsets.UTF_8);
+                    long r = this.writeRawLogTo(start, new FilterOutputStream(caos) {
```
svanoort: @jglick A block of code this dense really merits some comments to explain what it's doing and why -- otherwise any time we try to modify it we'll have to reverse-engineer the entirety.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
```diff
@@ -0,0 +1,276 @@
...
+                    });
+                    ConsoleAnnotators.setAnnotator(caos.getConsoleAnnotator());
+                    return r;
+                }
+            }
+        };
+    }
+
+    @Override public AnnotatedLargeText<FlowNode> stepLog(FlowNode node, boolean complete) {
+        String id = node.getId();
```
svanoort: @jglick Similarly, while this has a few comments it is not clear what the intent is without reverse-engineering the intent from what it's doing.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
```diff
@@ -0,0 +1,276 @@
...
+                    }
+                } else {
+                    raf.seek(pos);
+                    if (lastTransition > pos + Integer.MAX_VALUE) {
+                        throw new IOException("Cannot read more than 2Gib at a time"); // ByteBuffer does not support it anyway
+                    }
+                    // TODO can probably be done a bit more efficiently with FileChannel methods
+                    byte[] data = new byte[(int) (lastTransition - pos)];
+                    raf.readFully(data);
+                    buf.write(data);
```
svanoort: 🐜 Could we not write this a chunk at a time or stream it?  This is going to generate a *lot* of GC pressure due to excess array allocation.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
vivek: +1. Yeah, meaningful and granular comments is going to help code maintainability.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
jglick: ack

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
jglick: Sort of beside the point when the whole `buf` is memory-allocated. Which is hard to get around in current core versions since the only `[Annotated]LargeText` constructors take a file (useless) or a memory buffer. To write this properly, we would likely need for the [`Source`](https://github.com/stapler/stapler/blob/cafcb607e9e957bc8f57782eab3b0568a2ab3f42/core/src/main/java/org/kohsuke/stapler/framework/io/LargeText.java#L60-L67) and [`Session`](https://github.com/stapler/stapler/blob/cafcb607e9e957bc8f57782eab3b0568a2ab3f42/core/src/main/java/org/kohsuke/stapler/framework/io/LargeText.java#L381-L390) interfaces to be made `public`.

svanoort: Answering my own question after digging in a bit more: no, because our `ByteBuffer` cannot be pre-sized for our content since its `ensureCapacity` method is private.

For those who can't follow the code: essentially what we're doing it allocating an array containing all the log content and then copying it into the Jenkins-internal `ByteBuffer` implement (`org.kohsuke.stapler.framework.io.ByteBuffer`) which will dynamically resize to fit the content.  Essentially we're double-allocating.  

However the ByteBuffer will resize to *either* double *or* fit content (whichever is larger).   Which means that if we copied a chunk at a time, we'd likely end up triggering multiple backing-array resize operations for the `ByteBuffer.`

Kind of a wash in that case, without a way to more effectively size the buffer, which would require upstream changes to Stapler (which can be done later).

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
```diff
@@ -0,0 +1,276 @@
...
+             BufferedReader indexBR = index.isFile() ? Files.newBufferedReader(index.toPath(), StandardCharsets.UTF_8) : new BufferedReader(new NullReader(0))) {
+            String line;
+            long pos = -1; // -1 if not currently in this node, start position if we are
+            while ((line = indexBR.readLine()) != null) {
+                int space = line.indexOf(' ');
+                long lastTransition = -1;
+                try {
+                    lastTransition = Long.parseLong(space == -1 ? line : line.substring(0, space));
+                } catch (NumberFormatException x) {
+                    break; // corrupt index file; forget it
```
svanoort: 🐛  One line that is not number-parseable (perhaps due to interrupted writes or OS write caching under circumstances similar to those described in https://github.com/jenkinsci/jenkins/pull/3170) will prevent reading the entire log file IIUC.  What you *probably* want to do is log the exception somewhere and `continue`.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
svanoort: AFAICT the `FileChannel` comment -- which is another clever option -- won't work because our `ByteBuffer` (drumroll please) _does not_ extend off of `java.nio.ByteBuffer`... instead it extends `OutputStream`  😆 

So... I smell some room for core improvements in LargeText/AnnotatedLargeText that reduce our tendency to throw around (needless) byte array allocations.  Might help with some of the humongous object allocations I've seen spiking our GC logs.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    @Override public AnnotatedLargeText<FlowNode> stepLog(FlowNode node, boolean complete) {
+        String id = node.getId();
+        try (ByteBuffer buf = new ByteBuffer();
+             RandomAccessFile raf = new RandomAccessFile(log, "r");
+             BufferedReader indexBR = index.isFile() ? Files.newBufferedReader(index.toPath(), StandardCharsets.UTF_8) : new BufferedReader(new NullReader(0))) {
+            String line;
+            long pos = -1; // -1 if not currently in this node, start position if we are
+            while ((line = indexBR.readLine()) != null) {
```
svanoort: For those keeping track at home: any time we try to read a single node's logs we trigger a full read of the entire index file, and as we scan we look for lines with our FlowNode ID.  Which thankfully is *smaller* than the *entire* logfile by probably orders of magnitude. 

But still this feels both overly complicated and inefficient - with `O(m)` performance to find all a FlowNode's logs, where `m` is the size of the index.   Is there a reason we can't simply store when FlowNode is "done" for logging purposes -- we have that logic handy already.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
```diff
@@ -0,0 +1,276 @@
...
+                        throw new IOException("Cannot read more than 2Gib at a time"); // ByteBuffer does not support it anyway
+                    }
+                    // TODO can probably be done a bit more efficiently with FileChannel methods
+                    byte[] data = new byte[(int) (lastTransition - pos)];
+                    raf.readFully(data);
+                    buf.write(data);
+                    pos = -1;
+                }
+            }
+            if (pos != -1) {
```
svanoort: Comment: // Reading out the final dangling chunk of log content if we're still writing content for the node.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
```diff
@@ -0,0 +1,276 @@
...
+                raf.seek(pos);
+                long end = raf.length();
```
svanoort: Weird edge-casey possibility: what if we've logged a new FlowNode in between reading the last index entry and when we check the length of the file via `raf.length()`?  Wouldn't that make it possible for that content to be added as if it's part of this node's log even though it should not be?   

Probably way too edge-casey and timing sensitive to worry about, but interesting nonetheless.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
svanoort: Though who knows, this logic is like reading Don Quixote in the original old Spanish

jglick: Actually my uncle-in-law is a Don Quixote expert!

> will prevent reading the entire log file

Will prevent the step log association. The whole build log will be fine except for possibly missing some step annotations.

> and `continue`

Not so simple. If `log-index` is corrupt in one line, at best you can hope that the _next_ line (next transition) is accurate. Certainly this step’s log is going to be incomplete or empty.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
jglick: > store when FlowNode is "done"

Not sure what you are proposing exactly.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
jglick: I think it is possible in principle. The `log-index` writes are flushed and that happens “before” we start writing to `log`. But reading `log-index` will take a little bit of time, and of course on a network filesystem there are few guarantees about write order. At worst you momentarily get a bit of output in the wrong place, which is not a big deal, but it might be possible to tighten this up somewhat.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
svanoort: @jglick We know how to compute when a FlowNode is "closed" for logging, depending on if it's an AtomNode, StartNode, or EndNode.   This limits the scope we'd need to scan. 

Now, if we specially marked the offset for the first and last log content for a FlowNode, and stored it specially (say as fields in our LogAction implementation, or showing boundaries a bit more clearly in the index) we could further limit the amount of scanning done. 

However... eh, this can be optimized later.

jglick: > stored it specially (say as fields in our `LogAction` implementation

I preferred to avoid duplicating information like this. In other words, `log` + `log-index` should be self-contained.

Now another line type (not a transition) _could_ be added to `log-index` indicating that a given `FlowNode` has now been closed and that thus there is no need to search past that point in the file looking for subsequent blocks. The extra complexity just to avoid reading the rest of what would usually be a short text file (smaller than a typical “disk” block size I think) does not seem worth it.

svanoort: Please leave this comment *unresolved* so we know in the future where to start if we need to optimize.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
svanoort: @jglick The catch here is that we may have a single log index line that is corrupted but the rest may be okay (for example if the AWS or K8S volume is abruptly yanked) but the master restarts and resumes the build later.  

So it is quite possible that the index may have a *single* corrupted or missing entry with the rest of the index being fine. 

Because we break at the *first* NumberFormatException, we will not see logs for any FlowNodes after the failure.  However if we do a `continue` we can accept that failure and still get log context for the following steps.

svanoort: Resolved via code change.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
svanoort: Fair enough -- nothing to do here at the moment but I'd like to leave the conversation marked unresolved to nudge us about future enhancements to do.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
svanoort: @jglick Side comment: the simile of reading Don Quixote in the original Spanish is particularly appropriate because I did a fair bit of that (laboriously with painful effort because while I was almost-fluent in the language at the time, the particular usages were quite painful).

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
svanoort: I confirm this was resolved appropriately via code change -- the current behavior reads to the end of the file (based on the *initial* length).  Changes made *while* this call is executing will not be reflected in the terminal output -- which feels intuitively more correct.

PR #18, opened 2016-09-24
[JENKINS-34488] Use ProxyException for exceptions rejected by XStream

ikedam:
[JENKINS-34488](https://issues.jenkins-ci.org/browse/JENKINS-34488) Failing 'assert' hangs the pipeline

`hudson.util.XStream2` rejects to serialize some classes:
* https://github.com/jenkinsci/jenkins/blob/jenkins-2.22/core/src/main/java/hudson/util/XStream2.java#L442
* https://github.com/jenkinsci/remoting/blob/remoting-2.62/src/main/java/hudson/remoting/ClassFilter.java#L56

This results an unexpected exception when saving `ErrorAction` with some specific exceptions, 
E.g. `org.codehaus.groovy.runtime.powerassert.PowerAssertionError` thrown by `assert` without a message parameter in groovy.
Consequent codes in workflow-cps-plugin is skipped for the exception, and this results hang the pipeline.

`ErrorAction` already has codes to use `ProxyException` for unserializable exceptions.
This change marks exceptions that cannot be serialized by `hudson.util.XStream2` unserializable, which makes `ErrorAction` use `ProxyException` also for exceptions such as `PowerAssertionError`.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
@@ -25,6 +25,7 @@
 package org.jenkinsci.plugins.workflow.actions;
 
 import groovy.lang.MissingMethodException;
+import hudson.remoting.ClassFilter;
 import hudson.remoting.ProxyException;
 import org.codehaus.groovy.control.MultipleCompilationErrorsException;
 import org.jenkinsci.plugins.workflow.graph.AtomNode;
@@ -53,6 +54,15 @@ public class ErrorAction implements Action {
      * an equivalent that captures the same details but serializes nicely.
      */
     private boolean isUnserializableException(Throwable error) {
+        try {
+            // Some exceptions are refused to be serialized for security reasons.
+            // (E.g. PowerAssertionError thrown by "assert false")
+            // See also hudson.util.XStream2
+            ClassFilter.DEFAULT.check(error.getClass());
+            ClassFilter.DEFAULT.check(error.getClass().getName());
+        } catch (SecurityException x) {
+            return true;
+        }
         return error instanceof MultipleCompilationErrorsException ||
                error instanceof MissingMethodException;
     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
@@ -54,6 +54,10 @@ public class ErrorAction implements Action {
      * an equivalent that captures the same details but serializes nicely.
      */
     private boolean isUnserializableException(Throwable error) {
+        if (error == null) {
+            // This shouldn't happen.
+            return false;
+        }
         try {
             // Some exceptions are refused to be serialized for security reasons.
             // (E.g. PowerAssertionError thrown by "assert false")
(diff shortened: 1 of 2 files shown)


jglick:
This is the replacement for https://github.com/jenkinsci/workflow-cps-plugin/pull/62 I suppose.

src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
```diff
@@ -53,6 +54,15 @@ public ErrorAction(Throwable error) {
      * an equivalent that captures the same details but serializes nicely.
      */
     private boolean isUnserializableException(Throwable error) {
+        try {
+            // Some exceptions are refused to be serialized for security reasons.
+            // (E.g. PowerAssertionError thrown by "assert false")
+            // See also hudson.util.XStream2
+            ClassFilter.DEFAULT.check(error.getClass());
```
jglick: BTW this effectively obsoletes the `assert`ion on line 48.

src/test/java/org/jenkinsci/plugins/workflow/actions/ErrorActionTest.java
```diff
@@ -0,0 +1,110 @@
...
+                ret.add(e);
+            }
+        }
+        return ret;
+    }
+
+    @Test
+    public void simpleException() throws Exception {
+        final String EXPECTED = "For testing purpose";
+        WorkflowJob job = r.jenkins.createProject(WorkflowJob.class, r.getTestDescription().getMethodName());
```
jglick: Easier to just hardcode a name like `"p"`.

src/test/java/org/jenkinsci/plugins/workflow/actions/ErrorActionTest.java
```diff
@@ -0,0 +1,110 @@
...
+        job.setDefinition(new CpsFlowDefinition(
+                "node {\n"
+                        + "assert false;\n"
+                + "}"
+        ));
+        WorkflowRun b = r.assertBuildStatus(Result.FAILURE, job.scheduleBuild2(0).get());
+        List<ErrorAction> errorActionList = extractErrorActions(b.asFlowExecutionOwner().get());
+        assertThat(errorActionList, Matchers.not(Matchers.empty()));
+        for (ErrorAction e : errorActionList) {
+            assertEquals(ProxyException.class, e.getError().getClass());
```
jglick: More interesting would be to have a script which runs, say

```groovy
assert 2 + 2 == 5
```

and verify that the build log contains (`r.assertLogContains`) the expected output from `PowerAssertionError` (at least the number `4` somewhere).

src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
```diff
@@ -53,6 +54,19 @@ public ErrorAction(Throwable error) {
      * an equivalent that captures the same details but serializes nicely.
      */
     private boolean isUnserializableException(Throwable error) {
+        if (error == null) {
+            // This shouldn't happen.
+            return false;
+        }
```
ikedam: I can expect that `error` never be `null` as `ErrorAction` is always called like
```
if (outcome.isFailure()) {
    head.addAction(new ErrorAction(outcome.getAbnormal()));
}
```

On the other hand, I can't mark `error` with `NonNull` annotation as `getAbnormal` can be `null` if `isFailure()` is `false`
and static code analysis by FindBug can't know `error` never be `null` in this context.

PR #19, opened 2016-10-03
[JENKINS-38640] Use a GREEDY RemoteInputStream

andresrc:
See [JENKINS-38640] (https://issues.jenkins-ci.org/browse/JENKINS-38640)

Changed from `untarFrom` to `untar` as the former [does not use](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/FilePath.java#L721) a GREEDY stream (maybe that should be changed in core), which should provide better performance.

Tested using [this tests](https://github.com/jenkinsci/workflow-basic-steps-plugin/blob/master/src/test/java/org/jenkinsci/plugins/workflow/support/steps/stash/StashTest.java)

@reviewbybees esp. @jglick

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
@@ -119,13 +119,8 @@ public class StashManager {
         if (!storage.isFile()) {
             throw new AbortException("No such saved stash ‘" + name + "’");
         }
-        InputStream is = new FileInputStream(storage);
-        try {
-            workspace.untarFrom(is, FilePath.TarCompression.GZIP);
-            // currently nothing to print; listener is a placeholder
-        } finally {
-            is.close();
-        }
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
🐝 , but I'm not sure it changes the performance tremendously. https://github.com/jenkinsci/external-workspace-manager-plugin/blob/master/README.md could be a workaround (just pass entire workspace instead of stashing particular files and getting overhead on the Jenkins side)

andresrc:
Yes, sorry, let me elaborate:

* Regarding the tests, the intended meaning was that the existing tests in the `workflow-basic-steps` were used to check for no regressions.
* Regarding performance, in the following screenshot (Jenkins 2.19.1), builds 2-4 are before the change and 3-6 after. The stash operations create and stash a 200 MB file, one full of zeros and the other with random content.
greedy_test

* I'll file the PR to core for `FilePath.untarFrom`. I was waiting for feedback on this one first.

Thanks!

PR #20, opened 2016-10-05
Ensure that for SimpleBlockVisitor we find the REAL last branch for a parallel [JENKINS-38536]

svanoort:
... not just the last one declared on that parallel block.

Solves [JENKINS-38536](https://issues.jenkins-ci.org/browse/JENKINS-38536).

@reviewbybees especially @jglick

- [x] Integrate nits from https://github.com/jenkinsci/workflow-api-plugin/pull/25

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
@@ -27,6 +27,7 @@ package org.jenkinsci.plugins.workflow.graphanalysis;
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
 import org.jenkinsci.plugins.workflow.actions.ThreadNameAction;
+import org.jenkinsci.plugins.workflow.actions.TimingAction;
 import org.jenkinsci.plugins.workflow.graph.BlockEndNode;
 import org.jenkinsci.plugins.workflow.graph.BlockStartNode;
 import org.jenkinsci.plugins.workflow.graph.FlowEndNode;
@@ -553,11 +554,40 @@ public class ForkScanner extends AbstractFlowScanner {
         scanner.visitSimpleChunks(visitor, finder);
     }
 
-    /** Walk through flows  */
+    @CheckForNull
+    private static FlowNode findLastStartedNode(@Nonnull List<FlowNode> candidates) {
+        if (candidates.size() == 0) {
+            return null;
+        } else if (candidates.size() == 1) {
+            return candidates.get(0);
+        } else {
+            FlowNode returnOut = candidates.get(0);
+            long startTime = Long.MIN_VALUE;
+            for(FlowNode f : candidates) {
+                TimingAction ta = f.getAction(TimingAction.class);
+                if (ta != null) {
+                    long myStart = ta.getStartTime();
+                    if (myStart > startTime) {
+                        returnOut = f;
+                        startTime = myStart;
+                    }
+                }
+            }
+            return returnOut;
+        }
+    }
+
+    /** Walk through flows */
     public void visitSimpleChunks(@Nonnull SimpleChunkVisitor visitor, @Nonnull ChunkFinder finder) {
         FlowNode prev = null;
         if (finder.isStartInsideChunk() && hasNext()) {
-            visitor.chunkEnd(this.myNext, null, this);
+            if (currentParallelStart == null ) {
+                visitor.chunkEnd(this.myNext, null, this);
+            } else { // Last node is the last started branch
+                List<FlowNode> branchEnds = new ArrayList<FlowNode>(currentParallelStart.unvisited);
+                branchEnds.add(this.myNext);
+                visitor.chunkEnd(this.findLastStartedNode(branchEnds), null, this);
+            }
         }
         while(hasNext()) {
             prev = (myCurrent != myNext) ? myCurrent : null;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -565,12 +569,11 @@ private static FlowNode findLastStartedNode(@Nonnull List<FlowNode> candidates)
             long startTime = Long.MIN_VALUE;
             for(FlowNode f : candidates) {
                 TimingAction ta = f.getAction(TimingAction.class);
-                if (ta != null) {
-                    long myStart = ta.getStartTime();
-                    if (myStart > startTime) {
-                        returnOut = f;
-                        startTime = myStart;
-                    }
+                // Null timing with multiple heads is probably a node where the GraphListener hasn't fired to add TimingAction yet
```
svanoort: See also https://github.com/jenkinsci/workflow-support-plugin/pull/16 which fixes the SemaphoreStep issue from [JENKINS-38737](https://issues.jenkins-ci.org/browse/JENKINS-38737)

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -553,11 +554,49 @@ public static void visitSimpleChunks(@Nonnull Collection<FlowNode> heads, @Nonnu
...
+            return null;
+        } else if (candidates.size() == 1) {
+            return candidates.get(0);
+        } else {
+            FlowNode returnOut = candidates.get(0);
+            long startTime = Long.MIN_VALUE;
+            for(FlowNode f : candidates) {
+                TimingAction ta = f.getAction(TimingAction.class);
+                // Null timing with multiple heads is probably a node where the GraphListener hasn't fired to add TimingAction yet
+                long myStart = (ta == null) ? System.currentTimeMillis() : ta.getStartTime();
```
oleg-nenashev: Not an ideal presumption IMHO. One buggy node without TimingAction for a COMPLETED Pipeline will always make the node to be returned by this method. Could we just skip such nodes? 🐜

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -536,4 +539,62 @@ public void testTripleParallel() throws Exception {
         Assert.assertTrue(pbs.unvisited.contains(exec.getNode("8")));
         Assert.assertTrue(pbs.unvisited.contains(exec.getNode("9")));
     }
+
+    private void testParallelFindsLast(WorkflowJob job, String semaphoreName) throws Exception {
```
oleg-nenashev: /s/test/assert/

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
svanoort: @oleg-nenashev Agreed, I'm not a huge fan of having to make this assumption either -- looking at hardening up TimingAction addition (other PR, and a bit more investigation) because it's been a pain point.

However we *can't* safely skip these nodes either, because if present they are likely to be the most recently created node (there's a tiny window where it *might* be possible to have a brand new node without the action, in which case it would be the newest one -- just about impossible to test though because it would require precise timing).   Note that it's quite possible (and likely) that the action would be attached between this code block and the next time we poke the FlowNode.

In any case: it's likely a rare/never situation barring the issue noted with Semaphore steps, and I can't think of a more robust way to handle it.

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -536,4 +539,62 @@ public void testTripleParallel() throws Exception {
...
+    }
+
+    @Issue("JENKINS-38536")
+    @Test
+    public void testParallelCorrectEndNodeForVisitor() throws Exception {
+        // Verify that SimpleBlockVisitor actually gets the *real* last node not just the last declared branch
+        WorkflowJob jobPauseFirst = r.jenkins.createProject(WorkflowJob.class, "PauseFirst");
+        jobPauseFirst.setDefinition(new CpsFlowDefinition("" +
+                "stage 'primero'\n" +
+                "parallel 'wait' : {sleep 1; semaphore 'wait1';}, \n" +
```
jglick: Generally you should not need to use `sleep` steps if you are using `semaphore` correctly.

svanoort: Confirms what I thought but was being overly cautious about, thanks!

svanoort:
@jglick The general iteration algorithms (Bismuth level 1) don't care about timing, but the SimpleBlockVisitor (aka Bismuth Level 2, or the SAX-like APIs) is responsible for returning first and last nodes so it needs to impose a higher-level ordering for in-progress parallels.  That's what is generally coupled to the Timing/Status APIs in pipeline graph analysis plugin.

svanoort:
@reviewbybees done

I'm going to do a quick integration re-test and then merge/release (after checking the additional changes since last release).

jglick:
> The general iteration algorithms (Bismuth level 1)

Hmm, I guess I was not aware of the distinction. Obviously it is too late to use distinct packages for these, but perhaps `package-info.java` could clarify this division, and generally group the API types into logical units?

svanoort:
Blocking: initial integration test is showing that there additional issues to be resolved here (though this definitely resolves one of them).

svanoort:
Would you believe that the fix from #25 makes this one a complete solution to the bug?

Already reviewed, time to merge for pending release (rolling in a few more microfixes).

PR #21, opened 2016-10-06
Create & use optimized getAction API for FlowNodes that only uses nontransient actions

svanoort:
Implement [JENKINS-38867](https://issues.jenkins-ci.org/browse/JENKINS-38867), which should improve performance significantly (calls to getAction were a significant fraction of flow analysis time, in profiling). 

After extensive benchmarking to find out performance impacts on stage view (the place that feels this pain the worst), optimizations here improve performance 2-3x when doing the first run analysis.

Action items:

- [x] Create a new PersistentAction marker interface
- [x] Set up the getAction(Class) method to look if the action is a PersistentAction, and if so, not inspect transientactionfactories, otherwise fall back to that
- [x] Create a getPersistentAction method on flownode
- [x] Optimizations to the use of the actionable list and its iterator
- [x] Threading stuffs
- [x] Switch the GraphAnalysis APIs to consume getPersistentAction
- [x] Switch internal pipeline APIs to use it too (almost all the time where we're inspecting nodes)
- [x] Various painful micro-optimizations which offer another ~10% performance gain and probably in other places (worth it because the API is hit hard in pipeline)
- [x] Set up workflow support downstream PR to mark PauseAction as a PersistentAction - https://github.com/jenkinsci/workflow-support-plugin/pull/17
- [x] Benchmark the difference to confirm impact (in progress). - 50-66% reduction in run handling for stage view.

Related to https://github.com/jenkinsci/jenkins/pull/2582 but pipeline specific and does not require core bump.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-api</artifactId>
-    <version>2.5-optimized2-SNAPSHOT</version>
+    <version>2.5-optimized3-SNAPSHOT</version>
     <packaging>hpi</packaging>
     <name>Pipeline: API</name>
     <url>https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+API+Plugin</url>
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-api</artifactId>
-    <version>2.5-SNAPSHOT</version>
+    <version>2.5-optimized-SNAPSHOT</version>
     <packaging>hpi</packaging>
     <name>Pipeline: API</name>
     <url>https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+API+Plugin</url>
--- a/pom.xml
+++ b/pom.xml
@@ -33,6 +33,7 @@
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-api</artifactId>
+    <!-- Yes, this is temporary for benchmarking etc, will be reverted before merge -->
     <version>2.5-optimized3-SNAPSHOT</version>
     <packaging>hpi</packaging>
     <name>Pipeline: API</name>
--- a/pom.xml
+++ b/pom.xml
@@ -33,8 +33,7 @@
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-api</artifactId>
-    <!-- Yes, this is temporary for benchmarking etc, will be reverted before merge -->
-    <version>2.5-optimized3-SNAPSHOT</version>
+    <version>2.5-SNAPSHOT</version>
     <packaging>hpi</packaging>
     <name>Pipeline: API</name>
     <url>https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+API+Plugin</url>
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-api</artifactId>
-    <version>2.5-optimized-SNAPSHOT</version>
+    <version>2.5-optimized2-SNAPSHOT</version>
     <packaging>hpi</packaging>
     <name>Pipeline: API</name>
     <url>https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+API+Plugin</url>
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/PersistentAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/PersistentAction.java
@@ -3,8 +3,8 @@ package org.jenkinsci.plugins.workflow.actions;
 import hudson.model.Action;
 
 /**
- * Marker interface for an action that can't be contributed by a {@link jenkins.model.TransientActionFactory}
- * Actions implementing this can use more efficient getAction methods
+ * This is a marker interface for an action that can't be contributed by a {@link jenkins.model.TransientActionFactory}.
+ * Actions implementing this can use more efficient {@link hudson.model.Actionable#getAction(Class)} variant
  */
 public interface PersistentAction extends Action {
 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/PersistentAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/PersistentAction.java
@@ -1,0 +1,10 @@
+package org.jenkinsci.plugins.workflow.actions;
+
+import hudson.model.Action;
+
+/**
+ * Marker interface for an action that can't be contributed by a {@link jenkins.model.TransientActionFactory}
+ * Actions implementing this can use more efficient getAction methods
+ */
+public interface PersistentAction extends Action {
+}
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/PersistentAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/PersistentAction.java
@@ -4,7 +4,8 @@ import hudson.model.Action;
 
 /**
  * This is a marker interface for an action that can't be contributed by a {@link jenkins.model.TransientActionFactory}.
- * Actions implementing this can use more efficient {@link hudson.model.Actionable#getAction(Class)} variant
+ * Actions implementing this can use more efficient {@link org.jenkinsci.plugins.workflow.graph.FlowNode#getPersistentAction(Class)}
+ *   and {@link org.jenkinsci.plugins.workflow.graph.FlowNode#getAction(Class)} internally delegates to that
  */
 public interface PersistentAction extends Action {
 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
@@ -25,6 +25,7 @@
 package org.jenkinsci.plugins.workflow.graph;
 
 import com.google.common.collect.ImmutableList;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.model.Action;
 import hudson.model.Actionable;
 import hudson.model.BallColor;
@@ -248,18 +249,25 @@ public abstract class FlowNode extends Actionable implements Saveable {
      */
     protected synchronized void setActions(List<Action> actions) {
             this.actions = new CopyOnWriteArrayList<Action>(actions);
+//            this._wrapper = new ListWrapper(this);
+    }
+
+    private <T extends Action> T getDirectAction(Class<T> type) {
+        if (actions == null) {
+            getActions();
+        }
+        for (Action a : actions) {
+            if (type.isInstance(a)) {
+                return type.cast(a);
+            }
+        }
+        return null;
     }
 
     @Override
     public <T extends Action> T getAction(Class<T> type) {
-        if (type.isAssignableFrom(PersistentAction.class)) {
-            List<Action> acts = (actions == null) ? getActions() : actions;
-            for (Action a : acts) {
-                if (type.isInstance(a)) {
-                    return type.cast(a);
-                }
-            }
-            return null;
+        if (PersistentAction.class.isAssignableFrom(type)) {
+            return getDirectAction(type);
         } else {
             for (Action a : getAllActions()) {
                 if (type.isInstance(a)) {
@@ -269,68 +277,95 @@ public abstract class FlowNode extends Actionable implements Saveable {
             return null;
         }
     }
-/*
+
+    /*
     We can't use Actionable#actions to store actions because they aren't transient,
-    and we need to store actions elsewhere because this is the only mutable pat of FlowNode.
+    and we need to store actions elsewhere because this is the only mutable part of FlowNode.
 
     So we create a separate transient field and store List of them there, and intercept every mutation.
- */
+    */
+    private static class ListWrapper extends AbstractList<Action> {
+        FlowNode node;
+        CopyOnWriteArrayList<Action> myActions;
+
+        ListWrapper(FlowNode n) {
+            myActions = n.actions;
+            node = n;
+        }
+
+        @Override
+        public Action get(int index) {
+            return myActions.get(index);
+        }
+
+        @Override
+        public void add(int index, Action element) {
+            myActions.add(index, element);
+            node.persistSafe();
+        }
+
+        @Override
+        public boolean add(Action value) {
+            myActions.add(value);
+            node.persistSafe();
+            return true;
+        }
+
+        @Override
+        public Iterator<Action> iterator() {
+            return myActions.iterator();
+        }
+
+        @Override
+        public Action remove(int index) {
+            Action old = myActions.remove(index);
+            node.persistSafe();
+            return old;
+        }
+
+        @Override
+        public Action set(int index, Action element) {
+            Action old = myActions.set(index, element);
+            node.persistSafe();
+            return old;
+        }
+
+        @Override
+        public int size() {
+            return myActions.size();
+        }
+    }
+
+    private synchronized void loadActions() {
+        if (actions != null) {
+            return; //Mutation while we acquired lock
+        }
+        try {
+            actions = new CopyOnWriteArrayList<Action>(exec.loadActions(this));
+        } catch (IOException e) {
+            LOGGER.log(WARNING, "Failed to load actions for FlowNode id=" + id, e);
+            actions = new CopyOnWriteArrayList<Action>();
+        }
+    }
+
     @Exported
     @Override
-    public synchronized List<Action> getActions() {
-                if (actions==null) {
-                    try {
-                        actions = new CopyOnWriteArrayList<Action>(exec.loadActions(this));
-                    } catch (IOException e) {
-                        LOGGER.log(WARNING, "Failed to load actions for FlowNode id=" + id, e);
-                        actions = new CopyOnWriteArrayList<Action>();
-                    }
-                }
+    @SuppressFBWarnings(value = "UG_SYNC_SET_UNSYNC_GET", justification = "CopyOnWrite ArrayList, and field modification is synchronized")
+    public List<Action> getActions() {
+        if (actions==null) {
+            loadActions();
+        }
+        return new ListWrapper(this);
+    }
 
-        return new AbstractList<Action>() {
-            @Override
-            public Action get(int index) {
-                return actions.get(index);
-            }
-
-            @Override
-            public void add(int index, Action element) {
-                actions.add(index, element);
-                persist();
-            }
-
-            @Override
-            public Iterator<Action> iterator() {
-                return actions.iterator();
-            }
-
-            @Override
-            public Action remove(int index) {
-                Action old = actions.remove(index);
-                persist();
-                return old;
-            }
-
-            @Override
-            public Action set(int index, Action element) {
-                Action old = actions.set(index, element);
-                persist();
-                return old;
-            }
-
-            @Override
-            public int size() {
-                return actions.size();
-            }
-
-            private void persist() {
-                try {
-                    save();
-                } catch (IOException e) {
-                    LOGGER.log(WARNING, "failed to save actions for FlowNode id=" + id, e);
-                }
-            }
-        };
+    @Override
+    public void addAction(@Nonnull Action a) {
+        if (a==null) throw new IllegalArgumentException();
+        if (actions == null) {
+            loadActions();
+        }
+        actions.add(a);
+        persistSafe();
     }
 
     /**
@@ -339,6 +374,15 @@ public abstract class FlowNode extends Actionable implements Saveable {
      */
     public void save() throws IOException {
         exec.saveActions(this, actions);
+    }
+
+    // Persist, handling possible IOException
+    private void persistSafe() {
+        try {
+            save();
+        } catch (IOException e) {
+            LOGGER.log(WARNING, "failed to save actions for FlowNode id=" + this.id, e);
+        }
     }
 
     @Override
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
@@ -248,6 +248,26 @@ public abstract class FlowNode extends Actionable implements Saveable {
         this.actions = new CopyOnWriteArrayList<Action>(actions);
     }
 
+
+    // Much faster check of actions because it does not trigger the TransientActionFactory
+    @Override
+    public <T extends Action> T getAction(Class<T> type) {
+        // Need to check and see if this will cause issues -- better to change in place if we can
+        // Otherwise we need to do changes in many places to use a new, fast API
+        List<Action> acts = (actions == null) ? getActions() : actions;
+        for (Action a : acts) {
+            if (type.isInstance(a)) {
+                return type.cast(a);
+            }
+        }
+        return null;
+    }
+
+    // Replaces the legacy getAction method for the few cases we really need it
+    public <T extends Action> T getActionComplete(Class<T> type) {
+        return super.getAction(type);
+    }
+
 /*
     We can't use Actionable#actions to store actions because they aren't transient,
     and we need to store actions elsewhere because this is the only mutable pat of FlowNode.
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
@@ -49,6 +49,7 @@ import org.jenkinsci.plugins.workflow.flow.FlowExecution;
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.DoNotUse;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.export.Exported;
 import org.kohsuke.stapler.export.ExportedBean;
 
@@ -249,17 +250,19 @@ public abstract class FlowNode extends Actionable implements Saveable {
      */
     protected synchronized void setActions(List<Action> actions) {
             this.actions = new CopyOnWriteArrayList<Action>(actions);
-//            this._wrapper = new ListWrapper(this);
     }
 
     /**
-     * Return the first persistent action on the FlowNode, without consulting TransientActionFactories
-     * Used here because it is much faster to invoke a less-conditional, monomorphic method (50% faster)
+     * Return the first nontransient {@link Action} on the FlowNode, without consulting {@link jenkins.model.TransientActionFactory}s
+     * <p> This is not restricted to just Actions implementing {@link PersistentAction} but usually they should.
+     * Used here because it is much faster than base {@link ##getAction(Class)} method.
      * @param type Class of action
      * @param <T>  Action type
-     * @return First persistent action or null if not found
+     * @return First nontransient action or null if not found.
      */
-    public <T extends Action> T getPersistentAction(@Nonnull Class<T> type) {
+    @CheckForNull
+    @Restricted(NoExternalUse.class)  // Limit use to workflow-api packages until we have a case where we need the performance badly.
+    public final <T extends Action> T getPersistentAction(@Nonnull Class<T> type) {
         if (actions == null) {
             loadActions();
         }
@@ -282,6 +285,7 @@ public abstract class FlowNode extends Actionable implements Saveable {
     }
 
     @Override
+    @CheckForNull
     public <T extends Action> T getAction(Class<T> type) {
         // Delegates internally to methods that are not overloads, which are more subject to inlining and optimization
         // Normally a micro-optimization, but these methods are invoked *heavily* and improves performance 5%
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
@@ -249,9 +249,13 @@ public abstract class FlowNode extends Actionable implements Saveable {
     }
 
 
-    // Much faster check of actions because it does not trigger the TransientActionFactory
-    @Override
-    public <T extends Action> T getAction(Class<T> type) {
+    /**
+     * Return a persistent action o(transient actions are not included)
+     * @param type Action type to look for
+     * @param <T> Action type to look for
+     * @return First attached action of given type, or null if none found
+     */
+    public <T extends Action> T getDirectAction(Class<T> type) {
         // Need to check and see if this will cause issues -- better to change in place if we can
         // Otherwise we need to do changes in many places to use a new, fast API
         List<Action> acts = (actions == null) ? getActions() : actions;
@@ -261,11 +265,6 @@ public abstract class FlowNode extends Actionable implements Saveable {
             }
         }
         return null;
-    }
-
-    // Replaces the legacy getAction method for the few cases we really need it
-    public <T extends Action> T getActionComplete(Class<T> type) {
-        return super.getAction(type);
     }
 
 /*
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
@@ -104,7 +104,7 @@ public abstract class FlowNode extends Actionable implements Saveable {
      * This is just a convenience method.
      */
     public final @CheckForNull ErrorAction getError() {
-        return getAction(ErrorAction.class);
+        return getDirectAction(ErrorAction.class);
     }
 
     public @Nonnull FlowExecution getExecution() {
@@ -167,7 +167,7 @@ public abstract class FlowNode extends Actionable implements Saveable {
 
     @Exported
     public String getDisplayName() {
-        LabelAction a = getAction(LabelAction.class);
+        LabelAction a = getDirectAction(LabelAction.class);
         if (a!=null)    return a.getDisplayName();
         else            return getTypeDisplayName();
     }
@@ -177,7 +177,7 @@ public abstract class FlowNode extends Actionable implements Saveable {
         if (functionName == null) {
             return getDisplayName();
         } else {
-            LabelAction a = getAction(LabelAction.class);
+            LabelAction a = getDirectAction(LabelAction.class);
             if (a != null) {
                 return functionName + " (" + a.getDisplayName() + ")";
             } else {
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
@@ -107,7 +107,7 @@ public abstract class FlowNode extends Actionable implements Saveable {
      * This is just a convenience method.
      */
     public final @CheckForNull ErrorAction getError() {
-        return getAction(ErrorAction.class);
+        return getDirectAction(ErrorAction.class);
     }
 
     public @Nonnull FlowExecution getExecution() {
@@ -170,7 +170,7 @@ public abstract class FlowNode extends Actionable implements Saveable {
 
     @Exported
     public String getDisplayName() {
-        LabelAction a = getAction(LabelAction.class);
+        LabelAction a = getDirectAction(LabelAction.class);
         if (a!=null)    return a.getDisplayName();
         else            return getTypeDisplayName();
     }
@@ -180,7 +180,7 @@ public abstract class FlowNode extends Actionable implements Saveable {
         if (functionName == null) {
             return getDisplayName();
         } else {
-            LabelAction a = getAction(LabelAction.class);
+            LabelAction a = getDirectAction(LabelAction.class);
             if (a != null) {
                 return functionName + " (" + a.getDisplayName() + ")";
             } else {
@@ -252,7 +252,7 @@ public abstract class FlowNode extends Actionable implements Saveable {
 //            this._wrapper = new ListWrapper(this);
     }
 
-    private  <T extends Action> T getDirectAction(Class<T> type) {
+    public <T extends Action> T getDirectAction(Class<T> type) {
         if (actions == null) {
             loadActions();
         }
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
@@ -107,7 +107,7 @@ public abstract class FlowNode extends Actionable implements Saveable {
      * This is just a convenience method.
      */
     public final @CheckForNull ErrorAction getError() {
-        return getDirectAction(ErrorAction.class);
+        return getPersistentAction(ErrorAction.class);
     }
 
     public @Nonnull FlowExecution getExecution() {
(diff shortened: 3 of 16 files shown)


svanoort:
@jglick Where, specifically is it a problem?  I am aware of the other PR, and have reviewed it -- this is a workflow-specific optimization.

svanoort:
@jglick To be clear: the question I am asking is "when we use transient actions, are we fetching via getAction or via getAllActions"?

Meh, maybe better to take the safer approach of introducing a new, fast API though that means every consumer has to be updated in order to use it.

jglick:
> when we use transient actions, are we fetching via getAction or via getAllActions

Either.

> that means every consumer has to be updated in order to use it

In general they cannot. The whole point of `TransientActionFactory` is that external plugins can affect actions. So any client which is asking for an action could not use such a proposed call unless you could prove that no such factory could be affecting the result. So this seems like a dead end.

jglick:
Now if it helps, https://github.com/jenkinsci/jenkins/pull/2582 *could* be further developed, for example to allow `getAction(Class)` to return a non-null result without consulting transient factories if the `actions` field already has a match. (If it does not have a match, you still need to consult the factories.) No such optimization is possible for `getActions(Class)` of course.

svanoort:
@jglick You saw that I completely re-did the approach, yeah?  It's now a net-new API, so we can use it exclusively where we do not care about transient actions (most of what we do with flownodes).

jglick:
> we can use it exclusively where we do not care about transient actions

As noted above, you would have to do some work to prove in each case that we can safely ignore transient factories. I think it is more straightforward to optimize the existing semantics for the common cases.

svanoort:
> As noted above, you would have to do some work to prove in each case that we can safely ignore transient factories. I think it is more straightforward to optimize the existing semantics for the common cases.

Why would that matter when the legacy API already returns transient actions, and NOT including them is explicitly part of the new API?

svanoort:
>  I think it is more straightforward to optimize the existing semantics for the common cases.

I am 100% for optimizing the existing semantics as much as possible! 👍  and 🍻  all around for making common and critical paths as fast as possible.  Something I'd also like to contribute to, once this part is included and benchmarked.

This change is necessary because:

1. For pipeline almost all the cases just care about directly attached actions on flownodes, since we're using the to store the execution info.
2. Performance for pipelines & their visualization is already becoming rather critical, and this lets us sidestep a lot of unavoidable complexity for big performance wins.
3. We have an existing API that covers both transient and nontransient actions
4. Optimizations to the exiting API are a core change, which means anywhere from 3 to 12 months before people see the benefits.

jglick:
> almost all the cases

I suppose it depends on which particular cases you are referring to.

> Optimizations to the exiting API are a core change, which means anywhere from 3 to 12 months before people see the benefits.

Not necessarily. If you are mainly interested in `Actionable.getAction(Class)`, this is not `final`, so it would be straightforward to port any core optimization to `FlowNode` as well, with a comment in place to delete the override once the core baseline becomes sufficiently new.

svanoort:
**Benchmarks:**

Baseline

Result "stageViewBenchmark":
  28.817 ±(99.9%) 3.369 ms/op [Average]
  (min, avg, max) = (27.863, 28.817, 29.593), stdev = 0.875
  CI (99.9%): [25.449, 32.186] (assumes normal distribution)


# Run complete. Total time: 00:10:51

Benchmark                              Mode  Cnt   Score   Error  Units
StageViewBenchmark.stageViewBenchmark  avgt    5  28.817 ± 3.369  ms/op

Process finished with exit code 0

**Version info**

Jenkins version: 2.7.3
ace-editor:1.0.1
branch-api:1.10
cloudbees-folder:5.10
credentials:1.16.1
durable-task:1.12
git-client:1.11.0
git-server:1.5
handlebars:1.1
jquery-detached:1.2.1
mailer:1.13
momentjs:1.1
pipeline-build-step:2.2
pipeline-graph-analysis:1.1
pipeline-input-step:2.1
pipeline-rest-api:2.1
pipeline-stage-step:2.2
pipeline-stage-view:2.0
scm-api:1.1
script-security:1.21
ssh-credentials:1.10
structs:1.3
workflow-aggregator:2.3
workflow-api:2.2
workflow-basic-steps:2.1
workflow-cps:2.13
workflow-cps-global-lib:2.2
workflow-durable-task-step:2.4
workflow-job:2.6
workflow-multibranch:2.8
workflow-scm-step:2.2
workflow-step-api:2.3
workflow-support:2.2

**Benchmark methodology:**

For each benchmark: 

1. Delete the benchmark job
2. Run below pipeline 10x, then flush the flownode cache for all runs (using reflection to get at it and call invalidateAll), then invalidate the RunsCache in the stage view, then do GC
3. Benchmark method: fetch runs via http from the job, using the URL http://localhost:8080/jenkins/job/benchmarkPipeline/wfapi/runs?fullStages=true and throwing away content but measuring and returning its size in bytes to prevent dead code elimination
4.  Teardown per-invocation is to delete the project again. 
5. Benchmark runs in JMH mode AverageTime, 1 warmup iteration of 5 minutes, 5 measurement iterations of 60 seconds, shouldDoGC=true, 1 thread and 1 fork

```java
"for (int i=0; i<15; i++) {\n" +
                    "    stage \"stage $i\" \n" +
                    "    echo \"ran my stage is $i\"        \n" +
                    "    node {\n" +
                    "        echo 'whoami';\n" +
                    "    }\n" +
                    "}\n" +
                    "node {sh'whoami';} \n"+
                    "\n" +
                    "stage 'label based'\n" +
                    "echo 'wait for executor'\n" +
                    "node {\n" +
                    "    stage 'things using node'\n" +
                    "    for (int i=0; i<200; i++) {\n" +
                    "        echo \"we waited for this $i seconds\"    \n" +
                    "    }\n" +
                    "}"
```

svanoort:
**Benchmark, pulling in the new getDirectAction API**
(In stage view, pipeline graph analysis, and workflow-api methods):

Result "stageViewBenchmark":
  28.495 ±(99.9%) 7.720 ms/op [Average]
  (min, avg, max) = (26.827, 28.495, 31.872), stdev = 2.005
  CI (99.9%): [20.775, 36.214] (assumes normal distribution)


\# Run complete. Total time: 00:10:49

Benchmark                              Mode  Cnt   Score   Error  Units
StageViewBenchmark.stageViewBenchmark  avgt    5  28.495 ± 7.720  ms/op

Process finished with exit code 0

Jenkins version: 2.7.3
ace-editor:1.0.1
branch-api:1.10
cloudbees-folder:5.10
credentials:1.16.1
durable-task:1.12
git-client:1.11.0
git-server:1.5
handlebars:1.1
jquery-detached:1.2.1
mailer:1.13
momentjs:1.1
pipeline-build-step:2.2
pipeline-graph-analysis:1.3-optimized-SNAPSHOT (private-91c8b6b7-svanoort)
pipeline-input-step:2.1
pipeline-rest-api:2.2-optimized-SNAPSHOT (private-ebfdb682-svanoort)
pipeline-stage-step:2.2
pipeline-stage-view:2.0
scm-api:1.1
script-security:1.21
ssh-credentials:1.10
structs:1.3
workflow-aggregator:2.3
workflow-api:2.5-optimized-SNAPSHOT (private-722f9209-svanoort)
workflow-basic-steps:2.1
workflow-cps:2.13
workflow-cps-global-lib:2.2
workflow-durable-task-step:2.4
workflow-job:2.6
workflow-multibranch:2.8
workflow-scm-step:2.2
workflow-step-api:2.3
workflow-support:2.2

... unfortunately just not the numbers we want

svanoort:
AH!  Profiling found the issue: getAction is used in TimingAction.getStartTime, and also in NotExecutedNodeAction.isExecuted(Flownode f), plus other bits in core API. 

When I revise to not include the methods, execution time drops *a bit*:

**New Benchmark:**
11.469 ms/op


Result "stageViewBenchmark":
  12.681 ±(99.9%) 4.269 ms/op [Average]
  (min, avg, max) = (11.469, 12.681, 14.334), stdev = 1.109
  CI (99.9%): [8.411, 16.950] (assumes normal distribution)


\# Run complete. Total time: 00:11:02

Benchmark                              Mode  Cnt   Score   Error  Units
StageViewBenchmark.stageViewBenchmark  avgt    5  12.681 ± 4.269  ms/op

**Versions:**

Jenkins version: 2.7.3
ace-editor:1.0.1
branch-api:1.10
cloudbees-folder:5.10
credentials:1.16.1
durable-task:1.12
git-client:1.11.0
git-server:1.5
handlebars:1.1
jquery-detached:1.2.1
mailer:1.13
momentjs:1.1
pipeline-build-step:2.2
pipeline-graph-analysis:1.3-optimized-SNAPSHOT (private-91c8b6b7-svanoort)
pipeline-input-step:2.1
pipeline-rest-api:2.2-optimized-SNAPSHOT (private-77c27f8a-svanoort)
pipeline-stage-step:2.2
pipeline-stage-view:2.0
scm-api:1.1
script-security:1.21
ssh-credentials:1.10
structs:1.3
workflow-aggregator:2.3
workflow-api:2.5-optimized-SNAPSHOT (private-e4c55a2b-svanoort)
workflow-basic-steps:2.1
workflow-cps:2.13
workflow-cps-global-lib:2.2
workflow-durable-task-step:2.4
workflow-job:2.6
workflow-multibranch:2.8
workflow-scm-step:2.2
workflow-step-api:2.3
workflow-support:2.2

svanoort:
NOTE on the high error bars for benchmarks:  this is probably tied to JIT compilation - however we don't want to allow for full JIT compilation of all once-per-run methods, because that would require several thousand pipeline executions (which might be days), all visualized. 

In the real world, many of these requests will simply hit caches as well, so methods are far less exercised.

svanoort:
@reviewbybees especially @jglick  - key point here: look at getActions in FlowNode class to see why this is not just safe but also smart.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -248,6 +248,25 @@ protected void setActions(List<Action> actions) {
         this.actions = new CopyOnWriteArrayList<Action>(actions);
     }
 
+
+    /**
+     * Return a persistent action o(transient actions are not included)
```
jglick: `o`?

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -248,6 +248,25 @@ protected void setActions(List<Action> actions) {
...
+     * @param type Action type to look for
+     * @param <T> Action type to look for
+     * @return First attached action of given type, or null if none found
+     */
+    public <T extends Action> T getDirectAction(Class<T> type) {
+        // Need to check and see if this will cause issues -- better to change in place if we can
+        // Otherwise we need to do changes in many places to use a new, fast API
+        List<Action> acts = (actions == null) ? getActions() : actions;
```
jglick: :bug: You must call `getActions()` to avoid losing a memory barrier.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
svanoort: ya rly.  (will fix)

svanoort:
> This is clearly a hack and not something we want in the long term, yet it is being baked into the API.

@jglick  Vague.  What *specifically* is your objection to it?  Bear in mind that we *already* have a getActions API that does this, just not a single-action equivalent. This PR is merely filling that gap. 

Would you prefer I can rewrite every getDirectAction method to call the existing getAction method and run a for loop with filtering (duplicating this exact logic all over the codebase)?

> Revise TransientActionFactory to allow it to specify not just what kind of context it takes, but what kind of action it produces,

Why would we be touching TransientActionFactory at all for this?

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
svanoort: @jglick It's a CopyOnWriteArraylist.

svanoort:
@jglick This specifically addresses the lack of a getAction API corresponding to getActions.  While it's worth optimizing for the TransientActionFactories, trying to optimize and optimize that will never be as efficient as not handling them at all for the cases where we don't need to worry about them.  

I am also still waiting to see the benchmarks for your optimizations there, as I have done here.  Unless you can prove that the TransientActionFactory optimizations somehow manage to make it much faster than what we have here (impossible, AFAIK), we need this and should have done it ages ago.

jglick:
> we _already_ have a `getActions` API that does this

No, `getActions(Class)` delegates to `getAllActions`, thus honoring the `TransientActionFactory` contract.

jglick:
> What _specifically_ is your objection to it?

The fact that this is deliberately bypassing the usual contract of `TransientActionFactory`—that it will be consulted in any case where the set of actions associated with a context is considered—as a workaround for the fact the current design of that SPI does not allow those factories to indicate cases where they are not relevant.

> Would you prefer I can rewrite every `getDirectAction` method to call the existing `getAction` method and run a for loop with filtering

Makes no sense at all. Then you would pay the performance cost but still lose compatibility.

> Why would we be touching TransientActionFactory at all for this?

Because that is the source of the overhead as I understand it.

jglick:
> trying to optimize and optimize that will never be as efficient as not handling them at all for the cases where we don't need to worry about them

And that is exactly the problem I am trying to solve.

jglick:
> also still waiting to see the benchmarks for your optimizations there

Obviously I am not going to duplicate work by setting up a benchmark myself when you already have one. I can just propose code changes which could be tested by the same system.

svanoort:
@jglick I honestly am confused why this is so contentious for you, since it's a complimentary and logical extension to your work in Jenkins Core on TransientActionFactories -- unless you continue to argue that the core work is competing.  If you argue it's competing, then the burden of proof is on you to provide a similarly rigorous benchmark case to prove this is unnecessary.  If it's complimentary then the core work only needs the most trivial of performance smoketests.

> No, getActions(Class) delegates to getAllActions, thus honoring the TransientActionFactory contract.

https://github.com/jenkinsci/workflow-api-plugin/blob/master/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java#L259

> The fact that this is deliberately bypassing the usual contract of TransientActionFactory

It's a net-new API with an intentionally different name - why would it hold to the contract of a completely different API?  I took this approach  specifically because you objected to the in-place modification (with some reason, and I was somewhat iffy on it too although it looks safe in the current use case).  The way we use actions with pipeline is different than the way we normally use them, so it makes sense to create a new API. 

> Makes no sense at all. Then you would pay the performance cost but still lose compatibility.

I meant to say getActions( ). 

> as a workaround for the fact the current design of that SPI does not allow those factories to indicate cases where they are not relevant

No, it's the opposite: we *don't* want to get noise from TransientActionFactories when inspecting the graph.

svanoort:
> duplicate work by setting up a benchmark myself 

I don't have any existing benchmarking of TransientActionFactory performance, and because you are the one proposing to kill my proposed change under grounds that yours makes it unnecessary, the burden of proof is on you at this point. 

If they're considered as complimentary changes (my perspective) then a rigorous benchmark is not required.

jglick:
I guess we are talking about different overloads. I thought you were referring to `getActions(Class)`. Yes we already have a `getActions()`, which is just an override of the core version that loads a persisted list. When `TransientActionFactory` was introduced, this was deprecated to discourage outside use (ideally it would be `protected`).

> The way we use actions with pipeline is different than the way we normally use them

Not fundamentally different, though on `FlowNode`s we tend to have more persisted actions and fewer transient (contributed) actions.

> we _don't_ want to get noise from `TransientActionFactory`s when inspecting the graph

Well, we would if they were able to produce results cheaply, but currently they cannot.

> it's a complimentary _[sic]_ and logical extension to your work

Yes, I agree. See my latest upstream commits.

For a shorter-term fix here, my preferences still fall between

* introducing a `PersistentAction`, so that existing calls to `getAction(Class)` on API types we have decided are not eligible for `TransientActionFactory` will quietly work faster
* restricting `getDirectAction` and copying it into the two downstream PRs (just a few lines of obvious code, not a significant maintenance burden)

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
jglick: I think I see—you are trying to avoid the wrapper `FlowNode$1`. But then it would be simpler to just override

```java
@Override
public Iterator<Action> iterator() {
    return actions.iterator();
}
```

jglick: Alternately:

```java
List<Action> acts;
synchronized (this) {
    if (actions == null) {
        getActions();
    }
    acts = actions;
}
// …
```

svanoort:
> introducing a PersistentAction, so that existing calls to getAction(Class) on API types we have decided are not eligible for TransientActionFactory will quietly work faster

Going down this path, after discussion, since it's a cleverer version of the initial approach (and more forward compatible).

svanoort:
> introducing a PersistentAction, so that existing calls to getAction(Class) on API types we have decided are not eligible for TransientActionFactory will quietly work faster

To capture out-of-band discussion, going down this path, I think.  Might be worthwhile as a complimentary core change.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
svanoort: @jglick After digging in again on this, I've done the Iterator optimization (commit pending once I finish reverting the older code and benchmarking) as a mini-optimization.  

Since the field is only modified by `getActions( )` and `setActions(List<Action> newActions)`  (and field assignment is atomic, and in the null case we invoke getActions to for the same result), the result *should* be thread safe AFAICT.

svanoort:
Not such good performance when using the PersistentAction, still (without the wrapper optimization initially, but without synchronized block around nullcheck). 

Result "stageViewBenchmark":
  18.585 ±(99.9%) 3.864 ms/op [Average]
  (min, avg, max) = (17.734, 18.585, 20.262), stdev = 1.003
  CI (99.9%): [14.722, 22.449] (assumes normal distribution)


\# Run complete. Total time: 00:12:06

Benchmark                              Mode  Cnt   Score   Error  Units
StageViewBenchmark.stageViewBenchmark  avgt    5  18.585 ± 3.864  ms/op

Process finished with exit code 0

svanoort:
# BIG BENCHMARKING FOR OPTIMIZATION EFFORTS

PersistentAction based API

37ba27ede0c7470549a067d2496beff75432b0a4

Result "stageViewBenchmark":
  19.269 ±(99.9%) 3.665 ms/op [Average]
  (min, avg, max) = (17.945, 19.269, 22.066), stdev = 1.627
  CI (99.9%): [15.603, 22.934] (assumes normal distribution)


\# Run complete. Total time: 00:13:49

Benchmark                              Mode  Cnt   Score   Error  Units
StageViewBenchmark.stageViewBenchmark  avgt    7  19.269 ± 3.665  ms/op

Process finished with exit code 0

-- using base 2.4 workflow-api (retesting):
~ 30 ms/op

Result "stageViewBenchmark":
  30.641 ±(99.9%) 4.871 ms/op [Average]
  (min, avg, max) = (28.726, 30.641, 34.850), stdev = 2.163
  CI (99.9%): [25.770, 35.511] (assumes normal distribution)


\# Run complete. Total time: 00:13:38

Benchmark                              Mode  Cnt   Score   Error  Units
StageViewBenchmark.stageViewBenchmark  avgt    7  30.641 ± 4.871  ms/op

Process finished with exit code 0

# OPTIMIZED with wrapper list class collapsed back into getActions()
Hash: a029311937ea09fc97991ad49ba6ced18a88f1e6

Result "stageViewBenchmark":
  18.811 ±(99.9%) 2.921 ms/op [Average]
  (min, avg, max) = (17.079, 18.811, 20.860), stdev = 1.297
  CI (99.9%): [15.890, 21.732] (assumes normal distribution)


\# Run complete. Total time: 00:12:44

Benchmark                              Mode  Cnt   Score   Error  Units
StageViewBenchmark.stageViewBenchmark  avgt    7  18.811 ± 2.921  ms/op

Process finished with exit code 0

# OPTIMIZED TO ALLOW FOR INLINING SOME OF THE GETACTION LOGIC:
workflow-api:2.5-optimized3-SNAPSHOT (private-f1c12e36-svanoort)
f1c12e366e2d9315828b7fb81448e0e96ed14d14

Result "stageViewBenchmark":
  17.506 ±(99.9%) 2.353 ms/op [Average]
  (min, avg, max) = (16.175, 17.506, 18.916), stdev = 1.045
  CI (99.9%): [15.153, 19.859] (assumes normal distribution)

\# Run complete. Total time: 00:14:20

Benchmark                              Mode  Cnt   Score   Error  Units
StageViewBenchmark.stageViewBenchmark  avgt    7  17.506 ± 2.353  ms/op


# Publicized getDirectAction calls used just within workflow-api, hash e9c8b47d9a86dc8b1459c48b8fa4d734e1cdd87c

Result "stageViewBenchmark":
  12.725 ±(99.9%) 3.925 ms/op [Average]
  (min, avg, max) = (11.014, 12.725, 16.276), stdev = 1.743
  CI (99.9%): [8.800, 16.651] (assumes normal distribution)


\# Run complete. Total time: 00:14:01

Benchmark                              Mode  Cnt   Score   Error  Units
StageViewBenchmark.stageViewBenchmark  avgt    7  12.725 ± 3.925  ms/op

Process finished with exit code 0

# Final optimization streak: using the getDirectAction for stage view internals and pipeline graph analysis:
Result "stageViewBenchmark":
  12.482 ±(99.9%) 2.904 ms/op [Average]
  (min, avg, max) = (10.959, 12.482, 14.772), stdev = 1.289
  CI (99.9%): [9.578, 15.386] (assumes normal distribution)


\# Run complete. Total time: 00:13:05

Benchmark                              Mode  Cnt   Score   Error  Units
StageViewBenchmark.stageViewBenchmark  avgt    7  12.482 ± 2.904  ms/op

svanoort:
Conclusion: it's enough to optimize the getPersistentAction and consume this just within workflow-api... for now.  But we want to be able to use it in other places if it comes up as a bottleneck again.

src/main/java/org/jenkinsci/plugins/workflow/actions/PersistentAction.java
```diff
@@ -0,0 +1,10 @@
+package org.jenkinsci.plugins.workflow.actions;
+
+import hudson.model.Action;
+
+/**
+ * Marker interface for an action that can't be contributed by a {@link jenkins.model.TransientActionFactory}
```
jglick: Summary sentences must end with a period.

src/main/java/org/jenkinsci/plugins/workflow/actions/PersistentAction.java
```diff
@@ -0,0 +1,10 @@
...
+ * Actions implementing this can use more efficient getAction methods
```
jglick: Either `@link` to a specific method, or replace with a generic promise that something more efficient is possible.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -244,66 +247,120 @@ public String getUrl() throws IOException {
...
-    We can't use Actionable#actions to store actions because they aren't transient,
-    and we need to store actions elsewhere because this is the only mutable pat of FlowNode.
+    /**
+     * Return the first persistent action on the FlowNode, without consulting TransientActionFactories
+     * Used here because it is much faster to invoke a less-conditional, monomorphic method (50% faster)
+     * @param type Class of action
+     * @param <T>  Action type
+     * @return First persistent action or null if not found
+     */
+    public <T extends Action> T getPersistentAction(@Nonnull Class<T> type) {
```
jglick: `extends PersistentAction`

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -244,66 +247,120 @@ public String getUrl() throws IOException {
...
+    protected synchronized void setActions(List<Action> actions) {
+            this.actions = new CopyOnWriteArrayList<Action>(actions);
+//            this._wrapper = new ListWrapper(this);
     }
 
-/*
-    We can't use Actionable#actions to store actions because they aren't transient,
-    and we need to store actions elsewhere because this is the only mutable pat of FlowNode.
+    /**
+     * Return the first persistent action on the FlowNode, without consulting TransientActionFactories
```
jglick: Summary sentences must end with a period. Also `{@link TransientActionFactory}s`.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -244,66 +247,120 @@ public String getUrl() throws IOException {
...
+     * Used here because it is much faster to invoke a less-conditional, monomorphic method (50% faster)
```
jglick: You are positive that the call to `PersistentAction.class.isAssignableFrom(type)` is significant enough to be measurable?

Both methods are monomorphic BTW—they take a (`final`) `Class`.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -244,66 +247,120 @@ public String getUrl() throws IOException {
...
+     * @param type Class of action
+     * @param <T>  Action type
+     * @return First persistent action or null if not found
```
jglick: `@CheckForNull`

pom.xml
```diff
@@ -33,7 +33,7 @@
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-api</artifactId>
-    <version>2.5-SNAPSHOT</version>
+    <version>2.5-optimized3-SNAPSHOT</version>
```
stephenc: Unnecessary change

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
jglick: Can this be `@Restricted(NoExternalUse.class)` or are you planning to call this from some downstream PR?

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -244,66 +247,120 @@ public String getUrl() throws IOException {
...
+     */
+    public <T extends Action> T getPersistentAction(@Nonnull Class<T> type) {
+        if (actions == null) {
+            loadActions();
```
jglick: Since `actions` is not `volatile` and you are not acquiring a monitor, theoretically under heavy threading conditions it is possible for you to observe an incompletely initialized list here, which could result in anomalous behavior and errors.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -244,66 +247,120 @@ public String getUrl() throws IOException {
...
+            }
+        }
+        return null;
+    }
+
+    @Override
+    public <T extends Action> T getAction(Class<T> type) {
+        // Delegates internally to methods that are not overloads, which are more subject to inlining and optimization
+        // Normally a micro-optimization, but these methods are invoked *heavily* and improves performance 5%
+        // In flow analysis
```
jglick: Would it suffice to just make `getAction` be `final` and inline its delegates?

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
stephenc: I feel like you don't need this above `private` scope at all. You can just override `getAction(Class<T> type)` e.g.

```
public <T extends Action> T getAction(@Nonnull Class<T> type) {
  if (PersistentAction.class.isAssignableFrom(type)) {
    // short-cut
    if (actions == null) {
      loadActions();
    }
     for (Action a : actions) {
       if (type.isInstance(a)) {
         return type.cast(a);
       }
     }
     // no point checking super as we know its a persistent action
     return null;
  }
  // we have to check the transients now
  return super.getAction(type);
}
```

and that should be short enough to fit within inlining without requiring callers to know which getAction method to call

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -244,66 +247,120 @@ public String getUrl() throws IOException {
...
-                return actions.get(index);
-            }
 
-            @Override
-            public void add(int index, Action element) {
-                actions.add(index, element);
-                persist();
-            }
+                @Override
+                public Action get(int index) {
```
jglick: Why the reformatting? Makes diff gratuitously large.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -315,6 +372,15 @@ public void save() throws IOException {
         exec.saveActions(this, actions);
     }
 
+    // Persist, handling possible IOException
+    private void persistSafe() {
```
jglick: Why was this moved out of the inner class?

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
```diff
@@ -57,7 +51,7 @@ private FlowScanningUtils() {}
         return new Predicate<FlowNode>() {
             @Override
             public boolean apply(FlowNode input) {
-                return (input != null && input.getAction(actionClass) != null);
+                return (input != null && input.getPersistentAction(actionClass) != null);
```
jglick: An incompatible change in case the `Action` was not a `PersistentAction`. If intentional, must at least change the method signature to require `PersistentAction` so that it becomes source-incompatible for any callers requesting another `Action` type.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
svanoort: @stephenc  Did you look at where the code?

stephenc: yes I saw already that your getAction is basically this... what I am saying here is that *this* method can be `private`... no need for it to be `public` and then you are not adding a new method... I also question whether the increased cognitive load is worth not just calling `super` for the non-persistent form

pom.xml
stephenc: Then put a `<!-- FIXME remove custom version before merge-->` still a :bug: without the comment

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
svanoort: @jglick It could be for now.  I'll do it if it's the only way to get your bee, but still like the idea of being able to invoke it in other performance intensive paths.

svanoort: Valid, will add this.

svanoort: Capping it for now in pending commit.

svanoort: Ah, introduces a typing issue.  NVM.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -249,17 +250,19 @@ public String getUrl() throws IOException {
...
+     * Used here because it is much faster than base {@link #getAction(Class)} method.
      * @param type Class of action
      * @param <T>  Action type
-     * @return First persistent action or null if not found
+     * @return First nontransient action or null if not found.
      */
-    public <T extends Action> T getPersistentAction(@Nonnull Class<T> type) {
+    @CheckForNull
+    @Restricted(NoExternalUse.class)  // Limit use to workflow-api packages until we have a case where we need the performance badly.
+    public final <T extends Action> T getPersistentAction(@Nonnull Class<T> type) {
```
jglick: Advisable to make this `<T extends PersistentAction>`; mandatory if removing the restriction.

src/main/java/org/jenkinsci/plugins/workflow/actions/PersistentAction.java
```diff
@@ -0,0 +1,10 @@
+package org.jenkinsci.plugins.workflow.actions;
+
+import hudson.model.Action;
+
+/**
+ * This is a marker interface for an action that can't be contributed by a {@link jenkins.model.TransientActionFactory}.
+ * Actions implementing this can use more efficient {@link hudson.model.Actionable#getAction(Class)} variant
```
jglick: Feel free to `@link` to `getPersistentAction`—its being `@Restricted` does not prevent it from appearing in Javadoc.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -244,66 +248,123 @@ public String getUrl() throws IOException {
      *
      * This method provides such an opportunity for subtypes.
      */
-    protected void setActions(List<Action> actions) {
-        this.actions = new CopyOnWriteArrayList<Action>(actions);
+    protected synchronized void setActions(List<Action> actions) {
```
jglick: Unnecessary to make this `synchronized` as it is only ever called from subclass constructors. (Really it should better have been made `final`: calling an overridable method from a constructor is inherently dangerous in Java, unlike in say C++.)

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -244,66 +248,123 @@ public String getUrl() throws IOException {
...
-                        actions = new CopyOnWriteArrayList<Action>(exec.loadActions(this));
-                    } catch (IOException e) {
-                        LOGGER.log(WARNING, "Failed to load actions for FlowNode id=" + id, e);
-                        actions = new CopyOnWriteArrayList<Action>();
-                    }
-                }
+    @SuppressFBWarnings(value = "UG_SYNC_SET_UNSYNC_GET", justification = "CopyOnWrite ArrayList, and field load & modification is synchronized")
+    public List<Action> getActions() {
+        if (actions==null) {
+            loadActions();
```
jglick: :bug: Double-checked locking. Replace with simple

```java
loadActions();
```

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -244,66 +248,123 @@ public String getUrl() throws IOException {
...
+     * Used here because it is much faster than base {@link #getAction(Class)} method.
+     * @param type Class of action
+     * @param <T>  Action type
+     * @return First nontransient action or null if not found.
+     */
+    @CheckForNull
+    @Restricted(NoExternalUse.class)  // Limit use to workflow-api packages until we have a case where we need the performance badly.
+    public final <T extends Action> T getPersistentAction(@Nonnull Class<T> type) {
+        if (actions == null) {
+            loadActions();
```
jglick: Another case of double-checked locking. Replace with

```java
for (Action a : getActions()) {
```

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -244,66 +248,123 @@ public String getUrl() throws IOException {
...
+     * Return the first nontransient {@link Action} on the FlowNode, without consulting {@link jenkins.model.TransientActionFactory}s
+     * <p> This is not restricted to just Actions implementing {@link PersistentAction} but usually they should.
+     * Used here because it is much faster than base {@link #getAction(Class)} method.
+     * @param type Class of action
+     * @param <T>  Action type
+     * @return First nontransient action or null if not found.
+     */
+    @CheckForNull
+    @Restricted(NoExternalUse.class)  // Limit use to workflow-api packages until we have a case where we need the performance badly.
+    public final <T extends Action> T getPersistentAction(@Nonnull Class<T> type) {
```
jglick: Use `T extends PersistentAction`.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -244,66 +248,123 @@ public String getUrl() throws IOException {
...
+    protected synchronized void setActions(List<Action> actions) {
+            this.actions = new CopyOnWriteArrayList<Action>(actions);
     }
 
-/*
-    We can't use Actionable#actions to store actions because they aren't transient,
-    and we need to store actions elsewhere because this is the only mutable pat of FlowNode.
+    /**
+     * Return the first nontransient {@link Action} on the FlowNode, without consulting {@link jenkins.model.TransientActionFactory}s
+     * <p> This is not restricted to just Actions implementing {@link PersistentAction} but usually they should.
```
jglick: Under what conditions would you call this on a non-`PersistentAction`? Unless there is a documented case, better to just force it to be used only for `PersistentAction`s.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -244,66 +248,123 @@ public String getUrl() throws IOException {
...
-            @Override
-            public Action set(int index, Action element) {
-                Action old = actions.set(index, element);
-                persist();
-                return old;
-            }
+                @Override
+                public boolean add(Action value) {
+                    actions.add(value);
+                    persistSafe();
```
jglick: Get in the habit of running

    git diff master

to look for gratuitous hunks that could be reverted after design changes.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
svanoort: @jglick  Results in "type is not within its bound" compile error for the main getAction method invoking this.   Also not necessary, see javadocs: 

> This is not restricted to just Actions implementing {@link PersistentAction} but usually they should.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
svanoort: When we want to distinguish between versions of an Action directly attached vs. contributed by a TransientActionFactory.  Restricting it would add complexity to work around the generics quirk AFAIK (maybe there's an easy way to do that).

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
svanoort: Will be sorted the same way as the other use.

jglick:
> Results in "type is not within its bound" compile error for the main getAction method invoking this.

`Class.asSubclass`

jglick:
> When we want to distinguish between versions of an Action directly attached vs. contributed by a TransientActionFactory.

My understanding was that for a given `Action` type, either we declare it persistent or we do not—it is not up to the caller to choose.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -244,66 +248,113 @@ public String getUrl() throws IOException {
...
-    public synchronized List<Action> getActions() {
-                if (actions==null) {
-                    try {
-                        actions = new CopyOnWriteArrayList<Action>(exec.loadActions(this));
-                    } catch (IOException e) {
-                        LOGGER.log(WARNING, "Failed to load actions for FlowNode id=" + id, e);
-                        actions = new CopyOnWriteArrayList<Action>();
-                    }
-                }
+    @SuppressFBWarnings(value = "UG_SYNC_SET_UNSYNC_GET", justification = "CopyOnWrite ArrayList, and field load & modification is synchronized")
```
jglick: Is this still necessary?

svanoort: Yup, I checked.

PR #22, opened 2016-10-20
LinearScanner cannot be called with multiple heads

jglick:
Would make mistakes like in https://github.com/jenkinsci/workflow-support-plugin/pull/22 less likely.

@reviewbybees esp. @svanoort

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/LinearBlockHoppingScanner.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/LinearBlockHoppingScanner.java
@@ -32,6 +32,7 @@ import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import javax.annotation.concurrent.NotThreadSafe;
 import java.util.Collection;
+import java.util.Iterator;
 import java.util.List;
 
 /**
@@ -66,9 +67,13 @@ public class LinearBlockHoppingScanner extends LinearScanner {
 
     @Override
     protected void setHeads(@Nonnull Collection<FlowNode> heads) {
-        if (heads.size() > 0) {
-            this.myCurrent = jumpBlockScan(heads.iterator().next(), myBlackList);
+        Iterator<FlowNode> it = heads.iterator();
+        if (it.hasNext()) {
+            this.myCurrent = jumpBlockScan(it.next(), myBlackList);
             this.myNext = this.myCurrent;
+            if (it.hasNext()) {
+                throw new IllegalArgumentException("Multiple heads not supported for linear scanners");
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


svanoort:
Why so many overrides for LinearScanner?  It should be sufficient to just override the setup method since everything delegates to that.

jglick:
The only real override is of `setHeads`. All the rest are just to adjust Javadoc and (where possible) add a deprecation warning.

svanoort:
@jglick That just adds noise and clutter -- it's enough to mention this in the top-level javadoc, since that's what defines the behavior for each FlowScanner.

svanoort:
To be clear: the change itself seems reasonable and prudent.

jglick:
> it's enough to mention this in the top-level javadoc

Assuming someone *reads* the top-level Javadoc carefully. By deprecating the particular methods which are advertised to take a collection of nodes—but which really will not work if more than one is passed—the limitation becomes much more apparent. For example, the code I wrote in https://github.com/jenkinsci/workflow-support-plugin/pull/19 would have shown a deprecation warning in my IDE.

svanoort:
> Assuming someone reads the top-level Javadoc carefully. 

Well, if you expect people won't read the javadoc, why both making any change to it? ;)

I am -1 on deprecating these methods: it's perfectly legit to throw an exception here but the reason they will accept a list is so that you can pass in getCurrentHeads directly.

jglick:
> if you expect people won't read the javadoc, why both making any change to it?

Deprecation warnings appear more proactively. You do not need to go looking at the Javadoc to see that there is a problem.

> the reason they will accept a list is so that you can pass in getCurrentHeads directly

And that is exactly what you should not do, because then your code will work fine until the day something uses `parallel` and then it will explode.

jglick:
IOW with the deprecations in place, the caller is forced to pay attention to the fact that the scanner cannot process multiple heads, and so you must think about what you want to do instead. In my case, it was to scan from each head in sequence.

jglick:
CI build passed, status notifications seems to be down.

jglick:
> we're making LinearScanner violate the contracts of AbstractFlowScanner

Yes, that is exactly the idea. The behavior when multiple heads are passed already violates any sensible expectations; a clear exception makes it no worse.

> Logging a warning with multiple heads is appropriate and would assist debugging.

Somewhat, but only after you encounter the mistake. In the case I encountered, that would have made developing the fix slightly faster, but would not have stopped me from producing the buggy release to begin with.

If you are talking about retaining the deprecations but switching the two `IllegalArgumentException`s to warnings, I still think that is less desirable than the current PR, but certainly better than nothing.

jglick:
> useless TODOs for deprecation warnings

Was not immediately clear to me what the replacement for those methods would be. New overloads in `AbstractFlowScanner`? Or non-overriding methods in `LinearScanner`? Or no replacement (caller expected to use different idioms)?

svanoort:
@jglick Overloads would be my suggestion in AbstractFlowScanner, or in the Linear ones (either is fine by me)

jglick:
Makes sense. Deferring that until a rainy day (or someone needs such calls).

PR #24, opened 2016-11-04
Revert removal of TagsAction: Provide Binding for Attaching Metadata To Stages/Steps

svanoort:
This reverts commit 428319414b7fa7288bec1652202188a1ea52daa5.

The original code was approved in https://github.com/jenkinsci/workflow-api-plugin/pull/12 but reverted when the downstream consumer was killed (to be restored at a later date when the next consumer came around). 

Now a consumer has arisen anew!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-api</artifactId>
-    <version>2.6-SNAPSHOT</version>
+    <version>2.6-add-tagsaction-SNAPSHOT</version> <!-- Revert back to 2.6-SNAPSHOT after merge -->
     <packaging>hpi</packaging>
     <name>Pipeline: API</name>
     <url>https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+API+Plugin</url>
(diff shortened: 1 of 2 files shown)


svanoort:
@abayer all for you mang

jglick:
Pending a downstream PR.

abayer:
I'll have the new Declarative PR for synthetic stages up today, and then we'll be waiting on @vivek and others at Blue Ocean to consume *that* PR.

abayer:
:bee: and downstream PR created. =)

jglick:
https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/47 still appears to be under active discussion, and there is no apparent `blueocean` PR yet, so I think this is still in a pending state.

jglick:
What issue are you purporting to solve with this plus https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/47 plus https://github.com/jenkinsci/blueocean-plugin/pull/608? It does seem to have much of anything to do with the reported purpose of JENKINS-26522. Can you update the PR title and description to describe what the intent is?

svanoort:
@jglick This was already approved, the only blocker was the need for the consumer of this API to exist, which we now have.  I have to admit I'm bit confused -- why the change of heart here?

abayer:
re-:bee:, fwiw.

vivek:
🐝 there is blueocean PR that has use case to validate these APIs. LGTM.

i386:
I assume this is https://issues.jenkins-ci.org/browse/JENKINS-26522 ? Should be in review?

svanoort:
*Part* of Jenkins-26522.  There's a couple more pieces needed to get the complete result.

PR #25, opened 2016-11-23
Fix [JENKINS-38089] which is caused by one specific edge case with incomplete parallels

svanoort:
Resolves [JENKINS-38089](https://issues.jenkins-ci.org/browse/JENKINS-38089). 

Root cause was prematurely terminating the attempt to find the start of an incomplete set of parallels... because we hit the start of the flow with one search branch, but we hadn't hit the merge point with the other.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
@@ -327,7 +327,9 @@ public class ForkScanner extends AbstractFlowScanner {
         }
 
         // Walk through, merging flownodes one-by-one until everything has merged to one ancestor
-        while (iterators.size() > 1) {
+        boolean mergedAll = false;
+		// Ends when we merged all branches together, or hit the start of the flow without it
+        while (!mergedAll && iterators.size() > 0) {
             ListIterator<Filterator<FlowNode>> itIterator = iterators.listIterator();
             ListIterator<FlowPiece> pieceIterator = livePieces.listIterator();
 
@@ -374,6 +376,9 @@ public class ForkScanner extends AbstractFlowScanner {
                     // Merging removes the piece & its iterator from heads
                     itIterator.remove();
                     pieceIterator.remove();
+                    if (iterators.size() == 1) { // Merged in the final branch
+                        mergedAll = true;
+                    }
                 }
             }
         }
@@ -386,6 +391,7 @@ public class ForkScanner extends AbstractFlowScanner {
     protected void setHeads(@Nonnull Collection<FlowNode> heads) {
         if (heads.size() > 1) {
             parallelBlockStartStack = leastCommonAncestor(new LinkedHashSet<FlowNode>(heads));
+            assert parallelBlockStartStack.size() > 0;
             currentParallelStart = parallelBlockStartStack.pop();
             currentParallelStartNode = currentParallelStart.forkStart;
             myCurrent = currentParallelStart.unvisited.pop();
--- a/src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
@@ -33,7 +33,6 @@ import org.jenkinsci.plugins.workflow.cps.nodes.StepStartNode;
 import org.jenkinsci.plugins.workflow.cps.steps.ParallelStep;
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
 import org.jenkinsci.plugins.workflow.graph.BlockStartNode;
-import org.jenkinsci.plugins.workflow.graph.FlowGraphWalker;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
@@ -49,7 +48,6 @@ import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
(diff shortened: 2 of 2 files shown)


svanoort:
@reviewbybees

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -327,7 +327,9 @@ public Fork(BlockStartNode forkNode) {
         }
 
         // Walk through, merging flownodes one-by-one until everything has merged to one ancestor
-        while (iterators.size() > 1) {
+        boolean mergedAll = false;
+		// Ends when we merged all branches together, or hit the start of the flow without it
```
jglick: tabs

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -386,6 +391,7 @@ public Fork(BlockStartNode forkNode) {
     protected void setHeads(@Nonnull Collection<FlowNode> heads) {
         if (heads.size() > 1) {
             parallelBlockStartStack = leastCommonAncestor(new LinkedHashSet<FlowNode>(heads));
+            assert parallelBlockStartStack.size() > 0;
```
jglick: Pointless AFAICT because the next line would just throw an exception were it empty anyway. (That is the reported bug, right?)

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -406,6 +404,74 @@ public void testLeastCommonAncestor() throws Exception {
         Assert.assertEquals(2, starts.size());
     }
 
+    @Test
```
jglick:
```java
@Issue("JENKINS-38089")
```

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -406,6 +404,74 @@ public void testLeastCommonAncestor() throws Exception {
...
+    public void testVariousParallelCombos() throws Exception {
+        WorkflowJob job = r.jenkins.createProject(WorkflowJob.class, "ParallelTimingBug");
```
jglick: BTW the project name does not matter—can use just `"p"` or whatever for every test.

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -406,6 +404,74 @@ public void testLeastCommonAncestor() throws Exception {
...
+        job.setDefinition(new CpsFlowDefinition(
+                "stage 'test' \n" +
+                        "    parallel([\n" +  // ID 5 is start
+                        "        'unit': {\n" +
```
jglick: BTW can omit some brackets here:

```groovy
parallel unit: {…}, otherunit: {…}
```

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -406,6 +404,74 @@ public void testLeastCommonAncestor() throws Exception {
...
+                        "          retry(1) {\n" +
+                        "            sleep 1;\n" +
+                        "            sleep 10; echo 'hello'; \n" +
```
jglick: Seems like you are making the test gratuitously slow.

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -406,6 +404,74 @@ public void testLeastCommonAncestor() throws Exception {
...
+                        "        'otherunit': {\n" +
+                        "            retry(1) {\n" +
+                        "              sleep 1;\n" +
+                        "              sleep 5; \n" +
+                        "              echo 'goodbye'   \n" +
+                        "            }\n" +
+                        "        }\n" +  // end of branch:
+                        "    ])\n"
+        ));
+        /*Node dump follows, format:
```
jglick: Really need a new test utility for these kinds of assertions…

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -406,6 +404,74 @@ public void testLeastCommonAncestor() throws Exception {
...
+
+        // Test different start points in branch A & B, 20 and 19 were one error case.
+        for (int i=0; i < 4; i++) {
+            for (int j=0; j<5; j++) {
+                int branchANodeId = i+20;
+                int branchBNodeId = j+15;
+                System.out.println("Starting test with nodes "+branchANodeId+","+branchBNodeId);
+                ArrayList<FlowNode> starts = new ArrayList<FlowNode>();
+                FlowTestUtils.addNodesById(starts, exec, branchANodeId, branchBNodeId);
+                List<FlowNode> all = scan.filteredNodes(starts, Predicates.<FlowNode>alwaysTrue());
```
jglick: Is there not some method to get all nodes, unfiltered?

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -406,6 +404,74 @@ public void testLeastCommonAncestor() throws Exception {
...
+            A [21]{20}StepEndNode Retry the body up to N times : Body : End  [st=9]
+            A [22]{21}StepEndNode Retry the body up to N times : End  [st=8]
+            A [23]{22}StepEndNode Execute in parallel : Body : End  [st=6]
+        [24]{23,19}StepEndNode Execute in parallel : End  [st=4]
+        [25]{24}FlowEndNode End of Pipeline  [st=2]*/
+        WorkflowRun b = r.assertBuildStatusSuccess(job.scheduleBuild2(0));
+        FlowExecution exec = b.getExecution();
+        ForkScanner scan = new ForkScanner();
+
+        // Test different start points in branch A & B, 20 and 19 were one error case.
```
jglick: IIUC one of these iterations threw `NoSuchElementException` prior to the fix?

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
svanoort: @jglick That's the output of a utility in pipeline-graph-analysis to dump node structure, it's not generated by hand.

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
svanoort: Nope -- it's an iterator so there are a bunch of trivial one-liners that will accomplish that goal.  This is one but there are a ton you can do using Guava.

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
jglick: Understood; just harking back to a comment on your original analysis PR, that we should have a test utility which allows you to express a human-writable and -readable assertion about graph structure without needing to hard-code IDs or do a sample run.

jglick:
:bee:

abayer:
:bee:

svanoort:
Nits to be addressed in #20 since there's no point going into re-review if we've got another PR coming soon.

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
svanoort: @jglick Yup, I hadn't forgotten it -- I think such a case might make a reasonable graph analysis feature too (not just as a test framework) -- unfortunately as we know there are lots of edge cases in graph structure present that need to be handled.  

I think it's not really viable until we have canonical tree-structured navigation of pipeline flow graphs -- which is the case that would benefit most from this for testing.

svanoort: @jglick Yup, I hadn't forgotten it -- I think such a case might make a reasonable graph analysis feature too (not just as a test framework) -- unfortunately as we know there are lots of edge cases in graph structure present that need to be handled.  

I think it's not really viable until we have canonical tree-structured navigation of pipeline flow graphs -- which is the case that would benefit most from this for testing.

svanoort: @jglick Yup, I hadn't forgotten it -- I think such a case might make a reasonable graph analysis feature too (not just as a test framework) -- unfortunately as we know there are lots of edge cases in graph structure present that need to be handled.  

I think it's not really viable until we have canonical tree-structured navigation of pipeline flow graphs -- which is the case that would benefit most from this for testing.

svanoort: @jglick Yup, I hadn't forgotten it -- I think such a case might make a reasonable graph analysis feature too (not just as a test framework) -- unfortunately as we know there are lots of edge cases in graph structure present that need to be handled.  

I think it's not really viable until we have canonical tree-structured navigation of pipeline flow graphs -- which is the case that would benefit most from this for testing.

svanoort: @jglick Yup, I hadn't forgotten it -- I think such a case might make a reasonable graph analysis feature too (not just as a test framework) -- unfortunately as we know there are lots of edge cases in graph structure present that need to be handled.  

I think it's not really viable until we have canonical tree-structured navigation of pipeline flow graphs -- which is the case that would benefit most from this for testing.

PR #26, opened 2017-01-25
[JENKINS-37324] Step action API

svanoort:
Basic API for storing the step parameters (can be used to instantiate it from the StepDescriptor if needed) to have tracking of the parameters used. 

upstream of https://github.com/jenkinsci/workflow-cps-plugin/pull/98


src/main/java/org/jenkinsci/plugins/workflow/actions/StepInfoAction.java
```diff
@@ -0,0 +1,119 @@
...
+     *  if part of the parameters are not retained.
+     * @return The parameters for the Step.
+     */
+    @Nonnull
+    public Map<String,Object> getParameters() {
+        return Collections.unmodifiableMap(getParametersInternal());
+    }
+
+    @Nonnull
+    public static Map<String, Object> getNodeParameters(@Nonnull  FlowNode m) {
```
svanoort: @vivek This is probably the point of interest for you, although also getParameterValue could be of assistance

vivek: @svanoort yes, thanks. Yeah getParameterValue() is going to be handy too, without which I have to have such method to get reason or value:)

src/main/java/org/jenkinsci/plugins/workflow/actions/StepInfoAction.java
```diff
@@ -0,0 +1,119 @@
...
+ */
+public abstract class StepInfoAction implements PersistentAction {
+
+    /** Used as a placeholder for parameters not stored for various reasons */
+    public enum NotStoredReason {
+        /** Denotes an unsafe value that cannot be stored/displayed due to sensitive info */
+        MASKED_VALUE,
+
+        /** Denotes an object that is too big to retain, such as strings exceeding {@link #MAX_STRING_LENGTH} */
+        OVERSIZE_VALUE
```
vivek: How about giving truncated value in this case? Max limit is going to be env dependent (from perf perspective), so its arbitrary limit anyway.

svanoort: @vivek I've gone back and forth on this -- there are merits both ways.  If we store the truncated value then we have to separately record which fields are truncated though, which makes the clientside handling more complex, because values may currently be:

* Stored fully (yay)
* Null (explicitly set or unspecified, legitimately)
* Masked out
* Oversized (omitted)

But with truncation, you have to check for each value that is stored if it is the full content or truncated as well.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,177 @@
...
+     *  if part of the arguments are not retained.
+     * @return The arguments for the Step.
+     */
+    @Nonnull
+    public Map<String,Object> getArguments() {
+        return Collections.unmodifiableMap(getArgumentsInternal());
+    }
+
+    @Nonnull
+    public static Map<String, Object> getArguments(@Nonnull  FlowNode n) {
```
abayer: nit - extra space after `@Nonnull`. =)

abayer:
Looks good at first read-through.

svanoort:
@abayer We can't really test these core APIs without concrete implementations - that's why the tests have to more or less live in workflow-cps.

abayer:
Fair point. =)

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,177 @@
...
+     */
+    @Nonnull
+    public static Map<String, Object> getFilteredArguments(@Nonnull FlowNode n) {
+        ArgumentsAction act = n.getPersistentAction(ArgumentsAction.class);
+        return act != null ? act.getFilteredArguments() : Collections.EMPTY_MAP;
+    }
+
+    /** Return a tidy string description for the step arguments, or null if none is present or we can't make one */
+    @CheckForNull
+    public static String getArgumentDescriptionString(@Nonnull FlowNode n) {
```
svanoort: @vivek @abayer Think this would make more sense as "getDescription"?

svanoort:
@abayer @vivek Do you think it would be worthwhile to add an implementable API for StepDescriptors that filter or transform their arguments before saving the argumentAction?  I.E. to allow some values to be security-masked using information that is only visible in the description, or converted into a more-usable format.

Also do we have a solid use case for distinguishing from values supplied via defaults vs. explicitly specified by user? I can see value there, but it would increase complexity significantly. 

Maybe that's overengineering?

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
vivek: Sounds good to me.

vivek:
> Do you think it would be worthwhile to add an implementable API for StepDescriptors that filter or transform their arguments before saving the argumentAction? I.E. to allow some values to be security-masked using information that is only visible in the description, or converted into a more-usable format.

This sounds like good idea.

> Also do we have a solid use case for distinguishing from values supplied via defaults vs. explicitly specified by user? I can see value there, but it would increase complexity significantly.

Not sure if we have use case for that at present. @i386?

svanoort:
Beta cut as workflow-api version 2.14-stepstorage4-beta

svanoort:
@reviewbybees

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,169 @@
...
+    public enum NotStoredReason {
+        /** Denotes an unsafe value that cannot be stored/displayed due to sensitive info */
+        MASKED_VALUE,
+
+        /** Denotes an object that is too big to retain, such as strings exceeding {@link #MAX_STRING_LENGTH} */
+        OVERSIZE_VALUE
+    }
+
+    /** Largest string we'll persist in the Step, for performance reasons */
+    public static final int MAX_STRING_LENGTH = 1024;
```
stephenc: So I can store 2 x 1024 byte string arguments but not a single 1025 byte string argument. If storage size is an issue in performance it would seem more sensible to stop storing all strings above say 256 bytes when the total of all strings above 256 bytes is > 1024 bytes

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,169 @@
...
+    @CheckForNull
+    public Object getArgumentValueOrReason(@Nonnull String argumentName) {
+        return getArgumentsInternal().get(argumentName);
+    }
+
+    /**
+     * Test if {@link Step} arguments are persisted in an unaltered form.
+     * @return True if full arguments are retained, false if some have been removed for security, size, or other reasons.
+     */
+    public abstract boolean isFullArguments();
```
stephenc: Any reason you cannot provide a default implementation? (i.e. a transient `Boolean` field if `null` just walk `getArgumentsInternal().values()` looking for a `NotStoredReason` followed by an idempotent write to the boolean field... or just the walk of `getArgumentsInternal().values()` looking for a `NotStoredReason`)

If there is a reason why you cannot provide a default implementation using the walk then you probably want to advertise that reason in the javadoc so that implementations do not fall foul of the issue you are concerned with

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,169 @@
...
+    public static final int MAX_STRING_LENGTH = 1024;
+
+    @Override
+    public String getIconFileName() {
+        return null;
+    }
+
+    @Override
+    public String getDisplayName() {
+        return "Step Arguments";
```
jglick: Use `null`, or just extend `InvisibleAction`—unless you really intend to show this in, say, **Pipeline Steps**, in which case you would need an `iconFileName` and some Jelly views.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    @Override
+    public String getUrlName() {
+        return "stepArguments";
```
jglick: ditto

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,169 @@
...
+     *  if part of the arguments are not retained.
+     * @return The arguments for the Step.
+     */
+    @Nonnull
+    public Map<String,Object> getArguments() {
+        return Collections.unmodifiableMap(getArgumentsInternal());
+    }
+
+    @Nonnull
+    public static Map<String,Object> getArguments(@Nonnull FlowNode n) {
```
jglick: Deserves Javadoc, especially explaining that an empty map would be produced if there is missing information.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,169 @@
...
+        return "Step Arguments";
+    }
+
+    @Override
+    public String getUrlName() {
+        return "stepArguments";
+    }
+
+    /**
+     * Get the map of arguments for the {@link Step}, with a {@link NotStoredReason} instead of the value
```
jglick: Should clarify what "value” means in this context. Could it be part of a nested `Map`, for example?

And need to clarify what the argument types are even excluding `NotStoredReason`. The same as `instantiate`, so potentially including `UninstantiatedDescribable`? Or what?

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    @Nonnull
+    public static Map<String,Object> getArguments(@Nonnull FlowNode n) {
+        ArgumentsAction aa = n.getPersistentAction(ArgumentsAction.class);
+        return aa != null ? aa.getArguments() : Collections.EMPTY_MAP;
+    }
+
+    /**
+     * Get just the fully stored, non-null arguments
```
jglick: Summary sentences must end in `.`.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,169 @@
...
+        for (Map.Entry<String, Object> entry : internalArgs.entrySet()) {
+            if (entry.getValue() != null && !(entry.getValue() instanceof NotStoredReason)) {
+                filteredArguments.put(entry.getKey(), entry.getValue());
+            }
+        }
+        return filteredArguments;
+    }
+
+    /**
+     * Get just the fully stored, non-null arguments
```
jglick: ditto

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /**
+     * Get the argument value or its {@link NotStoredReason} if it has been intentionally omitted.
+     * @param argumentName Name of step argument to find value for
+     * @return Argument value, null if nonexistent/null, or NotStoredReason if it existed by was masked out.
+     */
+    @CheckForNull
+    public Object getArgumentValueOrReason(@Nonnull String argumentName) {
+        return getArgumentsInternal().get(argumentName);
```
jglick: If this is a nested `Map`, we are failing to ensure that it is immutable.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,169 @@
...
+    @CheckForNull
+    public Object getArgumentValue(@Nonnull String argumentName) {
+        Object val = getArgumentValueOrReason(argumentName);
+        return (val == null || val instanceof NotStoredReason) ? null : val;
+    }
+
+    /**
+     * Get the argument value or its {@link NotStoredReason} if it has been intentionally omitted.
+     * @param argumentName Name of step argument to find value for
+     * @return Argument value, null if nonexistent/null, or NotStoredReason if it existed by was masked out.
```
jglick: use `@link` whenever referring to types

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
jglick: Agreed, it seems like a default implementation would be straightforward.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/AbstractFlowScanner.java
```diff
@@ -339,6 +339,14 @@ public FlowNode findFirstMatch(@CheckForNull FlowExecution exec, @Nonnull Predic
         return this.filteredNodes(Collections.singleton(head), null, matchPredicate);
     }
 
+    @Nonnull
+    public List<FlowNode> filteredNodes(@CheckForNull FlowExecution exec, @Nonnull Predicate<FlowNode> matchPredicate) {
```
jglick: Javadoc?

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowTestUtils.java
```diff
@@ -1,4 +1,5 @@
-package org.jenkinsci.plugins.workflow.graphanalysis;/*
+package org.jenkinsci.plugins.workflow.graphanalysis;
+/*
```
jglick: Irrelevant to this PR AFAICT.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,169 @@
...
+     * @param n FlowNode to get arguments for
+     * @return Map of all completely stored arguments
+     */
+    @Nonnull
+    public static Map<String, Object> getFilteredArguments(@Nonnull FlowNode n) {
+        ArgumentsAction act = n.getPersistentAction(ArgumentsAction.class);
+        return act != null ? act.getFilteredArguments() : Collections.EMPTY_MAP;
+    }
+
+    /** Return a tidy string description for the step arguments, or null if none is present or we can't make one
```
jglick: ditto

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,169 @@
...
+    protected abstract Map<String, Object> getArgumentsInternal();
+
+    /**
+     * Get the value of a argument, or null if not present/not stored.
+     * Use {@link #getArgumentValueOrReason(String)} if you want to return the {@link NotStoredReason} rather than null.
+     * @param argumentName Argument name of step to look up.
+     * @return Argument value or null if not present/not stored.
+     */
+    @CheckForNull
+    public Object getArgumentValue(@Nonnull String argumentName) {
```
jglick: Should this be `final`?

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,169 @@
...
+        return (val == null || val instanceof NotStoredReason) ? null : val;
+    }
+
+    /**
+     * Get the argument value or its {@link NotStoredReason} if it has been intentionally omitted.
+     * @param argumentName Name of step argument to find value for
+     * @return Argument value, null if nonexistent/null, or NotStoredReason if it existed by was masked out.
+     */
+    @CheckForNull
+    public Object getArgumentValueOrReason(@Nonnull String argumentName) {
```
jglick: Ditto—should be `final` unless you can provide a specific reason why a subclass should be overriding it.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,169 @@
...
+    /**
+     * Get just the fully stored, non-null arguments
+     * This means the arguments with all {@link NotStoredReason} or null values removed
+     * @param n FlowNode to get arguments for
+     * @return Map of all completely stored arguments
+     */
+    @Nonnull
+    public static Map<String, Object> getFilteredArguments(@Nonnull FlowNode n) {
+        ArgumentsAction act = n.getPersistentAction(ArgumentsAction.class);
+        return act != null ? act.getFilteredArguments() : Collections.EMPTY_MAP;
```
jglick: BTW prefer the generics-friendly `emptyMap()`.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
jglick: Probably limits like this should be left up to the implementation (`ArgumentsActionImpl` downstream). There is no need to specify it in the API.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
svanoort: @jglick Grumble grumble, yeah you have a point here.  I'd like to provide a view (because I think that's rather useful for debugging) but that'll probably have to be a follow-up PR for display unless this needs some extra work in that area.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
svanoort: @stephenc Agree that the length naming is not as helpful as it might be -- but there's no sense in throwing away perfectly useful arguments just because one is oversized (such as file content). 

Would you and @jglick be satisfied if I renamed it to "MAX_ARRAY_LENGTH" and moved the workflow-cps filtering logic here in some fashion? (Assuming that will work, I'll have to double check we don't need anything provided by workflow-cps).

I've also considered offering an API for StepDescriptors that want to provide their own filtering/transformation of stored arguments beyond the security filtering (as mentioned to Vivek above).  But we just don't have a need for it yet, so YAGNI for now...

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
svanoort: Don't see any reason why I couldn't do that, I suppose.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
svanoort: [cough](https://www.merriam-webster.com/dictionary/pedant)  ;-)

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
jglick: Better to keep this all out of the API. Let the implementation decide whether and how to trim stuff. The API need only define the marker that such trimming occurred.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
jglick: I know, minor, but also easily corrected and will improve readability [here](http://javadoc.jenkins.io/plugin/workflow-api/) and from IDEs.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
michaelneale: I think it would be a disaster to leave it up to implementers. Something stupid will be done, and it will be very bad. Very bad. If this is used for what I think it is used for, think for 2 seconds and realise that a hard (small) limit is a very very good idea.

jglick: There will be only one implementation, which @svanoort has already written. I am just saying that from the perspective of the separation of API from implementation there is no need for this constant to be present here and so it should not be.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,169 @@
...
+ * Stores some or all of the arguments used to create and configure the {@link Step} executed by a {@link FlowNode}.
+ * This allows you to inspect information supplied in the pipeline script and otherwise discarded at runtime.
+ * Supplied argument values can be hidden and replaced with a {@link NotStoredReason} for security or performance.
+ */
+public abstract class ArgumentsAction implements PersistentAction {
+
+    /** Used as a placeholder for arguments not stored for various reasons */
+    public enum NotStoredReason {
+        /** Denotes an unsafe value that cannot be stored/displayed due to sensitive info */
+        MASKED_VALUE,
```
jglick: Note that this API design basically precludes a finer-grained behavior if that is desirable in the future. For example, if we decide that it is valuable to retain everything except the actual password in a step like

```groovy
withCredentials(…) {
  sh "mvn -DscmPass=$SCMPASS -B release:prepare release:perform"
}
```

(not itself a great example since use of `'` over `"` is more advisable here), the implementation would have no way of representing that: it could return a `String` that is partially masked out, but callers could not know this status.

Similarly, if we decided we wanted

```groovy
sh """
make world
if [ -f errors ]
then
  echo long and complicated script follows, far beyond maximum parameter length
fi
"""
```

to be translated to something like

```java
"make world\n…"
```

or whatever, rather than returning `OVERSIZE_VALUE`, an implementation could not do that without also informing clients that this had been done and that they should not expect the argument to be complete.

Whether this is really a problem, I am not sure, but wanted to flag the limitation.

svanoort: @jglick I'm not sure why you're complaining about lack of support for a feature we don't have a concrete need for yet?

There's nothing in the API code that would *prevent* us from adding support for additional metadata about arguments -- although obviously that would require adding additional APIs. 

Also I do not understand on which grounds you are considering this change blocked, @jglick.  Please mark it with a bug appropriately so I know which comments must be addressed.

jglick:
Inclusion of `MAX_STRING_LENGTH` here rather than downstream in the implementation.

jglick:
Also this should be handling metasteps as discussed downstream.

jglick:
> this should be handling metasteps

Sorry, I think that was meant rather for https://github.com/jenkinsci/workflow-step-api-plugin/pull/24.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
stephenc: I am still seeing this constant here... @svanoort what is you decision about this?

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
stephenc: @svanoort were you going to provide a defauly impl?

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
svanoort: @stephenc I've probably written and deleted replies 5 times now, but here's where I stand: I strongly disagree with pushing this down to the implementing concrete classes.  The reason something was put here is what @michaelneale described -- it needs to be at the lowest level to avoid people hanging themselves if they ever implement the API. 

Will it be just a constant though?  Probably not - it'll get more fleshed out here I think.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
svanoort: The generics actually don't play nicely in this specific case.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
svanoort: No, I'm open to a concrete implementation that just stores a single argument as a field rather than holding a full map.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
jglick: Well, depending on the Java source level you may need to use

```java
Collections.<Map<String, Object>>emptyMap()
```

IIRC in a newer level the inference is improved.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -24,65 +24,137 @@
...
+        /** Denotes an object that is too big to retain, such as strings exceeding {@link #MAX_RETAINED_LENGTH} */
         OVERSIZE_VALUE
     }
 
-    /** Largest string we'll persist in the Step, for performance reasons */
-    public static final int MAX_STRING_LENGTH = 1024;
+    /** Largest String, Collection, or array size we'll retain -- provides a rough size limit on any single field.
+     *  Set to 0 or -1 to remove length limits.
+     */
+    private static final int MAX_RETAINED_LENGTH = Integer.getInteger(ArgumentsAction.class.getName()+".maxRetainedLength", 1024);;
```
jglick:
```java
// TODO Jenkins 2.4+ use SystemProperties
```

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -24,65 +24,137 @@
...
         MASKED_VALUE,
 
-        /** Denotes an object that is too big to retain, such as strings exceeding {@link #MAX_STRING_LENGTH} */
+        /** Denotes an object that is too big to retain, such as strings exceeding {@link #MAX_RETAINED_LENGTH} */
         OVERSIZE_VALUE
     }
 
-    /** Largest string we'll persist in the Step, for performance reasons */
-    public static final int MAX_STRING_LENGTH = 1024;
+    /** Largest String, Collection, or array size we'll retain -- provides a rough size limit on any single field.
```
jglick: For `String` lengths, would be good to clarify that you are referring to character count, not byte count.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -24,65 +24,137 @@
...
+
+    /**
+     * Provides a basic check if an object contains any excessively large single elements.
+     * This recursively checks for String length, Map size, and array/collection size.
+     * We're not trying to cap total size just identify problem elements.
+     * @param o Object to check, with null allowed since we may see null inputs
+     * @param maxElements Max number of elements for a collection/map or characters in a string, or <0 to ignore length rules.
+     * @return True if object (or one of the contained objects) exceeds maxElements size.
+     */
+    public static boolean isOverSized(@CheckForNull Object o, final int maxElements) {
```
jglick: AFAICT this is something that is only going to be called by the implementation (currently in `workflow-cps`; sooner or later will need to be moved into `workflow-support`), so it would better be moved downstream—there is no reason for it to be in `workflow-api`.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -24,65 +24,137 @@
...
+            if (o instanceof Object[]){
+                for (Object element : (Object[])o) {
+                    if (isOverSized(element, maxElements)) {
+                        return true;
+                    }
+                }
+            }
+            if (o instanceof Map) {
+                for(Map.Entry<?,?> entry : ((Map<?,?>)o).entrySet()) {
+                    if (isOverSized((Object)(entry.getKey()), maxElements) || isOverSized((Object)(entry.getValue()), maxElements)) {
```
jglick: Why the no-op cast to `Object`?

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -24,65 +24,137 @@
...
+        } else {
+            return Collections.unmodifiableMap(getArgumentsInternal());
+        }
     }
 
+    /**
+     * Get the map of arguments supplied to instantiate the {@link Step} run in the {@link FlowNode} given
+     * or null if the arguments were not stored or the FlowNode was not a step.
+     *
+     * Internally
```
jglick: Huh?

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -24,65 +24,137 @@
...
+     * @param n FlowNode to fetch Step arguments for (including placeholders for masked values).
+     * @return
```
jglick: Better to just omit empty tags.

jglick:
Hmm, somehow GitHub got confused about outdated diffs, recommenting…

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,269 @@
...
+    /** Used as a placeholder marker for {@link Step} arguments not stored for various reasons. */
+    public enum NotStoredReason {
+        /** Denotes an unsafe value that cannot be stored/displayed due to sensitive info. */
+        MASKED_VALUE,
+
+        /** Denotes an object that is too big to retain, such as strings exceeding {@link #MAX_RETAINED_LENGTH} */
+        OVERSIZE_VALUE
+    }
+
+    /** Largest String, Collection, or array size we'll retain -- provides a rough size limit on any single field.
```
jglick: For `String` lengths, would be good to clarify that you are referring to character count, not byte count.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,269 @@
...
+     *  Set to 0 or -1 to remove length limits.
+     */
+    private static final int MAX_RETAINED_LENGTH = Integer.getInteger(ArgumentsAction.class.getName()+".maxRetainedLength", 1024);;
```
jglick:
```java
// TODO Jenkins 2.4+ use SystemProperties
```

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,269 @@
...
+
+    /**
+     * Provides a basic check if an object contains any excessively large single elements.
+     * This recursively checks for String length, Map size, and array/collection size.
+     * We're not trying to cap total size just identify problem elements.
+     * @param o Object to check, with null allowed since we may see null inputs
+     * @param maxElements Max number of elements for a collection/map or characters in a string, or &lt; 0 to ignore length rules.
+     * @return True if object (or one of the contained objects) exceeds maxElements size.
+     */
+    public static boolean isOverSized(@CheckForNull Object o, final int maxElements) {
```
jglick: AFAICT this is something that is only going to be called by the implementation (currently in `workflow-cps`; sooner or later will need to be moved into `workflow-support`), so it would better be moved downstream—there is no reason for it to be in `workflow-api`.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -0,0 +1,269 @@
...
+     *  Set to 0 or -1 to remove length limits.
+     */
+    private static final int MAX_RETAINED_LENGTH = Integer.getInteger(ArgumentsAction.class.getName()+".maxRetainedLength", 1024);;
+
+    /**
+     * Provides a basic check if an object contains any excessively large single elements.
+     * This recursively checks for String length, Map size, and array/collection size.
+     * We're not trying to cap total size just identify problem elements.
+     * @param o Object to check, with null allowed since we may see null inputs
+     * @param maxElements Max number of elements for a collection/map or characters in a string, or &lt; 0 to ignore length rules.
```
jglick: BTW can use `{@code < 0}`

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
svanoort: Generics complaints.   There's probably more than one way to solve it, but this does the trick.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
svanoort: @jglick I was actually considering going the *opposite* route and moving secret masking logic up to the API (along with testing) and having workflow-cps just provide the discrete portions & overrides needed.  

This would enable me to simplify the logic in workflow-cps, and *optionally* later provide a more trivial and lightweight implementation for cases where there's a single argument.

jglick: > This would enable me to simplify the logic in workflow-cps

By making it more complex here I guess.

My point is that the decision about whether a given argument is reasonable and safe to store or not is potentially depend on the storage of the `FlowNode`. Under the current XML-based storage, defined in `workflow-support`, we need to be pretty careful. Under a possible future better engine, we could raise the limits, or take other factors into consideration. But it is the concrete subclass of `ArgumentsAction`—today in `workflow-cps`, really should be in `workflow-support` for architectural reasons (nothing specific to Groovy AFAICT)—and the code which attaches that action to a step node which is best placed to make that decision. Therefore it does not belong in the API.

(Even if there were no expectation that the policy might change in the future, it would be preferable to move it out of the API plugin anyway, since there is only a single implementation, and other plugins depending on `workflow-api` should not be calling these utilities: it is an implementation detail what is or is not masked or truncated.)

pom.xml
```diff
@@ -68,7 +69,8 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-step-api</artifactId>
-            <version>2.7</version>
+            <!-- Temporary version, remove once we have merged and released -->
+            <version>2.10-stepstorage4-beta</version>
```
jglick: This is https://github.com/jenkinsci/workflow-step-api-plugin/pull/24 IIUC.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -61,7 +61,7 @@
     /** Largest String, Collection, or array size we'll retain -- provides a rough size limit on any single field.
      *  Set to 0 or -1 to remove length limits.
      */
-    private static final int MAX_RETAINED_LENGTH = Integer.getInteger(ArgumentsAction.class.getName()+".maxRetainedLength", 1024);;
+    protected static final int MAX_RETAINED_LENGTH = Integer.getInteger(ArgumentsAction.class.getName()+".maxRetainedLength", 1024);
```
jglick:
```java
// TODO 2.4+ use SystemProperties
```

Should be non`final`.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -110,6 +110,15 @@ public static boolean isOverSized(@CheckForNull Object o, final int maxElements)
         return false;
     }
 
+    /**
+     * Check for single oversized fields much like {@link #isOverSized(Object, int)} but using {@link #MAX_RETAINED_LENGTH}.
+     * @param o Object to check for being oversized or holding an oversized value.
+     * @return True if object contains an oversized input, else false.
+     */
+    protected static boolean isOversized(@CheckForNull Object o) {
+        return isOverSized(o, MAX_RETAINED_LENGTH);
```
jglick: Inconsistent capitalization of `s` vs. `S`.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -110,6 +110,15 @@ public static boolean isOverSized(@CheckForNull Object o, final int maxElements)
         return false;
     }
 
+    /**
+     * Check for single oversized fields much like {@link #isOverSized(Object, int)} but using {@link #MAX_RETAINED_LENGTH}.
+     * @param o Object to check for being oversized or holding an oversized value.
+     * @return True if object contains an oversized input, else false.
+     */
+    protected static boolean isOversized(@CheckForNull Object o) {
```
jglick: Still recommend these methods, and the constant, be removed from the API and moved to `ArgumentsActionImpl`.

PR #28, opened 2017-02-07
Add StepNode to support fetching StepDescriptor at higher level of API

svanoort:
Upstream of https://github.com/jenkinsci/workflow-cps-plugin/pull/104

This permits us to obtain the StepDescriptor without taking a dependency on workflow-cps -- needed for some APIs. 

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/StepNode.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/StepNode.java
@@ -1,0 +1,44 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2013-2017, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/graph/StepNode.java
```diff
@@ -0,0 +1,44 @@
...
+package org.jenkinsci.plugins.workflow.graph;
+
+import org.jenkinsci.plugins.workflow.steps.Step;
+import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
+
+import javax.annotation.CheckForNull;
+
+/**
+ * Optional interface for a {@link FlowNode} that has an associated {@link StepDescriptor}.
+ * Supertype of the equivalent in the workflow-cps to allow access at a higher level of the dependency tree.
```
jglick: Can omit this sentence—will soon be of historical interest anyway.

christ66:
:bee:

PR #30, opened 2017-02-13
Introduced FlowCopier.StandardActions

jglick:
Found that `ParametersAction` and `SCMRevisionAction` were not getting properly copied from one build to the next.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowCopier.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowCopier.java
@@ -24,11 +24,18 @@
 
 package org.jenkinsci.plugins.workflow.flow;
 
+import com.google.common.collect.ImmutableList;
+import hudson.Extension;
 import hudson.ExtensionPoint;
+import hudson.model.Action;
+import hudson.model.ParametersAction;
 import hudson.model.Queue;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/flow/FlowCopier.java
```diff
@@ -64,4 +71,26 @@
...
+
+        // TODO cloned from ReplayAction; consider whether it is appropriate to share these (related but not identical case)
+        private static final Iterable<Class<? extends Action>> COPIED_ACTIONS = ImmutableList.of(
+            ParametersAction.class,
+            SCMRevisionAction.class
+        );
+
+        @Override public void copy(Run<?, ?> original, Run<?, ?> copy, TaskListener listener) throws IOException, InterruptedException {
+            for (Class<? extends Action> type : COPIED_ACTIONS) {
+                Action a = original.getAction(type);
```
svanoort: 🐜  getPersistentAction

svanoort: nvm, ignore the previous comment (deleting it)

svanoort:
🐝

PR #31, opened 2017-02-14
[JENKINS-40771] FlowExecutionList race condition

jglick:
[JENKINS-40771](https://issues.jenkins-ci.org/browse/JENKINS-40771)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
@@ -41,7 +41,7 @@ import org.kohsuke.accmod.restrictions.DoNotUse;
  */
 @Extension
 public class FlowExecutionList implements Iterable<FlowExecution> {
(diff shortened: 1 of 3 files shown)


jglick:
CI failure is from bad checkout credentials.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -109,21 +110,26 @@ private synchronized void load() {
      * are no longer running.
      */
     public synchronized void register(final FlowExecutionOwner self) {
-        load();
```
jglick: The actual fix. The test would frequently (not always) reproduce the bug:

* `register` called with `p #2` (or vice-versa)
   * `load` produces `[]`
   * `saveLater` called with `[p #2]`
* `register` called with `p #3`
   * `load` produces `[]` (oops! because that is what is still on disk)
   * `saveLater` called with `[p #3]`
* `save` called with `[p #2]`
* `save` called with `[p #3]` (clobbering the mention of `p #2`)
* restart
* loaded `[p #3]`
* `p #3` resumes in `sleep`, while `p #2` just sits there until aborted

PR #33, opened 2017-02-22
Giant wad of fixes to the Forkscanner visitor/eventing behaviors

svanoort:
- [x] Fix and testcases for fix of [JENKINS-41685](https://issues.jenkins-ci.org/browse/JENKINS-41685) (duplicate events)
- [x] Testcase and fixes added for [JENKINS-39839](https://issues.jenkins-ci.org/browse/JENKINS-39839) - missing parallel events with nesting
- [x] Document the rules and assumptions for visitor APIs, and add test coverage to verify they are correctly implemented, in depth
- [x] Sanity-test case for [JENKINS-39841](https://issues.jenkins-ci.org/browse/JENKINS-39841) in the TestVisitor -- nulls where there shouldn't be
- [x] Unlogged defect: various bugs with ForkScanner and single-branch parallels (especially in-progress)
- [x] Whole raft of fixes around missing branch events in special edge cases
- [x] Various defects: mismatched branch start/end 
- [x] Clean up & extend javadocs to fully explain the visitor (SAX-like) APIs for graphanalysis, aka Bismuth Level 2, 

EXPLICITLY OUT OF SCOPE:

* Handling "find the last to begin" with nested parallels

**Known issues being addressed along with review changes:**

* Fix an issue with the ChunkEndEvent for the testParallelCorrectEndNode

@reviewbybees especially @vivek and @jglick

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/AbstractFlowScanner.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/AbstractFlowScanner.java
@@ -25,6 +25,7 @@
 package org.jenkinsci.plugins.workflow.graphanalysis;
 
 import com.google.common.base.Predicate;
+import com.google.common.base.Predicates;
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 
@@ -305,6 +306,25 @@ public abstract class AbstractFlowScanner implements Iterable <FlowNode>, Filter
             }
         }
         return nodes;
+    }
+
+    /** Convenience method to get the list all flownodes in the iterator order. */
+    @Nonnull
+    public List<FlowNode> allNodes(@CheckForNull Collection<FlowNode> heads) {
+        if (!setup(heads)) {
+            return Collections.EMPTY_LIST;
+        }
+        List<FlowNode> nodes = new ArrayList<FlowNode>();
+        for (FlowNode f : this) {
+            nodes.add(f);
+        }
+        return nodes;
+    }
+
+    /** Convenience method to get the list of all {@link FlowNode}s for the execution, in iterator order. */
+    @Nonnull
+    public List<FlowNode> allNodes(@CheckForNull FlowExecution exec) {
+        return (exec == null) ? Collections.EMPTY_LIST : allNodes(exec.getCurrentHeads());
     }
 
     /** Syntactic sugar for {@link #filteredNodes(Collection, Collection, Predicate)} with no blackList nodes */
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/BlockChunkFinder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/BlockChunkFinder.java
@@ -8,11 +8,13 @@ import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 
 /**
- * Matches start and end of a block.  Any block
+ * Matches start and end of a block.  Any block!
  * @author Sam Van Oort
  */
 public class BlockChunkFinder implements ChunkFinder {
 
+    /** NOTE: you will need to handle cases where you have a start node where the end node has not been generated yet!
+     *  This means you need to keep nodes around even after hitting the EndNode */
     @Override
     public boolean isStartInsideChunk() {
         return false;
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/BlockChunkFinder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/BlockChunkFinder.java
@@ -13,7 +13,7 @@ import javax.annotation.Nonnull;
  */
 public class BlockChunkFinder implements ChunkFinder {
 
-    /** NOTE: you will need to handle cases where you have a start node where the end node has not been generated yet!
+    /** NOTE: you will need to handle cases where you have a {@link BlockStartNode} where the {@link BlockEndNode} has not been generated yet!
      *  This means you need to keep nodes around even after hitting the EndNode */
     @Override
     public boolean isStartInsideChunk() {
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ChunkFinder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ChunkFinder.java
@@ -6,13 +6,34 @@ import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 
 /**
- * Used to define the start and end of a {@link FlowChunk} to split a {@link org.jenkinsci.plugins.workflow.flow.FlowExecution}
- * (For use with a {@link SimpleChunkVisitor} in the {@link ForkScanner#visitSimpleChunks(SimpleChunkVisitor, ChunkFinder)}
+ * Think of this as setting conditions to mark a region of interest in the graph of {@link FlowNode} from a {@link org.jenkinsci.plugins.workflow.flow.FlowExecution}.
+ * <p>This is used to define a linear "chunk" from the graph of FlowNodes returned by a {@link ForkScanner}, after it applies ordering.
+ * <p>This is done by invoking {@link ForkScanner#visitSimpleChunks(SimpleChunkVisitor, ChunkFinder)}.
+ * <p>Your {@link SimpleChunkVisitor} will receive callbacks about chunk boundaries on the basis of the ChunkFinder.
+ *   It isresponsible for tracking the state based on events fired
+ *
+ * <p><p><em>Common uses:</em>
+ * <ul>
+ *     <li>Find all nodes within a specific block type, such the block created by a timeout block, 'node' (executor) block, etc</li>
+ *     <li>Find all nodes between specific markers, such as labels, milestones, or steps generating an error</li>
+ * </ul>
+ *
+ * <p><em>Implementation Notes:</em>
+ * <ul>
+ *     <li>This can be used to detect both block-delimited regions of interest and marker-based regions</li>
+ *     <li>Block-delimited regions should END when encountering the right kind of {@link org.jenkinsci.plugins.workflow.graph.BlockEndNode}
+ *         and start when seeing the right kind of {@link org.jenkinsci.plugins.workflow.graph.BlockStartNode}</li>
+ *     <li>Marker-based regions should start when you find the marker, and END when the previous node is a marker</li>
+ *     <li>If you need to handle both for the same set of criteria... good grief. See the StageChunkFinder in the pipeline-graph-analysis plugin.</li>
+ * </ul>
+ *
  * @author Sam Van Oort
  */
 public interface ChunkFinder {
 
-    /** If true, a chunk is implicitly created whenever we begin */
+    /** If true, a chunk is implicitly created whenever we begin.
+     *  <p>If you are matching the start/end of a block, should always return false.
+     *  <p>If you are trying to match markers (such as single-node labels or milestones), should always be true. */
     boolean isStartInsideChunk();
 
     /**
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ChunkFinder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ChunkFinder.java
@@ -10,12 +10,12 @@ import javax.annotation.Nonnull;
  * <p>This is used to define a linear "chunk" from the graph of FlowNodes returned by a {@link ForkScanner}, after it applies ordering.
  * <p>This is done by invoking {@link ForkScanner#visitSimpleChunks(SimpleChunkVisitor, ChunkFinder)}.
  * <p>Your {@link SimpleChunkVisitor} will receive callbacks about chunk boundaries on the basis of the ChunkFinder.
- *   It isresponsible for tracking the state based on events fired
+ *   It is responsible for tracking the state based on events fired
  *
  * <p><p><em>Common uses:</em>
  * <ul>
- *     <li>Find all nodes within a specific block type, such the block created by a timeout block, 'node' (executor) block, etc</li>
- *     <li>Find all nodes between specific markers, such as labels, milestones, or steps generating an error</li>
+ *     <li>Find all {@link FlowNode}s within a specific block type, such the block created by a timeout block, 'node' (executor) block, etc</li>
+ *     <li>Find all {@link FlowNode}s between specific markers, such as labels, milestones, or steps generating an error</li>
  * </ul>
  *
  * <p><em>Implementation Notes:</em>
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
@@ -65,7 +65,9 @@ public final class FlowScanningUtils {
     /** Sorts flownodes putting the one begun last (oldest startTime) at the end, with null times last
      *  because likely they represent the newest nodes with a {@link TimingAction} not attached yet. */
     public static final Comparator<FlowNode> TIME_ORDER_COMPARATOR = new Comparator<FlowNode>() {
+
         /** Implements null checking because the use of this method will not easily permit FindBugs verification on NonNull*/
+        @Override
         public int compare(@CheckForNull FlowNode first, @CheckForNull FlowNode second) {
             if (first == null || second == null) {
                 return 0;  // Sorting by start time is 100% irrelevant
@@ -83,7 +85,8 @@ public final class FlowScanningUtils {
     };
 
     public static final Comparator<FlowNode> ID_ORDER_COMPARATOR = new Comparator<FlowNode>() {
-        /** Implements null checking because the use of this method will not easily permit FindBugs verification on NonNull*/
+        /** Implements null checking because it reduces the amount of null handling needed to use this */
+        @Override
         public int compare(@CheckForNull FlowNode first, @CheckForNull FlowNode second) {
             if (first == null || second == null) {
                 return 0;
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
@@ -69,8 +69,12 @@ public final class FlowScanningUtils {
         /** Implements null checking because the use of this method will not easily permit FindBugs verification on NonNull*/
         @Override
         public int compare(@CheckForNull FlowNode first, @CheckForNull FlowNode second) {
-            if (first == null || second == null) {
-                return 0;  // Sorting by start time is 100% irrelevant
+            if (first == null && second == null) {
+                return 0;
+            } else if (first == null) {
+                return 1;
+            } else if (second == null) {
+                return -1;
             }
             TimingAction timingFirst = first.getPersistentAction(TimingAction.class);
             TimingAction timingSecond = second.getPersistentAction(TimingAction.class);
@@ -88,8 +92,12 @@ public final class FlowScanningUtils {
         /** Implements null checking because it reduces the amount of null handling needed to use this */
         @Override
         public int compare(@CheckForNull FlowNode first, @CheckForNull FlowNode second) {
-            if (first == null || second == null) {
+            if (first == null && second == null) {
                 return 0;
+            } else if (first == null) {
+                return 1;
+            } else if (second == null) {
+                return -1;
             }
             try {
                 int id1 = Integer.parseInt(first.getId());
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
@@ -82,6 +82,22 @@ public final class FlowScanningUtils {
         };
     };
 
+    public static final Comparator<FlowNode> ID_ORDER_COMPARATOR = new Comparator<FlowNode>() {
+        /** Implements null checking because the use of this method will not easily permit FindBugs verification on NonNull*/
+        public int compare(@CheckForNull FlowNode first, @CheckForNull FlowNode second) {
+            if (first == null || second == null) {
+                return 0;
+            }
+            try {
+                int id1 = Integer.parseInt(first.getId());
+                int id2 = Integer.parseInt(second.getId());
+                return Integer.compare(id1, id2);
+            } catch (NumberFormatException nfe) {
+                return first.getId().compareTo(second.getId());
+            }
+        };
+    };
+
     /**
      * Returns all {@link BlockStartNode}s enclosing the given FlowNode, starting from the inside out.
      * This is useful if we want to obtain information about its scope, such as the workspace, parallel branch, or label.
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
@@ -27,10 +27,13 @@ package org.jenkinsci.plugins.workflow.graphanalysis;
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
 import hudson.model.Action;
+import org.jenkinsci.plugins.workflow.actions.TimingAction;
 import org.jenkinsci.plugins.workflow.graph.BlockStartNode;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 
+import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+import java.util.Comparator;
 
 /**
  * Library of common functionality when analyzing/walking flow graphs
@@ -58,6 +61,26 @@ public final class FlowScanningUtils {
 
     // Default predicates, which may be used for common conditions
     public static final Predicate<FlowNode> MATCH_BLOCK_START = (Predicate)Predicates.instanceOf(BlockStartNode.class);
+
+    /** Sorts flownodes putting the one begun last (oldest startTime) at the end, with null times last
+     *  because likely they represent the newest nodes with a {@link TimingAction} not attached yet. */
+    public static final Comparator<FlowNode> TIME_ORDER_COMPARATOR = new Comparator<FlowNode>() {
+        /** Implements null checking because the use of this method will not easily permit FindBugs verification on NonNull*/
+        public int compare(@CheckForNull FlowNode first, @CheckForNull FlowNode second) {
+            if (first == null || second == null) {
+                return 0;  // Sorting by start time is 100% irrelevant
+            }
+            TimingAction timingFirst = first.getPersistentAction(TimingAction.class);
+            TimingAction timingSecond = second.getPersistentAction(TimingAction.class);
+            if (timingFirst != null && timingSecond != null) {
+                return Long.compare(timingFirst.getStartTime(), timingSecond.getStartTime());
+            } else if (timingFirst == null && timingSecond == null) {
+                return 0;
+            } else { // Only one is null, that one should return the greater value
+                return (timingSecond == null) ? -1 : 1 ;
+            }
+        };
+    };
 
     /**
      * Returns all {@link BlockStartNode}s enclosing the given FlowNode, starting from the inside out.
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
@@ -25,7 +25,6 @@
 package org.jenkinsci.plugins.workflow.graphanalysis;
 
 import com.google.common.base.Predicate;
-import com.google.common.base.Predicates;
 import com.google.common.collect.Lists;
 import org.jenkinsci.plugins.workflow.actions.ThreadNameAction;
 import org.jenkinsci.plugins.workflow.actions.TimingAction;
@@ -154,10 +153,9 @@ public class ForkScanner extends AbstractFlowScanner {
             return object != null && object instanceof IsParallelPredicate;
         }
     }
-    
-    // A bit of a dirty hack, but it works around the fact that we need trivial access to classes from workflow-cps
-    // For this and only this test. So, we load them from a context that is aware of them.
-    // Ex: workflow-cps can automatically set this correctly. Not perfectly graceful but it works.
+
+    /** Originally a workaround to deal with needing the {@link StepDescriptor} to determine if a node is a parallel start
+     *  Now tidily solved by {@link IsParallelPredicate}*/
     private static Predicate<FlowNode> parallelStartPredicate = new IsParallelPredicate();
 
     // Invoke this passing a test against the ParallelStep conditions
@@ -619,6 +617,7 @@ public class ForkScanner extends AbstractFlowScanner {
     /** Trivial sorting of the current in-progress branches (See issue JENKINS-38536)... does not handle nesting correctly though */
     void sortParallelByTime() {
         // FIXME add nesting support by storing a full tree structure for branches and not the overly complex queue system
+        // FIXME this is horribly broken!!
         if (this.currentParallelStart == null) {  // Not in parallel
             return;
         }
@@ -640,7 +639,7 @@ public class ForkScanner extends AbstractFlowScanner {
         // We can't just  fire the extra chunkEnd event
         // We need to look at the parallels structure, and for each parallel re-sort the nodes by their timing info...
         // IFF they are open when beginning (if complete, it is irrelevant)
-        sortParallelByTime();
+//        sortParallelByTime();
         boolean needsEnd = false;
         if (finder.isStartInsideChunk()) {
             needsEnd = true;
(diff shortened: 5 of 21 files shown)


src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -614,20 +778,86 @@ private void testParallelFindsLast(WorkflowJob job, String semaphoreName) throws
...
         scan.visitSimpleChunks(visitor, labelFinder);
-        TestVisitor.CallEntry entry = visitor.calls.get(0);
-        Assert.assertEquals(TestVisitor.CallType.CHUNK_END, entry.type);
-        FlowNode lastNode = run.getExecution().getNode(Integer.toString(entry.ids[0]));
-        Assert.assertEquals("Wrong End Node: ("+lastNode.getId()+") "+lastNode.getDisplayName(), "semaphore", lastNode.getDisplayFunctionName());
+        TestVisitor.CallEntry parallelEnd = visitor.calls.get(0);
+        Assert.assertEquals(TestVisitor.CallType.PARALLEL_END, parallelEnd.type);
 
+        TestVisitor.CallEntry entry = visitor.calls.get(1);
+        //FIXME figure out why the chunk_end event doesn't fire, and hack in a way to work around needing to order events here.
```
svanoort: Yes, this is the item in my to-dos.  I'm addressing it along with the review feedback because or a PR this size there will be some.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/AbstractFlowScanner.java
```diff
@@ -307,6 +308,25 @@ public FlowNode findFirstMatch(@CheckForNull FlowExecution exec, @Nonnull Predic
         return nodes;
     }
 
+    /** Convenience method to get the list all flownodes in the iterator order. */
+    @Nonnull
+    public List<FlowNode> allNodes(@CheckForNull Collection<FlowNode> heads) {
+        if (!setup(heads)) {
+            return Collections.EMPTY_LIST;
```
jglick: Always use factories like `emptyList()` rather than the constants, so you can avoid `@SuppressWarnings("unchecked")`.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/BlockChunkFinder.java
```diff
@@ -8,11 +8,13 @@
...
 
 /**
- * Matches start and end of a block.  Any block
+ * Matches start and end of a block.  Any block!
  * @author Sam Van Oort
  */
 public class BlockChunkFinder implements ChunkFinder {
 
+    /** NOTE: you will need to handle cases where you have a start node where the end node has not been generated yet!
+     *  This means you need to keep nodes around even after hitting the EndNode */
```
jglick: `{@link BlockEndNode}`?

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ChunkFinder.java
```diff
@@ -6,13 +6,34 @@
 import javax.annotation.Nonnull;
 
 /**
- * Used to define the start and end of a {@link FlowChunk} to split a {@link org.jenkinsci.plugins.workflow.flow.FlowExecution}
- * (For use with a {@link SimpleChunkVisitor} in the {@link ForkScanner#visitSimpleChunks(SimpleChunkVisitor, ChunkFinder)}
+ * Think of this as setting conditions to mark a region of interest in the graph of {@link FlowNode} from a {@link org.jenkinsci.plugins.workflow.flow.FlowExecution}.
+ * <p>This is used to define a linear "chunk" from the graph of FlowNodes returned by a {@link ForkScanner}, after it applies ordering.
+ * <p>This is done by invoking {@link ForkScanner#visitSimpleChunks(SimpleChunkVisitor, ChunkFinder)}.
+ * <p>Your {@link SimpleChunkVisitor} will receive callbacks about chunk boundaries on the basis of the ChunkFinder.
+ *   It isresponsible for tracking the state based on events fired
```
jglick: missing space, `.`

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ChunkFinder.java
```diff
@@ -6,13 +6,34 @@
...
+ *
+ * <p><p><em>Common uses:</em>
+ * <ul>
+ *     <li>Find all nodes within a specific block type, such the block created by a timeout block, 'node' (executor) block, etc</li>
```
jglick: `{@code node}` etc.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ChunkFinder.java
```diff
@@ -6,13 +6,34 @@
...
+ *     <li>Find all nodes between specific markers, such as labels, milestones, or steps generating an error</li>
```
jglick: Is there any actually any use case for finding all nodes between two `milestone`s? This will not in general be a well-formed subtree.

(For labels, this only applies to the deprecated `stage` mode.)

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ChunkFinder.java
```diff
@@ -6,13 +6,34 @@
...
+ * </ul>
+ *
+ * <p><em>Implementation Notes:</em>
+ * <ul>
+ *     <li>This can be used to detect both block-delimited regions of interest and marker-based regions</li>
+ *     <li>Block-delimited regions should END when encountering the right kind of {@link org.jenkinsci.plugins.workflow.graph.BlockEndNode}
+ *         and start when seeing the right kind of {@link org.jenkinsci.plugins.workflow.graph.BlockStartNode}</li>
+ *     <li>Marker-based regions should start when you find the marker, and END when the previous node is a marker</li>
```
jglick: Is there a reason `END` is capitalized like a constant?

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
```diff
@@ -59,6 +62,42 @@ public boolean apply(FlowNode input) {
     // Default predicates, which may be used for common conditions
     public static final Predicate<FlowNode> MATCH_BLOCK_START = (Predicate)Predicates.instanceOf(BlockStartNode.class);
 
+    /** Sorts flownodes putting the one begun last (oldest startTime) at the end, with null times last
+     *  because likely they represent the newest nodes with a {@link TimingAction} not attached yet. */
+    public static final Comparator<FlowNode> TIME_ORDER_COMPARATOR = new Comparator<FlowNode>() {
+        /** Implements null checking because the use of this method will not easily permit FindBugs verification on NonNull*/
+        public int compare(@CheckForNull FlowNode first, @CheckForNull FlowNode second) {
```
jglick: `@Override`

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
```diff
@@ -59,6 +62,42 @@ public boolean apply(FlowNode input) {
     // Default predicates, which may be used for common conditions
     public static final Predicate<FlowNode> MATCH_BLOCK_START = (Predicate)Predicates.instanceOf(BlockStartNode.class);
 
+    /** Sorts flownodes putting the one begun last (oldest startTime) at the end, with null times last
+     *  because likely they represent the newest nodes with a {@link TimingAction} not attached yet. */
+    public static final Comparator<FlowNode> TIME_ORDER_COMPARATOR = new Comparator<FlowNode>() {
+        /** Implements null checking because the use of this method will not easily permit FindBugs verification on NonNull*/
```
jglick: You can use `@SuppressFBWarnings` if you are confident its analysis is incorrect.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
```diff
@@ -59,6 +62,42 @@ public boolean apply(FlowNode input) {
...
+                return Long.compare(timingFirst.getStartTime(), timingSecond.getStartTime());
+            } else if (timingFirst == null && timingSecond == null) {
+                return 0;
+            } else { // Only one is null, that one should return the greater value
+                return (timingSecond == null) ? -1 : 1 ;
+            }
+        };
+    };
+
+    public static final Comparator<FlowNode> ID_ORDER_COMPARATOR = new Comparator<FlowNode>() {
```
jglick: Please do not do this.

Anyway AFAICT this is unused, so why are you adding it?

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -558,9 +627,18 @@ public static void visitSimpleChunks(@Nonnull Collection<FlowNode> heads, @Nonnu
...
+        ArrayList<FlowNode> sorted = new ArrayList<FlowNode>(heads.size());
+        Collections.sort(sorted, FlowScanningUtils.TIME_ORDER_COMPARATOR);
+        scanner.setup(heads);
+        scanner.visitSimpleChunks(visitor, finder);
+    }
+
     /** Ensures we find the last *begun* node when there are multiple heads (parallel branches)
      *  This means that the simpleBlockVisitor gets the *actual* last node, not just the end of the last declared branch
-     *  (See issue JENKINS-38536)
+     *  ()
```
jglick: Huh?

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -558,9 +627,18 @@ public static void visitSimpleChunks(@Nonnull Collection<FlowNode> heads, @Nonnu
         scanner.visitSimpleChunks(visitor, finder);
     }
 
+    /** This is the API you want if you wish to have sorted chunks */
```
jglick: Will even you have any idea what this means in a few months? :-)

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/NodeStepNamePredicate.java
```diff
@@ -0,0 +1,58 @@
...
+        this.descriptorId = descriptorId;
+    }
+
+    public String getDescriptorId() {
+        return descriptorId;
+    }
+
+    @Override
+    public boolean apply(@Nullable FlowNode input) {
+        if (input != null && input instanceof StepNode) {
```
jglick: Null check is redundant: `instanceof` is always false for null.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/NodeStepTypePredicate.java
```diff
@@ -0,0 +1,63 @@
...
+import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/** Predicate that matches {@link FlowNode}s (specifically {@link StepNode}s) with a specific {@link StepDescriptor} type. */
+public final class NodeStepTypePredicate implements Predicate<FlowNode> {
+    StepDescriptor stepDescriptor;
+
+    public NodeStepTypePredicate(@Nonnull Step step) {
```
jglick: Why would you use this?

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ChunkFinderWithoutChunks.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+
+/**
+ * For test use: a ChunkFinder that never returns chunks, to use in testing parallel handling only.
+ */
+public class ChunkFinderWithoutChunks implements ChunkFinder {
```
jglick: This sounds like a Sartre play.

jglick:
BTW a repository outage seems to be responsible for CI failures.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
svanoort: Probably not :)  This is one reason review is helpful....

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ChunkFinderWithoutChunks.java
svanoort: @jglick Maybe not a play, but the ChunkFinderWithout chunks follows up on Being and Nothingness, exploring the metaphysical ramifications of finding Chunks without Finding them. 

Or maybe I just couldn't come up with a clever name for it.  Ideas?  Maybe NoOpChunkFinder?

svanoort:
Clopen for CI issues?

jglick:
Last build finished on one CI builder:

    Dead store to possibleOutput in org.jenkinsci.plugins.workflow.graphanalysis.ForkScanner.next(FlowNode, Collection) [org.jenkinsci.plugins.workflow.graphanalysis.ForkScanner] At ForkScanner.java:[line 584] DLS_DEAD_LOCAL_STORE

recampbell:
@svanoort The Blue Ocean team has a way for us to test our changes against their pipeline by creating a PR. We need to cut a beta version from this PR and submit it as a PR to the blueocean-plugin repository. That PR builder will then run all the Blue Ocean unit tests & ATH tests against this change which will give us a lot of confidence since there are specific test regressions which this change should turn to green. This way we can release this work in confidence knowing we are making things better.

svanoort:
Close/reopen due to CI falling over.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/SimpleChunkVisitor.java
```diff
@@ -34,25 +34,42 @@
...
+ * <p><p><em>Chunk callback rules:</em>
+ * <ol>
+ *     <li>For a single node, it may have EITHER OR BOTH chunkStart and chunkEnd events</li>
+ *     <li>Every node that doesn't get a startChunk/endChunk callback gets an atomNode callback.</li>
+ *     <li>For {@link ChunkFinder} implementations that match the {@link BlockStartNode} and {@link BlockEndNode} should never have both for a single</li>
+ *     <li>You cannot have multiple of any of the same specific type of callbacks for the same flownode</li>
+ *     <li>You cannot have a atomNode callback AND a start/end for the same flownode (application of the above).</li>
+ * </ol>
+ *
+ * <p>Parallel Structure Callbacks: Zero, One, or Multiple may be invoked for any given FlowNode</p>
```
vivek: @svanoort What do you mean by 'Zero, One or Multiple'? I am pretty sure you don't mean same event multiple times for a FlowNode.

vivek:
@svanoort Went over the code, looks like there are lots of improvements 👍 I am assuming no code change in my implementation using Bismuth, however it will be great to get list of improvements that I can possibly use later to improve my code.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ChunkFinder.java
svanoort: It's necessary for legacy stages, these are just suggestions about other ways to take advantage of that.  I think milestones might actually be useful -- and in this case we needn't guarantee tree completeness.  

This is for marking a sequence of nodes that comprise an interesting chunk -- handling of the tree itself and any further parsing is delegated to the `SimpleChunkVisitor` implementation.  This decouples concern.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ChunkFinder.java
svanoort: emphasis

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
svanoort: There's a valid use case for iota-based sorting, even though Bismuth attempts to replace the inappropriate uses of this with the Time Order comparator or proper graph-driven logic. 

Implementing this means I can point people at it rather than watching them reinvent this particular wheel for the n-th time.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/SimpleChunkVisitor.java
svanoort: @vivek Exactly what I said -- there are special cases where a node may belong to more than one category (I've clarified in the latest commit)

jglick:
> We need to cut a beta version from this PR and submit it as a PR to the blueocean-plugin repository.

No need for a beta release, you can `deploy` a snapshot and submit a downstream PR using the timestamp.

jglick:
> There's a valid use case for iota-based sorting

And what is that? Who would use that?

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
```diff
@@ -59,6 +62,45 @@ public boolean apply(FlowNode input) {
     // Default predicates, which may be used for common conditions
     public static final Predicate<FlowNode> MATCH_BLOCK_START = (Predicate)Predicates.instanceOf(BlockStartNode.class);
 
+    /** Sorts flownodes putting the one begun last (oldest startTime) at the end, with null times last
+     *  because likely they represent the newest nodes with a {@link TimingAction} not attached yet. */
+    public static final Comparator<FlowNode> TIME_ORDER_COMPARATOR = new Comparator<FlowNode>() {
```
rsandell: Friends don't let each other use Guava.

rsandell: But that's too late now :'(

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
```diff
@@ -59,6 +62,45 @@ public boolean apply(FlowNode input) {
...
+
+        /** Implements null checking because the use of this method will not easily permit FindBugs verification on NonNull*/
+        @Override
+        public int compare(@CheckForNull FlowNode first, @CheckForNull FlowNode second) {
+            if (first == null || second == null) {
+                return 0;  // Sorting by start time is 100% irrelevant
```
rsandell: This doesn't look right to me, if one of them is null throu would surely be ordered lower or higher than the other.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -80,8 +82,12 @@ public NodeType getNextType() {
         return nextType;
     }
 
-    /** Used to recognize special nodes */
-    public enum NodeType {
+    /** Used to recognize special nodes
```
rsandell: _nit_ multi line javadoc should be empty after the first line's `/**`

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
svanoort: @rsandell If we'd only had Java 8 functional features to play with  this would use 'em and the streaming APIs ;)

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
svanoort: Yeah, probably better to put nulls at the end since they reflect the newest nodes usually.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
jglick: Just look out for things marked `@Beta`.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
jglick: Why would there be nulls at all? Does not make sense to me.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -397,7 +455,14 @@ protected void setHeads(@Nonnull Collection<FlowNode> heads) {
             currentParallelStartNode = currentParallelStart.forkStart;
             myCurrent = currentParallelStart.unvisited.pop();
             myNext = myCurrent;
-            nextType = NodeType.PARALLEL_BRANCH_END;
+
+            // We may have a start type, so we need to override the beginning type
+            NodeType tempType = getNodeType(myCurrent);
```
rsandell: OK, so you added a nice new method to determine what type the node is, just to manipulate the result and almost not care what it gives you anyway?
Can't say it's wrong just that something smells funny.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -483,7 +548,10 @@ FlowNode hitParallelStart() {
                 }
             }
         } else {
-            throw new IllegalStateException("Hit a BlockStartNode with multiple children, and no record of the start!");
+            if (isWalkingFromFinish()) {  //Incomplete single-branch parallels can do this
+                throw new IllegalStateException("Hit a BlockStartNode with no record of the start!");
+            }
+            return myCurrent.getParents().get(0); // No branches to explore
```
rsandell: Is it guaranteed that the list will be non `null` and contain at least one element?

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/SimpleChunkVisitor.java
```diff
@@ -34,25 +34,45 @@
...
+ *     <li>{@link #atomNode(FlowNode, FlowNode, FlowNode, ForkScanner)} - most nodes, which aren't boundaries of chunks</li>
  * </ul>
  *
- * <em>Important implementation note: multiple callbacks can be invoked for a single node depending on its type.</em>
- * <p>For example, we may capture parallels as chunks.
+ * <p><p><strong>Chunk callback rules:</strong>
+ * <ol>
+ *     <li>For a single node, it may have EITHER OR BOTH chunkStart and chunkEnd events</li>
+ *     <li>Every node that doesn't get a startChunk/endChunk callback gets an atomNode callback.</li>
+ *     <li>For {@link ChunkFinder} implementations that match the {@link BlockStartNode} and {@link BlockEndNode} should never have both for a single</li>
```
rsandell: for a single what?

michaelneale:
I think @vivek is ok with this. If there is a beta we can try with, would be great.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
svanoort: Yes.

svanoort:
@jglick and @rsandell Please could I get a rebee on this?

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
```diff
@@ -59,6 +62,53 @@ public boolean apply(FlowNode input) {
     // Default predicates, which may be used for common conditions
     public static final Predicate<FlowNode> MATCH_BLOCK_START = (Predicate)Predicates.instanceOf(BlockStartNode.class);
 
+    /** Sorts flownodes putting the one begun last (oldest startTime) at the end, with null times last
+     *  because likely they represent the newest nodes with a {@link TimingAction} not attached yet. */
```
jglick: This now seems unused, either in this PR or the downstreams I can see. Added deliberately?

svanoort: Yes, it is generally more useful and correct than sorting by ID (even though people insist on doing that anyway).  It is also intended for use in the 3rd level set of APIs that provide a tree-structured format with navigation -- I may have had to abandon that work, but don't see the point in ripping out the independently useful portions.

svanoort:
Still not removing the ID_ORDER_COMPARATOR, simply because I see it reinvented in too many places.

michaelneale:
FWIW https://github.com/jenkinsci/blueocean-plugin/pull/817 - has a beta of this and other things, and it passed the gauntlet of tests, which is very positive.

svanoort:
@jglick @rsandell Any chance of a quickie re-bee on this one?  I removed the need to configure the parallel start predicate (consumers with a legacy version of workflow-cps would have had to do that anyway, but it appears did not always do so).

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -153,11 +151,13 @@ public boolean apply(@Nullable FlowNode input) {
...
     /** Originally a workaround to deal with needing the {@link StepDescriptor} to determine if a node is a parallel start
      *  Now tidily solved by {@link IsParallelStartPredicate}*/
-    private static Predicate<FlowNode> parallelStartPredicate = new IsParallelStartPredicate();
+    private static final Predicate<FlowNode> parallelStartPredicate = new IsParallelStartPredicate();
 
-    // Invoke this passing a test against the ParallelStep conditions
+    /** Now a complete no-op -- originally this was a workaround for dependency issues with workflow-cps.
+     *  Specifically, requiring classes from workflow-cps to detect if something is a parallel step.
+     */
+    @Deprecated
```
jglick: Is there some downstream PR which deletes this call?

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/NodeStepNamePredicate.java
```diff
@@ -52,6 +58,17 @@ public boolean apply(@Nullable FlowNode input) {
         if (input instanceof StepNode) {
             StepDescriptor sd = ((StepNode) input).getDescriptor();
             return (sd != null) ? descriptorId.equals(sd.getId()) : false;
+        } else if (input != null && !(input instanceof FlowStartNode || input instanceof FlowEndNode)) {
+            try {
+                // Workaround for cases where someone is using the latest workflow-api which has StepNode
+                //  but manages to skip the post-2.26 versions of workflow-cps where it a parent of workflow-cps StepNode.
+                // Technically consumers *were* supposed to call ForkScanner#setParallelStartPredicate, but better to be foolproof.
+                Method getDescriptorMethod = input.getClass().getDeclaredMethod("getDescriptor", null);
```
jglick: Why `getDeclaredMethod`, when this is supposed to have been `public`? And why the non-varargs call? I think you meant just

```java
Method getDescriptorMethod = input.getClass().getMethod("getDescriptor");
```

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/NodeStepNamePredicate.java
```diff
@@ -52,6 +58,17 @@ public boolean apply(@Nullable FlowNode input) {
...
+                StepDescriptor sd = (StepDescriptor)(getDescriptorMethod.invoke(input, null));
+                return  (sd != null && descriptorId.equals(sd.getId()));
+            } catch (NoSuchMethodException | SecurityException | IllegalAccessException | InvocationTargetException e) {
```
jglick: `NoSuchMethodException` is sort of expected (at least in the case of a `FlowNode` implementation coming outside the known plugins). The others, if thrown, indicate something haywire, and the exception should be logged as a warning.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/NodeStepTypePredicate.java
```diff
@@ -51,8 +55,19 @@ public StepDescriptor getStepDescriptor(){
...
             return ((StepNode) input).getDescriptor() == stepDescriptor;
+        } else if (input != null && !(input instanceof FlowStartNode || input instanceof FlowEndNode)) {
+            try {
+                // Workaround for cases where someone is using the latest workflow-api which has StepNode
+                //  but manages to skip the post-2.26 versions of workflow-cps where it a parent of workflow-cps StepNode.
+                // Technically consumers *were* supposed to call ForkScanner#setParallelStartPredicate, but better to be foolproof.
+                Method getDescriptorMethod = input.getClass().getDeclaredMethod("getDescriptor", null);
+                StepDescriptor sd = (StepDescriptor)(getDescriptorMethod.invoke(input, null));
+                return  (sd == stepDescriptor);
+            } catch (NoSuchMethodException | SecurityException | IllegalAccessException | InvocationTargetException e) {
```
jglick: Duplicated code, factor into a utility method.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/NodeStepNamePredicate.java
rsandell: awkward name.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
svanoort: A lot of the internal ones here were removed when I switched the config on this one.  Once this is released I can go to pipeline-graph-analysis and cut a PR to strip out the final ones and bump dependencies.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/NodeStepTypePredicate.java
svanoort: For <10 LoC?  Come on, let's not be silly here! ;)

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
jglick: As a matter of procedure I would suggest doing that ahead of time with a snapshot dependency to make sure you are not forgetting something in this PR.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/NodeStepNamePredicate.java
svanoort: I don't want to accept a Method that has arguments supplied -- it needs to be an exact match.  But yes, the getMethod is more accurate here (changed).

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/NodeStepTypePredicate.java
jglick: Well, 10 LoC I had two comments on. :-)

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/NodeStepNamePredicate.java
```diff
@@ -63,11 +65,13 @@ public boolean apply(@Nullable FlowNode input) {
                 // Workaround for cases where someone is using the latest workflow-api which has StepNode
                 //  but manages to skip the post-2.26 versions of workflow-cps where it a parent of workflow-cps StepNode.
                 // Technically consumers *were* supposed to call ForkScanner#setParallelStartPredicate, but better to be foolproof.
-                Method getDescriptorMethod = input.getClass().getDeclaredMethod("getDescriptor", null);
+                Method getDescriptorMethod = input.getClass().getMethod("getDescriptor", null);
```
jglick: non-varargs call (you should be able to delete `, null`)

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/NodeStepTypePredicate.java
```diff
@@ -62,11 +62,13 @@ public boolean apply(@Nullable FlowNode input) {
                 // Workaround for cases where someone is using the latest workflow-api which has StepNode
                 //  but manages to skip the post-2.26 versions of workflow-cps where it a parent of workflow-cps StepNode.
                 // Technically consumers *were* supposed to call ForkScanner#setParallelStartPredicate, but better to be foolproof.
-                Method getDescriptorMethod = input.getClass().getDeclaredMethod("getDescriptor", null);
+                Method getDescriptorMethod = input.getClass().getMethod("getDescriptor", null);
```
jglick: ditto

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/NodeStepNamePredicate.java
svanoort: @jglick Yes, I suppose I could -- this just makes it more explicit, but isn't really needed.

jglick: IIRC you will sometimes get a warning with this syntax, because it is ambiguous whether you mean `(Class[]) null` or `new Class[] {null}`; the compiler should resolve to the former (the one you want) but it is better not to rely on that. If you do not want to use varargs, then the way to make that explicit is

```java
.getMethod("getDescriptor", new Class[0])
```

PR #36, opened 2017-05-16
[JENKINS-43055] Add FlowExecutionListener ExtensionPoint.

abayer:
[JENKINS-43055](https://issues.jenkins-ci.org/browse/JENKINS-43055)

First attempt at JENKINS-43055, adding a `FlowExecutionListener` extension point that listens for started/paused/completed.

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
@@ -4,7 +4,7 @@ import hudson.ExtensionList;
 import hudson.ExtensionPoint;
 
 /**
- * Listens for significant status updates for a {@link FlowExecution}, such as started running, paused, or completed.
+ * Listens for significant status updates for a {@link FlowExecution}, such as started running or completed.
  *
  * @since 2.14
  * @author Andrew Bayer
@@ -17,14 +17,6 @@ public abstract class FlowExecutionListener implements ExtensionPoint {
      * @param execution The {@link FlowExecution} that has started running or resumed.
      */
     public void onRunning(FlowExecution execution) {
-    }
-
-    /**
-     * Called when a {@link FlowExecution} has paused.
-     *
-     * @param execution The {@link FlowExecution} that has paused.
-     */
-    public void onPaused(FlowExecution execution) {
     }
 
     /**
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
@@ -27,11 +27,4 @@ public abstract class FlowExecutionListener implements ExtensionPoint {
     public void onCompleted(FlowExecution execution) {
     }
 
-    /**
-     * All the registered {@link FlowExecutionListener}s.
-     */
-    public static ExtensionList<FlowExecutionListener> all() {
-        return ExtensionList.lookup(FlowExecutionListener.class);
-    }
-
 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
@@ -1,6 +1,5 @@
 package org.jenkinsci.plugins.workflow.flow;
 
-import hudson.ExtensionList;
 import hudson.ExtensionPoint;
 
 /**
@@ -15,8 +14,9 @@ public abstract class FlowExecutionListener implements ExtensionPoint {
      * Called when a {@link FlowExecution} has started running or resumed.
      *
      * @param execution The {@link FlowExecution} that has started running or resumed.
+     * @param resumed True the execution is resuming, false if it's starting for the first time.
      */
-    public void onRunning(FlowExecution execution) {
+    public void onRunning(FlowExecution execution, boolean resumed) {
     }
 
     /**
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
@@ -1,0 +1,45 @@
+package org.jenkinsci.plugins.workflow.flow;
+
+import hudson.ExtensionList;
+import hudson.ExtensionPoint;
+
+/**
+ * Listens for significant status updates for a {@link FlowExecution}, such as started running, paused, or completed.
+ *
+ * @since 2.14
+ * @author Andrew Bayer
+ */
+public abstract class FlowExecutionListener implements ExtensionPoint {
+
+    /**
+     * Called when a {@link FlowExecution} has started running or resumed.
+     *
+     * @param execution The {@link FlowExecution} that has started running or resumed.
+     */
+    public void onRunning(FlowExecution execution) {
+    }
+
+    /**
+     * Called when a {@link FlowExecution} has paused.
+     *
+     * @param execution The {@link FlowExecution} that has paused.
+     */
+    public void onPaused(FlowExecution execution) {
+    }
+
+    /**
+     * Called when a {@link FlowExecution} has completed.
+     *
+     * @param execution The {@link FlowExecution} that has completed.
+     */
+    public void onCompleted(FlowExecution execution) {
+    }
+
+    /**
+     * All the registered {@link FlowExecutionListener}s.
+     */
+    public static ExtensionList<FlowExecutionListener> all() {
+        return ExtensionList.lookup(FlowExecutionListener.class);
+    }
+
+}
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
@@ -26,8 +26,6 @@ public abstract class FlowExecutionListener implements ExtensionPoint {
     /**
      * Called when a {@link FlowExecution} has resumed.
      *
-     * The {@link FlowExecution} will already have been added to the {@link FlowExecutionList} by this point.
-     *
      * @param execution The {@link FlowExecution} that has resumed.
      */
     public void onResumed(FlowExecution execution) {
@@ -38,8 +36,8 @@ public abstract class FlowExecutionListener implements ExtensionPoint {
      *
      * The {@link FlowExecution} will already have been removed from the {@link FlowExecutionList} by this point,
      * {@link GraphListener.Synchronous#onNewHead(FlowNode)} will have already been called for the {@link FlowEndNode},
-     * {@link FlowEndNode#getResult()} will return non-null, and if the Pipeline has failed,
-     * {@link FlowExecution#getCauseOfFailure()} will return non-null.
+     * {@link FlowExecution#getCurrentHeads()} will have one element, a {@link FlowEndNode}, and if the Pipeline has
+     * failed, {@link FlowExecution#getCauseOfFailure()} will return non-null.
      *
      * @param execution The {@link FlowExecution} that has completed.
      */
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
@@ -49,7 +49,7 @@ public abstract class FlowExecutionListener implements ExtensionPoint {
     /**
      * Fires the {@link #onRunning(FlowExecution)} event.
      */
-    public static void fireRunning(FlowExecution execution) {
+    public static void fireRunning(@Nonnull FlowExecution execution) {
         for (FlowExecutionListener listener : ExtensionList.lookup(FlowExecutionListener.class)) {
             listener.onRunning(execution);
         }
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
```diff
@@ -0,0 +1,45 @@
...
+     *
+     * @param execution The {@link FlowExecution} that has started running or resumed.
+     */
+    public void onRunning(FlowExecution execution) {
+    }
+
+    /**
+     * Called when a {@link FlowExecution} has paused.
+     *
+     * @param execution The {@link FlowExecution} that has paused.
```
jglick: Should not be here. The concept of pausing is specific to `workflow-cps`.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @since 2.14
+ * @author Andrew Bayer
+ */
+public abstract class FlowExecutionListener implements ExtensionPoint {
+
+    /**
+     * Called when a {@link FlowExecution} has started running or resumed.
+     *
+     * @param execution The {@link FlowExecution} that has started running or resumed.
```
jglick: No way to tell the difference?

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
```diff
@@ -0,0 +1,45 @@
...
+     *
+     * @param execution The {@link FlowExecution} that has completed.
+     */
+    public void onCompleted(FlowExecution execution) {
+    }
+
+    /**
+     * All the registered {@link FlowExecutionListener}s.
+     */
+    public static ExtensionList<FlowExecutionListener> all() {
```
jglick: Can be omitted (inlined into callers).

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
abayer: Okiedokie.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
abayer: Ok.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
abayer: I wasn't sure how to tell the difference in `workflow-cps`, so mashed them together.

jglick:
If you are not going to _use_ `GraphListener` as an extension point, it would be better to back that out for now—we can keep it in our pocket for use if and when a use case comes up and there is a downstream PR consuming the API. (IIRC there was some desire in the past to have a global listener which would be notified of every `node` block being entered, for keeping aggregate statistics or whatever.)

abayer:
@jglick Alrightie - I'll move that aside and, once these are merged, I'll open WIP PRs for `workflow-api` and `workflow-cps` with the `GraphListener` extension point that we can revisit whenever we want in the future.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
jglick: This really needs a `boolean` parameter.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
```diff
@@ -0,0 +1,30 @@
...
+     *
+     * @param execution The {@link FlowExecution} that has started running or resumed.
+     */
+    public void onRunning(FlowExecution execution) {
+    }
+
+    /**
+     * Called when a {@link FlowExecution} has completed.
+     *
+     * @param execution The {@link FlowExecution} that has completed.
```
jglick: Should document (and test) that

* `isComplete` will be true at this point
* `getCauseOfFailure` may be used to check for failures
* `FlowEndNode.getResult` is set

Also needs to be documented (and tested) that `GraphListener.onNewHead` is going to report the `FlowEndNode` before this event is fired.

Also whether `FlowExecutionList` contains this `FlowExecution` at the time the start event is called and/or at the time the end event is called.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
abayer: Ok? Still not sure how to distinguish between start/resume calls in `workflow-cps`, but I'll add a boolean here, I guess.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
abayer: Will add this testing over in `workflow-cps`.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
```diff
@@ -0,0 +1,48 @@
...
+ * @since 2.14
+ * @author Andrew Bayer
+ */
+public abstract class FlowExecutionListener implements ExtensionPoint {
+
+    /**
+     * Called when a {@link FlowExecution} has started running or resumed.
+     *
+     * @param execution The {@link FlowExecution} that has started running or resumed.
+     * @param resumed True the execution is resuming, false if it's starting for the first time.
```
jglick: Or just define distinct listener methods. Might be clearer; no strong opinion.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
jglick: Testing sure, but documentation needs to be here.

abayer: So no matter what I do, `GraphListener.onNewHead` for the `FlowEndNode` is fired *after* this, because it's getting called in a separate thread. e.g., it's getting called after `WorkflowRun.finish` completes!

jglick: > it's getting called in a separate thread

`GraphL` is not `Synchronous`, so it will be called some time after the `FlowEndNode` is actually attached. `Synchronous` listeners would be notified of it before `finish`; asynch ones either before or after, arbitrarily, according to when the listener was attached.

This is why I suggested in the other PR that the firing would be better done from `workflow-job`, from `workflow-cps`. It is then more straightforward to reason about exactly when the events are fired.

abayer: More specifically, one of the places `WorkflowRun.finish` gets called is `WorkflowRun.GraphL.onNewHead`, and that gets invoked before any other `GraphListener`, so far as I can tell.

jglick: Probably before any other _asynchronous_ listener.

So, again, having `WorkflowRun` fire the change (probably from `finish` somewhere) would solve the problem neatly.

abayer: Ah-ha, that did the trick.

abayer: So as I mentioned on the JIRA, doing this in `WorkflowRun` means bumping Declarative to Java 8/core 2.60, which is not ideal. I've now got the `workflow-cps` PR behaving the right way - so is that going to be ok?

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
```diff
@@ -12,28 +14,53 @@
...
+     *
+     * @param execution The {@link FlowExecution} that has started running.
+     */
+    public void onRunning(FlowExecution execution) {
+    }
+
+    /**
+     * Called when a {@link FlowExecution} has resumed.
+     *
+     * The {@link FlowExecution} will already have been added to the {@link FlowExecutionList} by this point.
```
jglick: Well, yes, but perhaps misleading because it has been there for a while—this list is persistent.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
```diff
@@ -12,28 +14,53 @@
...
-    public void onRunning(FlowExecution execution, boolean resumed) {
+    public void onResumed(FlowExecution execution) {
     }
 
     /**
      * Called when a {@link FlowExecution} has completed.
      *
+     * The {@link FlowExecution} will already have been removed from the {@link FlowExecutionList} by this point,
+     * {@link GraphListener.Synchronous#onNewHead(FlowNode)} will have already been called for the {@link FlowEndNode},
+     * {@link FlowEndNode#getResult()} will return non-null, and if the Pipeline has failed,
```
jglick: Possibly misleading, since `FlowEndNode.result` is `final`, so it is not like it would have returned null at some earlier point—it is just that the `FlowEndNode` is now available at all, and BTW you might be interested in this method.

Would be good to add `@Nonnull` annotations to `FlowEndNode`.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
jglick: (as discussed separately, it is not the case that we need to bump a core dep)

abayer:
@jglick @svanoort Can we get this merged before the https://github.com/jenkinsci/workflow-api-plugin/pull/26 train starts?

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
```diff
@@ -0,0 +1,73 @@
...
+public abstract class FlowExecutionListener implements ExtensionPoint {
+
+    /**
+     * Called when a {@link FlowExecution} has started running.
+     *
+     * The {@link FlowExecution} will already have been added to the {@link FlowExecutionList} by this point.
+     *
+     * @param execution The {@link FlowExecution} that has started running.
+     */
+    public void onRunning(FlowExecution execution) {
```
svanoort: Recommend NotNull or CheckForNull here, strongly

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /**
+     * Called when a {@link FlowExecution} has resumed.
+     *
+     * @param execution The {@link FlowExecution} that has resumed.
+     */
+    public void onResumed(FlowExecution execution) {
```
svanoort: Recommend NotNull or CheckForNull here, strongly

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
```diff
@@ -0,0 +1,73 @@
...
+     * Called when a {@link FlowExecution} has completed.
+     *
+     * The {@link FlowExecution} will already have been removed from the {@link FlowExecutionList} by this point,
+     * {@link GraphListener.Synchronous#onNewHead(FlowNode)} will have already been called for the {@link FlowEndNode},
+     * {@link FlowExecution#getCurrentHeads()} will have one element, a {@link FlowEndNode}, and if the Pipeline has
+     * failed, {@link FlowExecution#getCauseOfFailure()} will return non-null.
+     *
+     * @param execution The {@link FlowExecution} that has completed.
+     */
+    public void onCompleted(FlowExecution execution) {
```
svanoort: Recommend NotNull or CheckForNull here, strongly

abayer: Good point, doing so right now.

abayer: And done.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
```diff
@@ -0,0 +1,75 @@
...
+    public static void fireResumed(FlowExecution execution) {
+        for (FlowExecutionListener listener : ExtensionList.lookup(FlowExecutionListener.class)) {
+            listener.onResumed(execution);
+        }
+    }
+
+    /**
+     * Fires the {@link #onCompleted(FlowExecution)} event.
+     */
+    public static void fireCompleted(FlowExecution execution) {
```
svanoort: @abayer I'm glad you added annotations to the nonstatic versions, but confused why not on the static fire methods?

abayer: Because I forgot. =)

PR #38, opened 2017-05-24
Making GraphListener an ExtensionPoint

abayer:
Replaces #37 on top of current state of master.

I have now concluded this is worth going forward with. =)

cc @reviewbybees


abayer:
Will do.

jglick:
Are you expecting to use this in https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/156, or…?

abayer:
Yup, actively working on that as we speak.

PR #40, opened 2017-06-06
[JENKINS-26522] Design document for FlowNode metadata

abayer:
[JENKINS-26522](https://issues.jenkins-ci.org/browse/JENKINS-26522)

I've been stuck on how to actually get the `PropagatingTagsAction`
from an individual `FlowNode` to its enclosing `BlockEndNode`, and am
giving up - need help with that.

cc @reviewbybees

### DIFF
--- a/NODE_METADATA_DESIGN.md
+++ b/NODE_METADATA_DESIGN.md
@@ -1,0 +1,49 @@
+## Design proposal for additional node metadata - [JENKINS-26522](https://issues.jenkins-ci.org/browse/JENKINS-26522)
+
+### Problem statement
+
+Before this change, `FlowNode`s only can have a limited amount of metadata. For example, status is simply determined by
+whether there is an `ErrorAction` on the `FlowNode` - if so, it's considered failed, otherwise, it's considered passed.
+There's no way to record or report on more granular status, such as `UNSTABLE` or `ABORTED` for an individual `FlowNode`
+(or block/stage). There also is no way to allow plugins to contribute additional metadata to `FlowNode`s, such as test
+results, so there's no way to report on which particular step or stage contributed which test results. Combined, this
+limits the reporting and visualization capabilities available to Blue Ocean and the Pipeline Stage View significantly.
+
+### Summary of Design
+
+To address this, we will use a new `PersistentAction` similar to `TagsAction` to record that needed additional metadata.
+This new action, named something like `PropagatingTagsAction` to signify that it plays a similar role to `TagsAction`, 
+with the noteworthy difference that its tags are propagated to enclosing blocks. Status in particular will be 
+special-cased in `PropagatingTagsAction` to enforce the rule that status can only get worse, not better. 
+
+`PropagatingTagsAction` will support multiple `String` values for a single tag, to support cases such as aggregating 
+test results from nested steps and blocks. While using a single `String` value with delimiters has been considered, 
+a `List` or `Set` of `String`s avoids potential pitfalls with use of a delimiter character in a tag value. Also, use of
+multiple tags with a common prefix has been considered, but could have problems due to uniqueness of the suffixes 
+across nested steps and blocks, as well as the pain of having to discover all tags with the relevant prefix but not a
+different prefix that contains the first prefix, etc.
+
+The `PropagatingTagsAction` public API will look like the following:
+* `void addTag(String, String)` - add a new tag with a single `String` value, overwriting any existing value for the tag.
+* `void addTag(String, Collection<String>)` - add a new tag with its value set to the contents of the given `Collection`, 
+overwriting any existing value for the tag.
+* `void removeTag(String)` - remove an existing tag and its value.
+* `void appendToTag(String, String)` - append a new `String` to the existing value for the given tag. If the tag is not 
+already present, it will be added.
+* `void appendToTag(String, Collection<String>)` - append all the contents of the given `Collection` to the existing 
+value for the given tag. If the tag is not already present, it will be added.
+* `void propagateTags(PropagatingTagsAction)` - propagates all tags from the given `PropagatingTagsAction` to this one,
+following status rules, etc.
+* `Set<String> getTagValue(String)` - get the value for a given tag. If the tag does not exist, this will return an 
+empty set.
+* `Map<String,Set<String>> getTags()` - returns a copy of the underlying map of tags.
+* `boolean setStatus(Result)` - sets the tag for `FlowNode` status if and only if it's not already set to a worse status 
+than the provided result. Returns true if it was able to set the status, false otherwise.
+* `boolean setStatus(String)` - calls `boolean setStatus(Result)` after calling `Result.fromString(String)` on the provided 
+`String`. Throws an exception if it can't convert the `String` to a valid `Result`.
+* `Result getStatus()` - returns the `Result.fromString(String)` value for the status tag on this `FlowNode`. If not set,
+will return `Result.SUCCESS`.
+
+### TODO
+
+* Determine how to get the propagated tags to the relevant `BlockEndNode`. 


NODE_METADATA_DESIGN.md
```diff
@@ -0,0 +1,49 @@
...
+whether there is an `ErrorAction` on the `FlowNode` - if so, it's considered failed, otherwise, it's considered passed.
+There's no way to record or report on more granular status, such as `UNSTABLE` or `ABORTED` for an individual `FlowNode`
+(or block/stage). There also is no way to allow plugins to contribute additional metadata to `FlowNode`s, such as test
+results, so there's no way to report on which particular step or stage contributed which test results. Combined, this
+limits the reporting and visualization capabilities available to Blue Ocean and the Pipeline Stage View significantly.
+
+### Summary of Design
+
+To address this, we will use a new `PersistentAction` similar to `TagsAction` to record that needed additional metadata.
+This new action, named something like `PropagatingTagsAction` to signify that it plays a similar role to `TagsAction`,
```
svanoort: Maybe ScopedTag action?  Since it has a scope of impact, not just single node

NODE_METADATA_DESIGN.md
```diff
@@ -0,0 +1,49 @@
...
+(or block/stage). There also is no way to allow plugins to contribute additional metadata to `FlowNode`s, such as test
+results, so there's no way to report on which particular step or stage contributed which test results. Combined, this
+limits the reporting and visualization capabilities available to Blue Ocean and the Pipeline Stage View significantly.
+
+### Summary of Design
+
+To address this, we will use a new `PersistentAction` similar to `TagsAction` to record that needed additional metadata.
+This new action, named something like `PropagatingTagsAction` to signify that it plays a similar role to `TagsAction`, 
+with the noteworthy difference that its tags are propagated to enclosing blocks. Status in particular will be 
+special-cased in `PropagatingTagsAction` to enforce the rule that status can only get worse, not better.
```
svanoort: s/can only get worse/can only get worse when moving from single nodes to blocks/g

It can get worse from one step to another (if errors are caught, for example).

NODE_METADATA_DESIGN.md
```diff
@@ -0,0 +1,49 @@
...
+limits the reporting and visualization capabilities available to Blue Ocean and the Pipeline Stage View significantly.
+
+### Summary of Design
+
+To address this, we will use a new `PersistentAction` similar to `TagsAction` to record that needed additional metadata.
+This new action, named something like `PropagatingTagsAction` to signify that it plays a similar role to `TagsAction`, 
+with the noteworthy difference that its tags are propagated to enclosing blocks. Status in particular will be 
+special-cased in `PropagatingTagsAction` to enforce the rule that status can only get worse, not better. 
+
+`PropagatingTagsAction` will support multiple `String` values for a single tag, to support cases such as aggregating
```
svanoort: Not sure if it's worth using a List of string for this single case. 

What about making it <String, Object> ?

NODE_METADATA_DESIGN.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Summary of Design
+
+To address this, we will use a new `PersistentAction` similar to `TagsAction` to record that needed additional metadata.
+This new action, named something like `PropagatingTagsAction` to signify that it plays a similar role to `TagsAction`, 
+with the noteworthy difference that its tags are propagated to enclosing blocks. Status in particular will be 
+special-cased in `PropagatingTagsAction` to enforce the rule that status can only get worse, not better. 
+
+`PropagatingTagsAction` will support multiple `String` values for a single tag, to support cases such as aggregating 
+test results from nested steps and blocks. While using a single `String` value with delimiters has been considered,
```
svanoort: True.

NODE_METADATA_DESIGN.md
```diff
@@ -0,0 +1,49 @@
...
+
+To address this, we will use a new `PersistentAction` similar to `TagsAction` to record that needed additional metadata.
+This new action, named something like `PropagatingTagsAction` to signify that it plays a similar role to `TagsAction`, 
+with the noteworthy difference that its tags are propagated to enclosing blocks. Status in particular will be 
+special-cased in `PropagatingTagsAction` to enforce the rule that status can only get worse, not better. 
+
+`PropagatingTagsAction` will support multiple `String` values for a single tag, to support cases such as aggregating 
+test results from nested steps and blocks. While using a single `String` value with delimiters has been considered, 
+a `List` or `Set` of `String`s avoids potential pitfalls with use of a delimiter character in a tag value. Also, use of
+multiple tags with a common prefix has been considered, but could have problems due to uniqueness of the suffixes
```
svanoort: Yep, agreed 100% - prefixed tags = not good.

NODE_METADATA_DESIGN.md
abayer: That'd work for me - the main thing I cared about was not restricting to solely `String`.

NODE_METADATA_DESIGN.md
abayer: You mean an individual step can have a better status than a previous step, right?

NODE_METADATA_DESIGN.md
abayer: I am profoundly agnostic on naming. =)

NODE_METADATA_DESIGN.md
svanoort: That is correct.  But within a block, overall status can only get worse -- it's like entropy, it only goes in one direction.

NODE_METADATA_DESIGN.md
svanoort: So seeing the APIs laid out here formally (thank you!) actually makes me realize there's a couple questions we need to answer collectively:

* How much logic do we want to associate with a tag?  I've been assuming it's trivial, but if not, then we do want a NodeTagExtension that provides APIs which interact with the TagAction/ScopedTagAction
* How much are we storing custom data here?  My gut is that this needs to be pretty light in general, otherwise we're going to be accumulating a lot of crap in the FlowGraph that might better be attached to the Run.

NODE_METADATA_DESIGN.md
```diff
@@ -0,0 +1,49 @@
...
+* `void appendToTag(String, String)` - append a new `String` to the existing value for the given tag. If the tag is not 
+already present, it will be added.
+* `void appendToTag(String, Collection<String>)` - append all the contents of the given `Collection` to the existing 
+value for the given tag. If the tag is not already present, it will be added.
+* `void propagateTags(PropagatingTagsAction)` - propagates all tags from the given `PropagatingTagsAction` to this one,
+following status rules, etc.
+* `Set<String> getTagValue(String)` - get the value for a given tag. If the tag does not exist, this will return an 
+empty set.
+* `Map<String,Set<String>> getTags()` - returns a copy of the underlying map of tags.
+* `boolean setStatus(Result)` - sets the tag for `FlowNode` status if and only if it's not already set to a worse status
```
svanoort: Probably status APIs need to be an API on the flownode (it's pretty core-level)

NODE_METADATA_DESIGN.md
abayer: Hey, I had that `NodeTagExtension` idea earlier. =) I'm not sure, honestly. Status is definitely a special case, but I don't know if it might not be more common. @i386, @michaelneale - any thoughts?

Custom data - this would be an advantage to the `List<String>` approach: we would at least guarantee that nothing more complex than a `String` could be stored, where `Object` is a little riskier, though we could still restrict it in practice to `String` or `List<String>`. I definitely don't want much here - either cases like status where the `String` itself is all the data involved, or the test result case where there'd be one or more `String`s but they'd just be pointers to more elaborate data on the `Run`...

NODE_METADATA_DESIGN.md
abayer: Makes sense. Playing into the `NodeTagExtension` concept, we could then block setting a tag with a known key except through the status etc APIs, so we could ensure they'd follow the right rules.

jglick:
I do not see why it would be either necessary or desirable for such an action to propagate. A step should record any information it has on its own `FlowNode` and leave it at that. If and when a build is visualized, a higher-level API can choose to aggregate some information across a particular block such as a `stage`, by checking for actions on its descendant (in the tree sense) nodes, and processing that information in a way which is likely to be specific to the kind of metadata in question.

I am not really sure I see why we need any new _generic_ API for this at all, with all the problems arising from untyped `String` tags. `Result`-based status deserves a new `Action`, as would test results, etc.

abayer:
@jglick I'll let @svanoort and/or @vivek talk to those questions.

svanoort:
So, to capture the out-of-band discussion with @jglick 

* There's a need to be able to hand metadata passed through blocks from the outside-in (similarly to how we have `WorkspaceAction`s etc) or inside-out (such as build status)
* We don't necessarily need to code this propagation out of box, just provide a method that can be overridden to handle it.
* Using a generic top-level API avoids some of the cyclic-dependency issues we have when we need to do inspections in workflow-api or workflow-cps that would otherwise depend on Actions contributed by plugins depending on them
* Consolidating metadata on FlowNodes to one Tag/ScopedTagAction reduces the complexity from attaching numerous actions, with the logic around them scattered all over the place
* Programming the iteration logic to collect information that is not explicitly propagated is complex, and storing some of it explicitly greatly simplifies working with the flow graph.  The graphanalysis ("bismuth") facilities make this a lot simpler but separate the collection of this information from the storage of it.

I am afraid he is not convinced yet though -- it appears the sticking point is a lack of consumers of generic APIs beyond status & build result at this time, AFAICT.  Also the classic YAGNI principle (a solid concern).

CC @abayer

abayer:
@svanoort I'll be honest - I don't really follow all that. I think you and/or @jglick may need to handle this after all.

jglick:
Is this still valid?

PR #42, opened 2017-06-20
[JENKINS-44636] Action for tracking node block queued/running status

abayer:
[JENKINS-44636](https://issues.jenkins-ci.org/browse/JENKINS-44636)

Decided it made sense to put this in `workflow-api` to simplify dependencies elsewhere. This is downstream of https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/42.

cc @reviewbybees


src/main/java/org/jenkinsci/plugins/workflow/actions/ExecutorTaskInfoAction.java
```diff
@@ -0,0 +1,103 @@
...
+
+    public boolean isQueued() {
+        return whyBlocked != null && whenStartedOrCanceled == -1;
+    }
+
+    public boolean isLaunched() {
+        return launched;
+    }
+
+    public boolean isCanceled() {
```
jglick: @stephenc should this have two `ll`s? :fire:

src/main/java/org/jenkinsci/plugins/workflow/actions/ExecutorTaskInfoAction.java
```diff
@@ -0,0 +1,103 @@
...
+ */
+public class ExecutorTaskInfoAction extends InvisibleAction implements FlowNodeAction, PersistentAction {
+    private static final long serialVersionUID = 1;
+
+    private String whyBlocked;
+    private boolean launched;
+    // Initialized at -1 for "not started yet".
+    private long whenStartedOrCanceled = -1L;
+
+    private transient FlowNode parent;
```
jglick: Do we really need this for something? Seems unlikely, in which case delete and drop `FlowNodeAction`.

src/main/java/org/jenkinsci/plugins/workflow/actions/ExecutorTaskInfoAction.java
```diff
@@ -0,0 +1,103 @@
...
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+
+/**
+ * Records information for a {@code node} block.
+ */
+public class ExecutorTaskInfoAction extends InvisibleAction implements FlowNodeAction, PersistentAction {
+    private static final long serialVersionUID = 1;
+
+    private String whyBlocked;
```
jglick: Why not a `CauseOfBlockage`? Would allow typesafe access to more information.

abayer: I defer to @svanoort - I was wary of adding too much to the action for fear of performance issues.

src/main/java/org/jenkinsci/plugins/workflow/actions/ExecutorTaskInfoAction.java
stephenc: Yes it should :bug: until this is fixed

abayer: I will concede, despite one `l` being the American spelling. =)

stephenc: http://grammarist.com/spelling/cancel/ Since the majority spelling is "cancelled" and only US English differs, we should go with the majority and use i18n to handle exceptions ;-)

stephenc: @abayer I am only half serious btw (but that does mean I am at least half serious)

src/main/java/org/jenkinsci/plugins/workflow/actions/ExecutorTaskInfoAction.java
abayer: Done.

src/main/java/org/jenkinsci/plugins/workflow/actions/ExecutorTaskInfoAction.java
abayer: @stephenc In fairness, if US English spells it "canceled", that means the majority of native English speakers spell it "canceled". =) But yeah, I have no problem with switching it.

src/main/java/org/jenkinsci/plugins/workflow/actions/ExecutorTaskInfoAction.java
```diff
@@ -0,0 +1,88 @@
...
+
+    public boolean isLaunched() {
+        return launched;
+    }
+
+    public boolean isCancelled() {
+        return !launched && whyBlocked == null && whenStartedOrCancelled > -1;
+    }
+
+    public static boolean isNodeQueued(@Nonnull FlowNode node) {
```
svanoort: Instead of these booleans, why not return an existing queue status or an enum (mapping to Queue statuses) and not a bunch of booleans?

src/main/java/org/jenkinsci/plugins/workflow/actions/ExecutorTaskInfoAction.java
```diff
@@ -0,0 +1,88 @@
...
+import hudson.model.InvisibleAction;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+
+/**
+ * Records information for a {@code node} block.
+ */
+public class ExecutorTaskInfoAction extends InvisibleAction implements PersistentAction {
```
svanoort: :ant:  Any reason we couldn't extend the `WorkspaceActionImpl` or WorkspaceAction to contain these APIs?

src/main/java/org/jenkinsci/plugins/workflow/actions/ExecutorTaskInfoAction.java
```diff
@@ -0,0 +1,88 @@
...
+    private static final long serialVersionUID = 1;
+
+    private String whyBlocked;
+    private boolean launched;
+    // Initialized at -1 for "not started yet".
+    private long whenStartedOrCancelled = -1L;
```
svanoort: :bug:  Can't we obtain this by looking at the flowgraph, specifically the timings of the nested blocks for the ExecutorStep/WorkspaceStep?   The TimingAction should give us when we started IIRC.

src/main/java/org/jenkinsci/plugins/workflow/actions/ExecutorTaskInfoAction.java
svanoort: :bug: I think for representing the states a FlowNode may be in as a set of booleans when it should only have a limited set of discrete states?

src/main/java/org/jenkinsci/plugins/workflow/actions/ExecutorTaskInfoAction.java
abayer: *shrug* I didn't want to overload existing `Action`s?

src/main/java/org/jenkinsci/plugins/workflow/actions/ExecutorTaskInfoAction.java
abayer: Yeah, probably.

src/main/java/org/jenkinsci/plugins/workflow/actions/ExecutorTaskInfoAction.java
abayer: I totally don't grok your last comment. fwiw I don't think we actually have an existing queue status enum anywhere, but we could create one here, yes.

abayer:
@svanoort Yes, the queuing information is theoretically stored elsewhere, but is not viable for Blue Ocean to get at, since it's not stored anywhere in the flow graph. 

re: measurable performance hit from new `Action`s - do we have that performance hit documented anywhere? I'd like to understand it better, since I'm not a fan of overloading existing `Action`s all over the place from a development perspective at all - i.e., `UNSTABLE` status stuff and the like with `TagsAction` resulting in there not actually being a simple API for setting status.

src/main/java/org/jenkinsci/plugins/workflow/actions/ExecutorTaskInfoAction.java
abayer: Oooooh, now I remember why I completely avoided overloading `WorkspaceAction` - it has `@Nonnull` on `getNode()`, `getPath()`, and `getLabels()`. So we'd be breaking contracts if we added new methods for "am I still queued?" and returned `null` on `getNode()` and `getPath()` until we're actually on the agent (`getLabels()` could just return an empty `Set`, so that's not a problem).

svanoort: Ahh, okay, that makes sense then -- but that suggests we should be removing this Action once it has a proper Computer/Workspace to play with, then replacing it with the appropriate WorkspaceAction/etc that has missing info filled in.  Neh?

The behavior of the Action here just feels *wrong* because of this -- but I think this key runtime information should be readily exposed by API somehow (until we get an executor anyway).

src/main/java/org/jenkinsci/plugins/workflow/actions/ExecutorTaskInfoAction.java
svanoort: I mean that rather than booleans, have an enum for the states defined for queued tasks [here](https://github.com/jenkinsci/jenkins/blob/51c46c6cf22a57860c71c7d7236ae30f6baa6651/core/src/main/java/hudson/model/Queue.java#L145). 

And yeah I dunno why that never had an enum for Queue task state since a task should only be in one of the `ItemList`s -- this is a heck of a lot simpler than checking different conditionals

svanoort:
> Yes, the queuing information is theoretically stored elsewhere, but is not viable for Blue Ocean to get at, since it's not stored anywhere in the flow graph.

@abayer So, if we know the `Task` in question (which we should be able to record once running the step, if not already stored), then we can poke the `Queue` to see what it's status is.  What I'd like to do is offer a convenience method for BlueOcean (and others) to call that does this for them, to surface that info.... because you're right, that ought to be exposed. 

> measurable performance hit from new `Action`s - do we have that performance hit documented anywhere

It triggers a full serialization and disk write of the` FlowNode` + all its actions (encapsulated in a `Tag` object) every time an action is added or save invoked.  This is is fairly obvious from the code, but not normally considered problematic because we only attach a limited number of `Action`s to a node and rarely mutate those actions.  This change would break from that.  

Something I intend to address in [JENKINS-30896](https://issues.jenkins-ci.org/browse/JENKINS-30896) or by using a `BulkChange` to lump together `Action` writes until execution begins... but that wouldn't help for your use case anyway.

abayer:
@svanoort Ok - so then the question is do we have the information needed to map to the queue in the flow graph without this change. I'll see what I can figure out.

abayer:
@svanoort So we don't have the `Task` anywhere in the flow graph currently - we'll need to record that somehow in an `Action` on the `FlowNode`, either a new `Action` or `WorkspaceAction`, I guess. Also, we currently map from an `ExecutorStepExecution` to its `PlaceholderTask` via the `ExecutorStepExecution`'s `getContext()` - so we'd need to add some additional identifying information in the `PlaceholderTask` that lets us go from the additional information in the `Action` to the `PlaceholderTask`. Doable, yes, but definitely more complexity than the current implementation, fwiw. 

And re: new `Action`s - so the issue isn't so much how many `Action`s as how often we `save()` `Action`s on a `FlowNode`? i.e., if we're mutating a single `Action` N times, the performance cost isn't going to be much better than if we add N different `Action`s to the `FlowNode` but only write to them once each? (I'd assume the latter would be at least slightly less performant than the former due to the increased number of `Action`s making serialization nominally more expensive, but I could be wrong)

src/main/java/org/jenkinsci/plugins/workflow/actions/ExecutorTaskInfoAction.java
abayer: Ah, yeah, we could definitely remove this once we add the `WorkspaceAction`. That seems reasonable.

abayer:
@svanoort Ah, interesting discovery/remembering of something - we can't actually remove `Action`s!

svanoort:
> i.e., if we're mutating a single Action N times, the performance cost isn't going to be much better than if we add N different Actions to the FlowNode but only write to them once each? (I'd assume the latter would be at least slightly less performant than the former due to the increased number of Actions making serialization nominally more expensive, but I could be wrong)

More or less accurate.  

>  Ah, interesting discovery/remembering of something - we can't actually remove Actions!

`getActions()` and then calling remove on the list *should* work I believe (though there was a bug that was fixed).   `Actionable` doesn't support direct removal though, that is correct.

abayer:
@svanoort Oh riiiiight. I forgot about that. Fixing.

abayer:
@svanoort Also, orthogonal to this particular change, but given that mutating an existing `Action` is as costly (or thereabouts) as adding a new one, is there real value in using `TagsAction` at least for `UNSTABLE` status stuff over a new `Action`? That'd diminish the complexity of code that had to deal with status and would provide a much nicer API...

src/main/java/org/jenkinsci/plugins/workflow/actions/TaskInfoAction.java
```diff
@@ -0,0 +1,94 @@
...
+        CANCELLED,
+        LAUNCHED,
+        UNKNOWN
+    }
+
+    /**
+     * Used to identify the {@link org.jenkinsci.plugins.workflow.steps.StepContext} in the task, so that
+     * its status can be identified.
+     */
+    protected int taskContextHashcode;
```
svanoort: :bug: Remove this and allow the Impl to define how you're storing this -- because the abstract version isn't using it, and shouldn't be tied to this in case we change it later

abayer: On it.

abayer:
@svanoort FYI, that remove did *not* work. Getting
```
java.lang.UnsupportedOperationException
	at java.util.concurrent.CopyOnWriteArrayList$COWIterator.remove(CopyOnWriteArrayList.java:1176)
	at java.util.AbstractCollection.remove(AbstractCollection.java:293)
	at org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution$PlaceholderTask$PlaceholderExecutable.run(ExecutorStepExecution.java:648)
```
So...removing the remove again. =)

svanoort:
Erm @abayer that's a bug in the FlowNode implementation :( - should be an easy fix for us though

abayer:
@svanoort Well, feel free to fix said bug? =)

src/main/java/org/jenkinsci/plugins/workflow/actions/TaskInfoAction.java
```diff
@@ -0,0 +1,84 @@
...
+import hudson.model.Queue;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+
+/**
+ * Records information for a {@code node} block.
+ */
+public abstract class TaskInfoAction extends InvisibleAction implements PersistentAction {
```
jglick: This name is really hard to decipher. What is a “task” in this context and why would you need “info” on it?
 Perhaps `QueueItemAction` would be clearer?

src/main/java/org/jenkinsci/plugins/workflow/actions/TaskInfoAction.java
```diff
@@ -0,0 +1,84 @@
...
+    private static final long serialVersionUID = 1;
+
+    public enum QueueState {
+        QUEUED,
```
jglick: Javadoc!

src/main/java/org/jenkinsci/plugins/workflow/actions/TaskInfoAction.java
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Gets the {@link Queue.Item} for this task, if it exists.
+     *
+     * @return The item, or null if it's not in the queue.
+     */
+    @CheckForNull
+    protected abstract Queue.Item itemInQueue();
+
+    @CheckForNull
+    public String getWhyBlocked() {
```
jglick: I would get rid of this. Callers can call `getWhy` themselves if they want. But really they should be using `CauseOfBlockage`.

src/main/java/org/jenkinsci/plugins/workflow/actions/TaskInfoAction.java
```diff
@@ -0,0 +1,84 @@
...
+
+        return item != null ? item.getWhy() : null;
+    }
+
+    /**
+     * Gets whether this task is currently queued.
+     *
+     * @return True if there's an item in the queue for this task, false otherwise.
+     */
+    private boolean isQueued() {
```
jglick: Only used once. Maybe just inline it.

src/main/java/org/jenkinsci/plugins/workflow/actions/TaskInfoAction.java
```diff
@@ -0,0 +1,84 @@
...
+        if (workspaceAction != null) {
+            return QueueState.LAUNCHED;
+        } else {
+            TaskInfoAction action = node.getPersistentAction(TaskInfoAction.class);
+            if (action != null) {
+                if (action.isQueued()) {
+                    return QueueState.QUEUED;
+                } else {
+                    // Getting here means we queued a task, but it's not in the queue any more and we
+                    // never actually launched an executable, so implicitly it's cancelled.
```
jglick: I think there is a race condition here. `PlaceholderExecutable.run` may have just started.

src/main/java/org/jenkinsci/plugins/workflow/actions/TaskInfoAction.java
```diff
@@ -0,0 +1,84 @@
...
+                    return QueueState.CANCELLED;
+                }
+            } else {
+                return QueueState.UNKNOWN;
+            }
+        }
+    }
+
+    @CheckForNull
+    public static String getWhyBlockedForNode(@Nonnull FlowNode node) {
```
jglick: Again I would rather delete this in favor of

```java
public static @CheckForNull Queue.Item getQueueItem(@Nonnull FlowNode node) {
    TaskInfoAction action = node.getPersistentAction(TaskInfoAction.class);
    return action != null ? action.itemInQueue() : null;
}
```

src/main/java/org/jenkinsci/plugins/workflow/actions/TaskInfoAction.java
```diff
@@ -0,0 +1,84 @@
...
+        UNKNOWN
+    }
+    
+    /**
+     * Gets the {@link Queue.Item} for this task, if it exists.
+     *
+     * @return The item, or null if it's not in the queue.
+     */
+    @CheckForNull
+    protected abstract Queue.Item itemInQueue();
```
jglick: `public` would be fine I think.

src/main/java/org/jenkinsci/plugins/workflow/actions/TaskInfoAction.java
abayer: Works for me.

src/main/java/org/jenkinsci/plugins/workflow/actions/TaskInfoAction.java
abayer: Yessir. =)

src/main/java/org/jenkinsci/plugins/workflow/actions/TaskInfoAction.java
abayer: Now that we're exposing `itemInQueue()`, that seems reasonable.

src/main/java/org/jenkinsci/plugins/workflow/actions/TaskInfoAction.java
abayer: We don't add the action until the task has been scheduled, though?

src/main/java/org/jenkinsci/plugins/workflow/actions/TaskInfoAction.java
abayer: *nod*

src/main/java/org/jenkinsci/plugins/workflow/actions/TaskInfoAction.java
abayer: Done.

src/main/java/org/jenkinsci/plugins/workflow/actions/QueueItemAction.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+public abstract class QueueItemAction extends InvisibleAction implements PersistentAction {
+    private static final long serialVersionUID = 1;
+
+    /**
+     * Possible queue stats for the item associated with this {@link FlowNode}.
+     */
+    public enum QueueState {
+        QUEUED,
+        CANCELLED,
```
jglick: These constants need Javadoc.

src/main/java/org/jenkinsci/plugins/workflow/actions/QueueItemAction.java
```diff
@@ -0,0 +1,76 @@
...
+        WorkspaceAction workspaceAction = node.getPersistentAction(WorkspaceAction.class);
+        if (workspaceAction != null) {
+            return QueueState.LAUNCHED;
+        } else {
+            QueueItemAction action = node.getPersistentAction(QueueItemAction.class);
+            if (action != null) {
+                Queue.Item item = action.itemInQueue();
+                if (item == null) {
+                    // The item has left the queue completely and there isn't a WorkspaceAction.
+                    return QueueState.CANCELLED;
```
jglick: Or it is actually being launched, but just spent >5m in `PlaceholderExecutable.run` prior to attaching `WorkspaceAction`. Possible but seems unlikely.

src/main/java/org/jenkinsci/plugins/workflow/actions/QueueItemAction.java
abayer: Okiedokie.

src/main/java/org/jenkinsci/plugins/workflow/actions/QueueItemAction.java
abayer: Yeah, that's an edge case I'm willing to accept.

abayer:
@svanoort @jglick I'm merging this - can we get a release sometime? Thanks!

abayer:
Oh, wait, I can do a release!

PR #44, opened 2017-06-29
[JENKINS-31582] ArgumentsAction.getResolvedArguments

jglick:
[JENKINS-31582](https://issues.jenkins-ci.org/browse/JENKINS-31582)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
@@ -25,20 +25,25 @@
 package org.jenkinsci.plugins.workflow.actions;
 
 import com.google.common.primitives.Primitives;
-import org.apache.commons.collections.CollectionUtils;
-import org.jenkinsci.plugins.workflow.graph.FlowNode;
-import org.jenkinsci.plugins.workflow.graph.StepNode;
-import org.jenkinsci.plugins.workflow.steps.Step;
-import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
-
-import javax.annotation.CheckForNull;
-import javax.annotation.Nonnull;
+import hudson.PluginManager;
+import hudson.model.Describable;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.CheckForNull;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -128,7 +135,7 @@ public String getUrlName() {
         if (args.isEmpty()) {
             return Collections.<String,Object>emptyMap();
         } else {
-            return Collections.unmodifiableMap(getArgumentsInternal());
+            return Collections.unmodifiableMap(args);
```
jglick: unrelated micro-optimization

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -241,6 +249,7 @@ static boolean checkArgumentsLackPlaceholders(@Nonnull Map<String,Object> namedA
             if (ob instanceof NotStoredReason) {
                 return false;
             }
+            // TODO this would need to also check nested Map/List arguments
```
jglick: Originally tried to use `checkArgumentsLackPlaceholders` to avoid even trying to instantiate `Step`s when some of the arguments were placeholders, but saw that this was not going to work anyway. For example

```groovy
archiveArtifacts '1.txt,2.txt,…,9999.txt'
```

will currently claim to be fully-formed even though the argument here is not stored.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -253,4 +262,47 @@ public boolean isUnmodifiedArguments() {
...
+        Map<String, Object> args = aa.getArgumentsInternal();
+        if (n instanceof StepNode) {
+            StepDescriptor d = ((StepNode) n).getDescriptor();
+            if (d != null) {
+                try {
+                    return resolve(new DescribableModel<>(d.clazz), args).getArguments();
+                } catch (Exception x) { // all sorts of things could go wrong here
+                    LOGGER.log(Level.FINE, "coult not resolve " + args + " for " + d.clazz.getName(), x);
+                    // TODO this does not handle NotStoredReason’s well
+                    // more robust to recursively traverse the tree and use e.g. DescribableModel.resolveClass to populate details
```
jglick: Made a brief attempt to do this but seemed too complicated to take on now, and would require some nontrivial API changes to `structs`; it is a corner case anyway.

PR #45, opened 2017-07-27
[JENKINS-38223] Introduced FlowNode.isActive

jglick:
[JENKINS-38223](https://issues.jenkins-ci.org/browse/JENKINS-38223) and also [JENKINS-26139](https://issues.jenkins-ci.org/browse/JENKINS-26139): a more useful API than `isRunning` for some purposes.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,7 +86,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.9</version>
+            <version>2.11.1</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -62,9 +62,9 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>1.642.3</jenkins.version>
+        <jenkins.version>2.7.3</jenkins.version>
         <no-test-jar>false</no-test-jar>
-        <workflow-support-plugin.version>2.13</workflow-support-plugin.version>
+        <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
     </properties>
     <dependencies>
         <dependency>
@@ -75,7 +75,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>scm-api</artifactId>
-            <version>1.3</version>
+            <version>2.0.8</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -92,7 +92,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>2.27</version>
+            <version>2.33</version>
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -111,7 +111,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-scm-step</artifactId>
-            <version>2.3</version>
+            <version>2.4</version>
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -123,7 +123,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
-            <version>1.25</version>
+            <version>1.27</version>
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -131,6 +131,11 @@
             <artifactId>pipeline-stage-step</artifactId>
             <version>2.2</version>
             <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>structs</artifactId>
+            <version>1.7</version>
         </dependency>
     </dependencies>
 </project>
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
@@ -117,10 +117,10 @@ public abstract class FlowNode extends Actionable implements Saveable {
      * <p>It will be false for a node which still has active children, like a step with a running body.
      * It will also be false for something that has finished but is pending child node creation,
      * such as a completed fork branch which is waiting for the join node to be created.
-     * <p>This can only go from true to false and is a shortcut for {@link FlowExecution#isCurrentHead}.
-     * @see #isActive
+     * <p>This can only go from true to false.
+     * @deprecated Usually {@link #isActive} is what you want. If you really wanted the original behavior, use {@link FlowExecution#isCurrentHead}.
      */
-    @Exported
+    @Deprecated
     public final boolean isRunning() {
         return getExecution().isCurrentHead(this);
     }
@@ -150,6 +150,7 @@ public abstract class FlowNode extends Actionable implements Saveable {
      * Unlike {@link #isRunning}, this behaves intuitively for a {@link BlockStartNode}:
      * it will be considered active until the {@link BlockEndNode} is added.
      */
+    @Exported(name="running")
     public final boolean isActive() {
         if (this instanceof FlowEndNode) { // cf. JENKINS-26139
             return false;
@@ -161,7 +162,7 @@ public abstract class FlowNode extends Actionable implements Saveable {
                 } else {
                     // Need workflow-cps 2.33+ to use the optimization.
                     LOGGER.log(Level.FINE, "falling back to old isActive implementation for {0}", this);
-                    if (isRunning()) {
+                    if (getExecution().isCurrentHead(this)) {
                         return true;
                     }
                     List<FlowNode> headNodes = exec.getCurrentHeads();
@@ -170,7 +171,7 @@ public abstract class FlowNode extends Actionable implements Saveable {
                 }
             }
         } else {
-            return isRunning();
+            return getExecution().isCurrentHead(this);
         }
     }
 
@@ -272,8 +273,9 @@ public abstract class FlowNode extends Actionable implements Saveable {
     @Exported
     public BallColor getIconColor() {
         BallColor c = getError()!=null ? BallColor.RED : BallColor.BLUE;
-        // TODO this should probably also be _anime in case this is a step node with a body and the body is still running (try FlowGraphTable for example)
-        if (isRunning())        c = c.anime();
+        if (isActive()) {
+            c = c.anime();
+        }
         return c;
     }
 
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
@@ -154,6 +154,9 @@ public abstract class FlowNode extends Actionable implements Saveable {
     public final boolean isActive() {
         if (this instanceof FlowEndNode) { // cf. JENKINS-26139
             return false;
+        } else if (this instanceof FlowStartNode) {
+            // TODO currently workflow-cps fails to notify GraphListener on this
+            return !exec.isComplete();
         } else if (this instanceof BlockStartNode) { // cf. JENKINS-38223
             synchronized (unclosedBlocks) {
                 Set<BlockStartNode> blocks = unclosedBlocks.get(exec);
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
@@ -26,6 +26,7 @@ package org.jenkinsci.plugins.workflow.graph;
 
 import com.google.common.collect.ImmutableList;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import hudson.Extension;
 import hudson.model.Action;
 import hudson.model.Actionable;
 import hudson.model.BallColor;
@@ -35,8 +36,12 @@ import java.io.IOException;
 import java.io.ObjectStreamException;
 import java.util.AbstractList;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.WeakHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.logging.Level;
 import static java.util.logging.Level.*;
@@ -47,6 +52,7 @@ import org.jenkinsci.plugins.workflow.actions.ErrorAction;
 import org.jenkinsci.plugins.workflow.actions.LabelAction;
 import org.jenkinsci.plugins.workflow.actions.PersistentAction;
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
+import org.jenkinsci.plugins.workflow.flow.GraphListener;
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.DoNotUse;
@@ -111,10 +117,49 @@ public abstract class FlowNode extends Actionable implements Saveable {
      * It will also be false for something that has finished but is pending child node creation,
      * such as a completed fork branch which is waiting for the join node to be created.
      * <p>This can only go from true to false and is a shortcut for {@link FlowExecution#isCurrentHead}.
+     * @see #isActive
      */
     @Exported
     public final boolean isRunning() {
         return getExecution().isCurrentHead(this);
+    }
+
+    private static final Map<FlowExecution, Set<BlockStartNode>> unclosedBlocks = new WeakHashMap<>();
+    @Restricted(DoNotUse.class)
+    @Extension public static class BlockListener implements GraphListener.Synchronous {
+        @Override public void onNewHead(FlowNode node) {
+            if (node instanceof BlockStartNode || node instanceof BlockEndNode) {
+                FlowExecution exec = node.getExecution();
+                synchronized (unclosedBlocks) {
+                    Set<BlockStartNode> blocks = unclosedBlocks.get(exec);
+                    if (blocks == null) {
+                        unclosedBlocks.put(exec, blocks = new HashSet<>());
+                    }
+                    if (node instanceof BlockStartNode) {
+                        blocks.add((BlockStartNode) node);
+                    } else {
+                        blocks.remove(((BlockEndNode) node).getStartNode());
+                    }
+                }
+            }
+        }
+    }
+    /**
+     * Checks whether a node is still part of the active part of the graph.
+     * Unlike {@link #isRunning}, this behaves intuitively for a {@link BlockStartNode}:
+     * it will be considered active until the {@link BlockEndNode} is added.
+     */
+    public final boolean isActive() {
+        if (this instanceof FlowEndNode) { // cf. JENKINS-26139
+            return false;
+        } else if (this instanceof BlockStartNode) { // cf. JENKINS-38223
+            synchronized (unclosedBlocks) {
+                Set<BlockStartNode> blocks = unclosedBlocks.get(exec);
+                return blocks != null && blocks.contains((BlockStartNode) this);
+            }
+        } else {
+            return isRunning();
+        }
     }
 
     /**
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
@@ -27,6 +27,8 @@ package org.jenkinsci.plugins.workflow.graph;
 import com.google.common.base.Predicates;
 import com.google.common.collect.ImmutableList;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import hudson.Extension;
+import hudson.ExtensionList;
 import hudson.model.Action;
 import hudson.model.Actionable;
 import hudson.model.BallColor;
@@ -36,8 +38,10 @@ import java.io.IOException;
 import java.io.ObjectStreamException;
 import java.util.AbstractList;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -47,6 +51,9 @@ import org.jenkinsci.plugins.workflow.actions.ErrorAction;
 import org.jenkinsci.plugins.workflow.actions.LabelAction;
 import org.jenkinsci.plugins.workflow.actions.PersistentAction;
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
+import org.jenkinsci.plugins.workflow.flow.FlowExecutionListener;
+import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
+import org.jenkinsci.plugins.workflow.flow.GraphListener;
 import org.jenkinsci.plugins.workflow.graphanalysis.LinearBlockHoppingScanner;
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
 import org.kohsuke.accmod.Restricted;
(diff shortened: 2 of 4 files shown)


svanoort:
I'd like to review this before merge if that's okay @jglick ?  (Holding a slot because I'm pretty heavily booked today and it might get merged before I can take a look at it)

jglick:
Sure. There are multiple plausible implementations for this, including the original unoptimized version which seemed to be a critical performance problem; listening for the end node seemed the most straightforward optimization to me.

jglick:
Downstream test failures, checking them.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -110,12 +118,62 @@ protected Object readResolve() throws ObjectStreamException {
...
      * such as a completed fork branch which is waiting for the join node to be created.
      * <p>This can only go from true to false and is a shortcut for {@link FlowExecution#isCurrentHead}.
+     * @see #isActive
      */
     @Exported
     public final boolean isRunning() {
         return getExecution().isCurrentHead(this);
     }
 
+    private static final Map<FlowExecution, Set<BlockStartNode>> unclosedBlocks = new WeakHashMap<>();
```
svanoort: So, how will this behave if the master restarts?    Normally the FlowExecution would be safe to use as a weak key because we only care about running flows, where a strong reference will be held.  But won't this be unpopulated when we restart, meaning we can't refer to it for information on running flows?

svanoort: and by 'information on running flows' I mean that this cache won't have the current unclosed blocks present I think?

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -110,12 +118,62 @@ protected Object readResolve() throws ObjectStreamException {
...
+    @Restricted(DoNotUse.class)
+    @Extension public static class BlockListener implements GraphListener.Synchronous {
+        @Override public void onNewHead(FlowNode node) {
+            if (node instanceof BlockStartNode || node instanceof BlockEndNode) {
+                FlowExecution exec = node.getExecution();
+                synchronized (unclosedBlocks) {
```
svanoort: Ehhhh, any chance we could find a way to avoid all flows with BlockStartNodes and BlockEndNodes contending for this lock?   Maybe make a cache per FlowExecution or something?

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -110,12 +118,62 @@ protected Object readResolve() throws ObjectStreamException {
...
+     * @see #isActive
      */
     @Exported
     public final boolean isRunning() {
         return getExecution().isCurrentHead(this);
     }
 
+    private static final Map<FlowExecution, Set<BlockStartNode>> unclosedBlocks = new WeakHashMap<>();
+    @Restricted(DoNotUse.class)
+    @Extension public static class BlockListener implements GraphListener.Synchronous {
```
svanoort: Overall this isn't a bad way to approach this.  I'd considered the possibility of storing some information on nesting of blocks to facilitate other lookups (with a more complex structure though) but this approach will get the job done nicely.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -110,12 +118,62 @@ protected Object readResolve() throws ObjectStreamException {
...
+        if (this instanceof FlowEndNode) { // cf. JENKINS-26139
+            return false;
+        } else if (this instanceof BlockStartNode) { // cf. JENKINS-38223
+            synchronized (unclosedBlocks) {
+                Set<BlockStartNode> blocks = unclosedBlocks.get(exec);
+                if (blocks != null) {
+                    return blocks.contains((BlockStartNode) this);
+                } else {
+                    // Need workflow-cps 2.33+ to use the optimization.
+                    LOGGER.log(Level.FINE, "falling back to old isActive implementation for {0}", this);
```
svanoort: I like that you did this, a lot.  The implementation is now in workflow-api, which is much more reasonable for this.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -110,12 +118,62 @@ protected Object readResolve() throws ObjectStreamException {
...
+                }
+            }
+        }
+    }
+    /**
+     * Checks whether a node is still part of the active part of the graph.
+     * Unlike {@link #isRunning}, this behaves intuitively for a {@link BlockStartNode}:
+     * it will be considered active until the {@link BlockEndNode} is added.
+     */
+    public final boolean isActive() {
```
svanoort: Love it.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -110,12 +118,62 @@ protected Object readResolve() throws ObjectStreamException {
...
+        if (this instanceof FlowEndNode) { // cf. JENKINS-26139
+            return false;
+        } else if (this instanceof BlockStartNode) { // cf. JENKINS-38223
+            synchronized (unclosedBlocks) {
+                Set<BlockStartNode> blocks = unclosedBlocks.get(exec);
+                if (blocks != null) {
+                    return blocks.contains((BlockStartNode) this);
```
svanoort: Note that while initially it would seem that this check fully covers the case where you have restated the master with an in-progress flow and lost the map, there's a potential to be inside a long running block (for example a parallel running automation)... and then begin another block, which populates the map... but now you're missing the block start for the parallel start.

svanoort: Which would generate a false negative for the parallel being active... if my chain of logic holds up here, and it may not (bearing in mind it's presently 2300 hours).

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -110,12 +118,62 @@ protected Object readResolve() throws ObjectStreamException {
...
+                } else {
+                    // Need workflow-cps 2.33+ to use the optimization.
+                    LOGGER.log(Level.FINE, "falling back to old isActive implementation for {0}", this);
+                    if (isRunning()) {
+                        return true;
+                    }
+                    List<FlowNode> headNodes = exec.getCurrentHeads();
+                    AbstractFlowScanner scanner = (headNodes.size() > 1) ? new DepthFirstScanner() : new LinearBlockHoppingScanner();
+                    return scanner.findFirstMatch(headNodes, Predicates.equalTo(this)) != null;
+                }
```
svanoort: Future enhancement if we do something like this: if we gotta scan the whole flow, fully populate the cache of the block structure along the way, so we never have to do that full scan again.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
jglick: Right, this case needs to be tested.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
jglick: The work done inside the monitor is so trivial that contention is not likely to be an issue: HotSpot should just switch to a spin lock.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -109,14 +109,39 @@ protected Object readResolve() throws ObjectStreamException {
...
+        if (this instanceof FlowEndNode) { // cf. JENKINS-26139
+            return false;
+        }
+        List<FlowNode> currentHeads = exec.getCurrentHeads();
+        if (currentHeads.contains(this)) {
+            return true;
+        }
+        if (this instanceof BlockStartNode) {
+            for (FlowNode headNode : currentHeads) {
+                if (new LinearBlockHoppingScanner().findFirstMatch(headNode, Predicates.equalTo(this)) != null) {
```
jglick: Not as fast as the `GraphListener`-based approach I tried first, but still far faster than the original `LogActionImpl.isRunning`, and more correct in the face of `parallel` as well. With the core and `workflow-durable-task-step` patches plus this, the test case at n=1000 completes in ~9m, not as impressive as the ~4m with the cache of `BlockStartNode`…but compare to the 33m baseline. As noted in [JENKINS-45553](https://issues.jenkins-ci.org/browse/JENKINS-45553), most of the calls to `isActive` would go away with fixes of JENKINS-38381 & JENKINS-36547, reducing pressure to optimize this further.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -59,12 +60,11 @@
 @ExportedBean
 public abstract class FlowNode extends Actionable implements Saveable {
     private transient List<FlowNode> parents;
-    private List<String> parentIds;
+    private final List<String> parentIds;
```
jglick: just some warnings

pom.xml
```diff
@@ -62,9 +62,9 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>1.642.3</jenkins.version>
+        <jenkins.version>2.7.3</jenkins.version>
```
jglick: test against `workflow-cps` 2.33

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -212,8 +239,9 @@ public String getDisplayFunctionName() {
     @Exported
     public BallColor getIconColor() {
         BallColor c = getError()!=null ? BallColor.RED : BallColor.BLUE;
-        // TODO this should probably also be _anime in case this is a step node with a body and the body is still running (try FlowGraphTable for example)
-        if (isRunning())        c = c.anime();
+        if (isActive()) {
```
jglick: Visible in flow graph table. You can now see at a glance which blocks are still running.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
svanoort: Would be nice to get more detail on how you're doing these tests for performance?  The improvements sound good, but would like to know what specifically is being measured.

svanoort: @jglick 🐛  If you use this implementation you're potentially scanning all the nodes before a parallel multiple times, remember?   So if you have this: 

```
stage ('some long stuff') {
  for(int i=0; i<1000; i++) {
     echo("say something for the $i-th time")
  }
}
parallel withNBranches
```
then you're scanning 'some long stuff' up to N times for N*1000+4 nodes. 

In *this* specific testcase you're seeing faster build times but you're optimizing just for that case.  

Better solution: go back to the cache-like method, but make the cache persisted within the FlowExecution or something like that (and rebuild it if not present).  

Long term we need to support this within either the `FlowNodeStorage` or provide a lookup object with graph structure information that we can use to accelerate lookups (AKA Bismuth Level 3, the tree-structured view of blocks).

svanoort: Ref: https://github.com/jenkinsci/workflow-support-plugin/pull/19#discussion_r84333212 for the same class of problem cropping up again. 

If we don't provide a comprehensive lookup for block info, the next best thing we can do is provide something that scans along nodes within a branch before going beyond the ParallelStartNode.  Basically like a  `ForkScanner`  mostly but without so strictly enforcing iteration order.   Might be easier now that we have trivial ways to check if a node is a ParallelStartNode.

jglick:
Possible optimization: implement `GraphListener` *and* `FlowExecutionListener`; per `FlowExecutionOwner` (no need for weak keys but can remove entries once completed), keep a `Map<String, Boolean>` of start node IDs to see if they are closed. If `workflow-cps` is too old, the IDs will never be registered, so we fall back to the algorithmic version. If `workflow-job` is too old, the flow will never be registered, so ditto. If we restart, we use the optimized version for start nodes registered after the restart, and the algorithmic version for earlier ones. Should be reasonably simple and safe while speeding up most cases.

svanoort:
Yes @jglick that's exactly along the lines I was thinking -- it avoids one case of abusive performance with a long run of nodes followed by a dense parallel

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
svanoort: @jglick You need to ensure that at least the parallel is not generating abusive performance if there are a lot of steps preceding the parallel - this reverts the previous fix that avoided that.  If that's solved, this is ready to go forward.

pom.xml
```diff
@@ -86,13 +86,13 @@
...
             <artifactId>workflow-job</artifactId>
-            <version>2.9</version>
+            <version>2.11.1</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>2.27</version>
+            <version>2.33</version>
```
jglick: Also tried `FlowNodeTest` with `workflow-cps` 2.32 (no `GraphListener`); `workflow-job` 2.9 (no `FlowExecutionListener`); and both together. Passed in all cases, albeit without optimization.

jglick:
New version brought test project time down to under 4m, the best yet.

jglick:
> one case of abusive performance with a long run of nodes followed by a dense parallel

As noted in https://github.com/jenkinsci/workflow-api-plugin/pull/45/commits/f58028addbc477d5c81ee2e4d543e9c87c01827e, the original implementation was _not correct_ so I am not going back to it. Theoretically there could be degraded performance in a weird case like

```groovy
[0..9999].each {
  echo 'what a dumb idea'
}
def branches = [:]
[0..9999].each {x ->
  branches["b$x"] = {sleep 999}
}
parallel branches
```

and then restarted during the sleeps, since checking liveness of each branch `BlockStartNode` will require doing a blockhopping scan from each 10k `sleep` head nodes past each 10k `echo` prior nodes. Whether this is worth optimizing again for, I do not know—seems an unlikely case. Probably `onResumed` could be improved to process all prior block start/end nodes using a `DepthFirstScanner`, which would fix this case (so long as `workflow-cps` and `workflow-job` are sufficiently new).

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
```diff
@@ -0,0 +1,124 @@
...
+                SemaphoreStep.waitForStart("last/1", b);
+                assertActiveSteps(b, "Start of Pipeline", "semaphore: last");
+                SemaphoreStep.success("last/1", null);
+                rr.j.waitForCompletion(b);
+                assertActiveSteps(b);
+            }
+        });
+    }
+    private static void assertActiveSteps(WorkflowRun b, String... expected) {
+        List<String> actual = new ArrayList<>();
```
svanoort: 🐜  Did you mean to omit the second type?  I only ask because you made a point of adding it for the other ones.

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
```diff
@@ -0,0 +1,124 @@
...
+        for (FlowNode n : new FlowGraphWalker(b.getExecution())) {
```
svanoort: 🐜  DepthFirstScanner, not FlowGraphWalker ;)

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
```diff
@@ -0,0 +1,124 @@
...
+import org.jvnet.hudson.test.LoggerRule;
+import org.jvnet.hudson.test.RestartableJenkinsRule;
+
+public class FlowNodeTest {
+
+    @Rule public RestartableJenkinsRule rr = new RestartableJenkinsRule();
+    @Rule public LoggerRule logging = new LoggerRule().record(FlowNode.class, Level.FINER);
+
+    @Issue("JENKINS-38223")
+    @Test public void isActive() {
```
svanoort: This is a really well-written testcase.  👍 👍

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -109,14 +117,124 @@ protected Object readResolve() throws ObjectStreamException {
...
+    @Restricted(DoNotUse.class)
+    @Extension public static final class FlowL extends FlowExecutionListener {
+        final Map<FlowExecutionOwner, Map<String, Boolean>> startNodesAreClosedByFlow = new HashMap<>();
+        static Map<FlowExecutionOwner, Map<String, Boolean>> startNodesAreClosedByFlow() {
+            FlowL flowL = ExtensionList.lookup(FlowExecutionListener.class).get(FlowL.class);
+            return flowL != null ? flowL.startNodesAreClosedByFlow : /* ? */ new HashMap<FlowExecutionOwner, Map<String, Boolean>>();
+        }
+        @Override public void onRunning(FlowExecution execution) {
+            LOGGER.finer("FlowExecutionListener working");
+            assert !startNodesAreClosedByFlow.containsKey(execution.getOwner());
```
svanoort: Mixed feelings about the assert here -- would rather we log something rather than accept an AssertionError.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -109,14 +117,124 @@ protected Object readResolve() throws ObjectStreamException {
...
+                    startNodesAreClosed.put(((BlockEndNode) n).getStartNode().getId(), true);
+                } else if (n instanceof BlockStartNode) {
+                    if (!startNodesAreClosed.containsKey(n.getId())) {
+                        startNodesAreClosed.put(n.getId(), false);
+                    }
+                }
+            }
+        }
+        @Override public void onCompleted(FlowExecution execution) {
+            assert startNodesAreClosedByFlow.containsKey(execution.getOwner());
```
svanoort: Ditto about the asserts.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -109,14 +117,124 @@ protected Object readResolve() throws ObjectStreamException {
...
+                }
+            }
+        }
+    }
+    @Restricted(DoNotUse.class)
+    @Extension public static final class FlowL extends FlowExecutionListener {
+        final Map<FlowExecutionOwner, Map<String, Boolean>> startNodesAreClosedByFlow = new HashMap<>();
+        static Map<FlowExecutionOwner, Map<String, Boolean>> startNodesAreClosedByFlow() {
+            FlowL flowL = ExtensionList.lookup(FlowExecutionListener.class).get(FlowL.class);
+            return flowL != null ? flowL.startNodesAreClosedByFlow : /* ? */ new HashMap<FlowExecutionOwner, Map<String, Boolean>>();
```
svanoort: A bit confused about the intent here?

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
svanoort: Merits an explanatory comment I think, to the effect of "Lookup map is empty, regenerate it by iterating through all nodes and marking blocks that where they're terminated."  Or whatever you think, but the logic checks out nicely for me -- just need to make sure people can follow it later.

svanoort: Maybe mention we're marking all nodes closed when we see the end, but for start nodes only toggle them if not present - this is necessary b/c of the order of DepthFirstScanner iteration. 

ForkScanner is guaranteed to see all nodes in last -> first order, but the overhead of tracking state for branches means DepthFirstScanner is the way to go here.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
svanoort: 🐛 I think this *may* generate incorrect results if we somehow call this method prematurely -- the map can be written to by the GraphListener I think and that write will be lost. 

Would be very hard to test I think though --  also I think the ExtensionList lookup *should* ensure the extension is loaded. 

Would prefer to perhaps throw an exception if extension not loaded.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -109,14 +117,124 @@ protected Object readResolve() throws ObjectStreamException {
...
+     * For old {@code workflow-job}, the top-level entries will be missing;
+     * for old {@code workflow-cps}, the second-level entries will be missing.
+     */
+    @Restricted(DoNotUse.class)
+    @Extension public static final class GraphL implements GraphListener.Synchronous {
+        @Override public void onNewHead(FlowNode node) {
+            LOGGER.finer("GraphListener working");
+            if (node instanceof BlockStartNode || node instanceof BlockEndNode) {
+                Map<String, Boolean> startNodesAreClosed = FlowL.startNodesAreClosedByFlow().get(node.getExecution().getOwner());
+                if (startNodesAreClosed != null) {
```
svanoort: Mention this is safe to do because the FlowExecutionListener has ensured that the FlowExecution is only null if flow is done (otherwise always set for all running builds).

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -109,14 +117,124 @@ protected Object readResolve() throws ObjectStreamException {
...
+                    if (node instanceof BlockStartNode) {
+                        assert !startNodesAreClosed.containsKey(node.getId());
```
svanoort: Again would prefer not to risk AssertionErrors here -- instead log something.  Best not to risk breaking builds in flight if something wacky happens with the maps.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -109,14 +117,124 @@ protected Object readResolve() throws ObjectStreamException {
...
         return getExecution().isCurrentHead(this);
     }
 
     /**
+     * Checks whether a node is still part of the active part of the graph.
+     * Unlike {@link #isRunning}, this behaves intuitively for a {@link BlockStartNode}:
+     * it will be considered active until the {@link BlockEndNode} is added.
+     */
+    @Exported(name="running")
+    public final boolean isActive() {
```
svanoort: Because this method is an extremely hot path and performance-critical for pipeline, I would **strongly** prefer to remove all the logging and order conditionals to put the common case first.   If you want logging, have two versions of this method, one of which includes logging and one which doesn't, and have a static boolean used to decide which version to invoke (toggleable via scriptconsole).

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -109,14 +117,124 @@ protected Object readResolve() throws ObjectStreamException {
...
+            Map<FlowExecutionOwner, Map<String, Boolean>> startNodesAreClosedByFlow = FlowL.startNodesAreClosedByFlow();
+            LOGGER.log(Level.FINER, "for {0}, startNodesAreClosedByFlow={1}", new Object[] {this, startNodesAreClosedByFlow});
+            Map<String, Boolean> startNodesAreClosed = startNodesAreClosedByFlow.get(exec.getOwner());
+            if (startNodesAreClosed != null) {
+                Boolean closed = startNodesAreClosed.get(id);
+                if (closed != null) { // quick version
+                    LOGGER.log(Level.FINER, "quick closed={0}", closed);
+                    return !closed;
+                } else {
+                    LOGGER.log(Level.FINER, "no record of {0} in {1}, presumably GraphListener not working", new Object[] {this, exec});
```
svanoort: Worth mentioning that this will detect cases where GraphListeners as extensions can't be invoked for all FlowNodes.

```java
Method m = hudson.util.ReflectionUtils.findMethod(CpsFlowExecution.class, "getListenersToRun", null);
            if (m != null) {
                return true;
            }
            return true;
```

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
svanoort: Specifically:  I think the most common case is a node either being in the current heads (active=true) or NOT being a block start/end node AND not being in the current heads (active=false).  I think putting that conditional first may make the hottest path perform well.

svanoort: (I'm aware there's all sort of speculative branch prediction and whatnot that happens at the hardware and maybe some re-ordering in the JIT compilation where allowed, but can't hurt). 

Happy to take a crack at that ordering myself in an add-on PR if you like.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
svanoort: (Not suggesting you use this code, just that I had a case where it was desirable to have it myself)

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
jglick: > remove all the logging

If you look more carefully you will see that all of the logging calls use a message format with arguments that require no method calls, so the cost of the statement comes down to
* create a small array
* `log` checks that logger is enabled; it is not, so returns

This is cheap enough to leave in.

> put the common case first

Read more carefully. I am putting the *cheap* checks first. Calling `CpsFlowExecution.getCurrentHeads` requires constructions of a list and a map traversal—pretty cheap, but more work than the cases being handled before it.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
jglick: Not sure what you are getting at here. Either `workflow-cps` is 2.33+ and it will be called on all `FlowNode`s; or it is 2.32- and it will never be called.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
jglick: Not sure what you are saying. That what is safe to do?

It will be null if we have an old `workflow-job`, FWIW.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
jglick: Assertions will normally be disabled in production systems. These are here to make sure mistakes will be caught in tests.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
jglick: If `flowL == null`? Should never happen unless Jenkins is basically trashed anyway (in which case we would see all sorts of errors earlier). But FindBugs forced me to put in the check.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
jglick: Again, assertions are for tests, not production.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
jglick: > we're marking all nodes closed when we see the end, but for start nodes only toggle them if not present

I guess you are referring to `onResumed` here, not `onCompleted` where the comment was placed?

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
jglick: The “second type”? This is `-source 7`, so I am using diamond inference.

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
jglick: Either works fine, and this is test code so it does not really matter.

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
jglick: Though it would be useful to have more test coverage going forward.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
svanoort: Yes, I'm not entirely sure why the comment landed here when I clicked there, but perhaps I fatfingered something.

jglick:
@reviewbybees done

PR #46, opened 2017-07-27
Unrestricting getPersistentAction

jglick:
See downstream, coming shortly.

@reviewbybees


svanoort:
Heck, if you didn't do it, I was going to :)

Would it make sense to offer this in the core as an API?  That's the other thing I was thinking of and would be happy to add.

jglick:
> Would it make sense to offer this in the core as an API?

Perhaps. Would need some use cases.

PR #48, opened 2017-08-25
[JENKINS-27395] Start recording enclosing blocks for flow nodes

abayer:
[JENKINS-27395](https://issues.jenkins-ci.org/browse/JENKINS-27395)

Specifically for AtomNodes, BlockStartNodes, and BlockEndNodes.

Will be upstream of https://github.com/jenkinsci/junit-plugin/pull/76 shortly.

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
@@ -38,6 +38,7 @@ import java.io.IOException;
 import java.io.ObjectStreamException;
 import java.util.AbstractList;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
@@ -47,6 +48,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+
 import org.jenkinsci.plugins.workflow.actions.ErrorAction;
 import org.jenkinsci.plugins.workflow.actions.LabelAction;
 import org.jenkinsci.plugins.workflow.actions.PersistentAction;
@@ -69,6 +71,9 @@ import org.kohsuke.stapler.export.ExportedBean;
 public abstract class FlowNode extends Actionable implements Saveable {
     private transient List<FlowNode> parents;
     private final List<String> parentIds;
+    private transient List<FlowNode> enclosingBlocks;
+    private final List<String> allEnclosingIds = new ArrayList<>();
+    protected String enclosingId;
 
     private String id;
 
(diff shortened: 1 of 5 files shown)


src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
```diff
@@ -121,4 +124,251 @@ private static void assertActiveSteps(WorkflowRun b, String... expected) {
         assertEquals(Arrays.asList(expected), actual);
     }
 
+    @Issue("JENKINS-27395")
+    @Test
+    public void enclosingBlocks() throws Exception {
```
svanoort: I suspect this is giving good coverage, but it really needs to be split up into several smaller testcases to let us debug any issues introduced in a specific case (for example: incomplete parallels).

abayer: awwwww. Can you give me some examples of dummy Jenkinsfiles that you want covered?

svanoort: Hold on, let me review other parts and then will explain more

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -69,6 +71,9 @@
 public abstract class FlowNode extends Actionable implements Saveable {
     private transient List<FlowNode> parents;
     private final List<String> parentIds;
+    private transient List<FlowNode> enclosingBlocks;
+    private final List<String> allEnclosingIds = new ArrayList<>();
```
svanoort: 🐛  We do not want to attach this to individual FlowNodes, and *definitely* don't want to add additional persistent content to them.  This is because I intend this information to be retained and exposed via a higher-level construct that provides fast search, probably tied to the FlowNodeStorage.  Per what I said earlier: 

> One note: we may want to ensure the actual implementation itself is never exposed in case (cough) this is supported directly via the FlowNodeStorage implementations in the future.

Once we have this tied into the storage (or a StorageIndex class) then having the extra persistent data on the nodes will be just cruft.

abayer: Ok, I was confused by your earlier comments about persisting all the enclosing blocks here.

jglick: Yeah we do not want a new non`transient` field here. (Especially one which is not restored by `readResolve` or some `onLoad` or whatever.)

abayer: Yeah, waiting to see @svanoort's variant on this theme. =)

svanoort: It's coming.  I was working on it on the flight but couldn't get into the zone very well

svanoort:
@jglick I think you'll like my variant on the theme then. ;-)

abayer:
Closing in favor of #50

PR #49, opened 2017-09-01
Provide FlowNode APIs to let us be more selective about when we persist data

svanoort:
Upstream of major performance optimization to FlowNode storage that significantly decreases persistence calls.

This lets us:
- [x] Signal if a FlowNode shouldn't be directly persisted with every new action added
- [x] Attach actions to a FlowNode explicitly without triggering disk persistence
- [x] Trigger explicit serialization of FlowNodes

More detail TBD when I finish filling in all the administrivia and file appropriate JIRAs.

- Although doesn't directly consume API from https://github.com/jenkinsci/workflow-step-api-plugin/pull/29 we need to ensure that both APIs are present in all plugins consuming this version, so we add a dependency

Snapshot deployed:

```
<dependency>
            <!-- Temp version for https://github.com/jenkinsci/workflow-api-plugin/pull/49 -->
            <version>2.21-deferactionwrite-20170901.212705-1</version>
        </dependency>
```

Note:

- Test coverage is via the FlowNodeStorage impls in Workflow-support plugin (downstream)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/TimingAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/TimingAction.java
@@ -31,7 +31,7 @@ import org.jenkinsci.plugins.workflow.graph.FlowNode;
  *
  * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
  */
-public class TimingAction implements Action {
+public class TimingAction implements PersistentAction {
 
     // TODO perhaps add a FlowNodeViewColumn rendering <i:formatDate value="…" type="both" dateStyle="medium" timeStyle="medium"/>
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/actions/TimingAction.java
```diff
@@ -31,7 +31,7 @@
  *
  * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
  */
-public class TimingAction implements Action {
+public class TimingAction implements PersistentAction {
```
svanoort: Just tying up a loose end while I see it.

jglick:
Still valid or superseded by #55?

svanoort:
Supersede by #55 indeed.

PR #50, opened 2017-09-13
[JENKINS-37573] / [JENKINS-45553] Provide a fast view of block structures in the flow graph

svanoort:
Consider this the extension of the graph analysis APIs, by providing some structural information about blocks in the Flow and their structure.  A precursor to providing more detailed structural information in an easy view. 

Replaces:  #48 and reimplements #45

Limited implementation of [JENKINS-37573](https://issues.jenkins-ci.org/browse/JENKINS-37573), kinda.   Ties to [JENKINS-45553](https://issues.jenkins-ci.org/browse/JENKINS-45553).

* SNAPSHOT deployed as version: 2.21-20170918.154256-6

TODO:

- [x] Benchmark in the pipeline scalability lab vs. existing APIs
- [x] Um, some consistency and naming stuffs.

#Benchmarks:
128 executors running on a Swarm Agent.  Requires this PR + upstream workflow-cps PR + the workflow-support PR (current build of master).  Note that we've done a warmup with the Scalability Lab until runtimes & node creation rates with 200 branches are more or less constant.

Script:
```groovy
properties([buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '20')), pipelineTriggers([])])

int numBranches = 200; // or 400 or 1000

def stepsForParallel = [:]
for (int i = 0; i < numBranches; i++) {
  def s = "subjob_${i}" 
  stepsForParallel[s] = {
    node {
      echo "hello"
    }
  }
}
parallel stepsForParallel
```

* With 100 branches, runtime: 2.7s vs. 5s pre-patch
* With 200 branches, runtime: 7s vs.  24s pre-patch
* With 400 branches, runtime: ~33s  vs.  2m 27s (147s) pre-patch
* With 800 branches, runtime: 3 min, 5 sec (185s).
* With 1000 branches, runtime: 6 min, 23s (383s)

So, not *linear* scaling, but far more acceptable -- 30s to run 400 branches seems pretty A-OK to me.
If you need more than 400 branches it is probably time to reconsider what you're doing. ;)

Questions:
* Is there a use-case for returning all FlowNodes in a block?  If so, when?  I removed a previous variant but can restore it easily.
* Would it be better to try to force working with just the FlowNode IDs?

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
@@ -50,6 +50,8 @@ import jenkins.model.Jenkins;
 import jenkins.model.queue.AsynchronousExecution;
 import org.acegisecurity.Authentication;
 import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 
 /**
  * State of a currently executing workflow.
@@ -239,6 +241,7 @@ public abstract class FlowExecution implements FlowActionStorage, GraphLookupVie
     /** Tests if the node (in this execution) is a currently running head, or the start of a block that has not completed executing
      * @throws IllegalArgumentException If the input {@link FlowNode} does not belong to this execution
      */
+    @Restricted(NoExternalUse.class)  // Only public because graph, flow, and graphanalysis are separate packages
     public boolean isActive(@Nonnull  FlowNode node) {
         if (!this.equals(node.getExecution())) {
             throw new IllegalArgumentException("Can't look up info for a FlowNode that doesn't belong to this execution!");
@@ -251,6 +254,7 @@ public abstract class FlowExecution implements FlowActionStorage, GraphLookupVie
      *  @throws IllegalArgumentException If the input {@link FlowNode} does not belong to this execution
      */
     @CheckForNull
+    @Restricted(NoExternalUse.class)  // Only public because graph, flow, and graphanalysis are separate packages
     public BlockEndNode getEndNode(@Nonnull BlockStartNode startNode) {
         if (!this.equals(startNode.getExecution())) {
             throw new IllegalArgumentException("Can't look up info for a FlowNode that doesn't belong to this execution!");
@@ -265,6 +269,7 @@ public abstract class FlowExecution implements FlowActionStorage, GraphLookupVie
      * @throws IllegalArgumentException If the input {@link FlowNode} does not belong to this execution
      */
     @CheckForNull
+    @Restricted(NoExternalUse.class)  // Only public because graph, flow, and graphanalysis are separate packages
     public BlockStartNode findEnclosingBlockStart(@Nonnull FlowNode node) {
         if (!this.equals(node.getExecution())) {
             throw new IllegalArgumentException("Can't look up info for a FlowNode that doesn't belong to this execution!");
@@ -278,6 +283,7 @@ public abstract class FlowExecution implements FlowActionStorage, GraphLookupVie
      * @throws IllegalArgumentException If the input {@link FlowNode} does not belong to this execution
      */
     @Nonnull
+    @Restricted(NoExternalUse.class)  // Only public because graph, flow, and graphanalysis are separate packages
     public List<BlockStartNode> findAllEnclosingBlockStarts(@Nonnull FlowNode node) {
         if (!this.equals(node.getExecution())) {
             throw new IllegalArgumentException("Can't look up info for a FlowNode that doesn't belong to this execution!");
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
@@ -291,4 +291,12 @@ public abstract class FlowExecution implements FlowActionStorage, GraphLookupVie
         return getInternalGraphLookup().findAllEnclosingBlockStarts(node);
     }
 
+    @Nonnull
+    @Restricted(NoExternalUse.class)
+    public Iterable<BlockStartNode> iterateEnclosingBlocks(@Nonnull FlowNode node) {
+        if (!this.equals(node.getExecution())) {
+            throw new IllegalArgumentException("Can't look up info for a FlowNode that doesn't belong to this execution!");
+        }
+        return getInternalGraphLookup().iterateEnclosingBlocks(node);
+    }
 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
@@ -70,7 +70,7 @@ import org.kohsuke.accmod.restrictions.NoExternalUse;
  * @author Kohsuke Kawaguchi
  * @author Jesse Glick
  */
-public abstract class FlowExecution implements FlowActionStorage, GraphLookupView {
+public abstract class FlowExecution implements FlowActionStorage, GraphLookupView {  // Implements GraphLookupView because FlowNode lives in another package
 
     protected transient GraphLookupView internalGraphLookup = null;
 
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
@@ -34,7 +34,6 @@ import org.jenkinsci.plugins.workflow.graph.BlockStartNode;
 import org.jenkinsci.plugins.workflow.graph.FlowActionStorage;
 import org.jenkinsci.plugins.workflow.graph.FlowEndNode;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
-import org.jenkinsci.plugins.workflow.graph.FlowStartNode;
 import org.jenkinsci.plugins.workflow.graph.GraphLookupView;
 import org.jenkinsci.plugins.workflow.graph.StandardGraphLookupView;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
@@ -29,9 +29,14 @@ import hudson.Util;
 import hudson.model.Executor;
 import jenkins.model.CauseOfInterruption;
 import org.jenkinsci.plugins.workflow.actions.ErrorAction;
+import org.jenkinsci.plugins.workflow.graph.BlockEndNode;
+import org.jenkinsci.plugins.workflow.graph.BlockStartNode;
 import org.jenkinsci.plugins.workflow.graph.FlowActionStorage;
 import org.jenkinsci.plugins.workflow.graph.FlowEndNode;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
+import org.jenkinsci.plugins.workflow.graph.FlowStartNode;
+import org.jenkinsci.plugins.workflow.graph.GraphLookupView;
+import org.jenkinsci.plugins.workflow.graph.StandardGraphLookupView;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
 import hudson.model.Result;
 import hudson.security.ACL;
@@ -63,7 +68,17 @@ import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
  * @author Kohsuke Kawaguchi
  * @author Jesse Glick
  */
-public abstract class FlowExecution implements FlowActionStorage {
+public abstract class FlowExecution implements FlowActionStorage, GraphLookupView {
+
+    protected transient GraphLookupView internalGraphLookup = null;
+
+    /** This may be overridden if the FlowExecution has a better source of structural information, such as the {@link FlowNode} storage.*/
+    protected synchronized GraphLookupView getInternalGraphLookup() {
+        if (internalGraphLookup == null) {
+            internalGraphLookup = new StandardGraphLookupView(this);
+        }
+        return internalGraphLookup;
+    }
 
     /**
      * Called after {@link FlowDefinition#create(FlowExecutionOwner, List)} to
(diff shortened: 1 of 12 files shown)


svanoort:
@abayer You mind taking an initial peek?  I did a lot of this in one large rush and it probably needs a second set of eyes to sanity-check.  

Note that I've decided not to cache the entire list of enclosing blocks -- HashMap lookups to traverse that list probably aren't too expensive.  We can always throw in some caching if those calls become frequent and expensive -- I suspect the worst part will probably be fetching from the FlowNodeStorage cache.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -267,15 +178,58 @@ public final boolean isActive() {
...
+        FlowNode enclosing = this.exec.findEnclosingBlockStart(this);
+        return enclosing != null ? enclosing.getId() : null;
+    }
+
+    /**
+     * Get the list of enclosing blocks, starting from innermost, for this node. May be empty. Loads the flow nodes into
+     * the transient {@code enclosingBlocks} field if needed.
+     */
+    @Nonnull
+    public List<FlowNode> getEnclosingBlocks() {
```
abayer: *faints with happiness*

svanoort: Might want to go to a doctor if that makes you faint ;)

abayer: What can I say, I'm fond of easy APIs. =)

svanoort:
@reviewbybees

jglick:
W.r.t. to the testing script, I would suggest

* define parallelism as a build parameter with a default
* run an actual `sh` step on the `node`, as this will exercise some code paths likely to be used in realistic scenarios which you are currently omitting

BTW your PR summary implies that this is upstream or downstream of a couple other PRs but does not link to them—would be useful background.

svanoort:
@jglick Thanks - WDYT about the code?

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
```diff
@@ -0,0 +1,237 @@
...
+                        nearestEnclosingBlock.put(newHead.getId(), enclosingId);
+                    }
+                }
+            }
+        }
+    }
+
+    /** Create a lookup view for an execution */
+    public StandardGraphLookupView(FlowExecution execution) {
+        execution.addListener(this);
```
jglick: It is an antipattern to attach an instance of some class to some external source (such as adding a listener) in its constructor when the class is not `final`—you can actually expose a partially-constructed object. Better to have the caller of the constructor do this.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -133,112 +145,11 @@ public final boolean isRunning() {
...
-                    }
-                }
-            }
-        }
-        @Override public void onCompleted(FlowExecution execution) {
-            assert startNodesAreClosedByFlow.containsKey(execution.getOwner());
-            // After a build finishes, we do not need the cache any more, since we do the equivalent of FlowExecution.isComplete relatively quickly:
-            startNodesAreClosedByFlow.remove(execution.getOwner());
+        } if (this instanceof BlockEndNode || this instanceof AtomNode) {
+            return this.exec.isCurrentHead(this);
```
jglick: I think you could just always delegate to `FlowExecution.isActive`, right? Seems like you are duplicating the special cases here.

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
```diff
@@ -0,0 +1,237 @@
...
+    BlockStartNode bruteForceScanForEnclosingBlock(@Nonnull final FlowNode node) {
+        FlowNode current = node;
+
+        while (!(current instanceof FlowStartNode)) {  // Hunt back for enclosing blocks, a potentially expensive operation
+            if (current instanceof BlockEndNode) {
+                // Hop over the block to the start
+                BlockStartNode start = ((BlockEndNode) current).getStartNode();
+                blockStartToEnd.put(start.getId(), current.getId());
+                current = start;
+                continue;  // Allows us to use this to handle special cases
```
jglick: (I am not even trying to review this kind of code. Hope your test coverage is good.)

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
```diff
@@ -121,4 +127,303 @@ private static void assertActiveSteps(WorkflowRun b, String... expected) {
         assertEquals(Arrays.asList(expected), actual);
```
jglick: You were able to leave my existing test untouched—that is encouraging. :-)

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
```diff
@@ -121,4 +127,303 @@ private static void assertActiveSteps(WorkflowRun b, String... expected) {
...
+                assertExpectedEnclosing(execution, "2", null);
+
+                // CatchError block, outer block
+                assertExpectedEnclosing(execution, "3", "2");
+
+                // CatchError block, inner block (body)
+                assertExpectedEnclosing(execution, "4", "3");
+
+                // Echo step, enclosed in body block for catchError step
+                assertExpectedEnclosing(execution, "5", "4");
```
jglick: As mentioned in a long-ago PR, this is going to be pretty fragile; worth spending the time improving the test harness to let you specify step names etc., something like I did with `assertActiveSteps` above. Even with the benefit of a commented-out legend the tests here are close to unreadable.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
```diff
@@ -220,4 +237,66 @@ public boolean blocksRestart() {
...
+    public BlockStartNode findEnclosingBlockStart(@Nonnull FlowNode node) {
+        if (!this.equals(node.getExecution())) {
+            throw new IllegalArgumentException("Can't look up info for a FlowNode that doesn't belong to this execution!");
+        }
+        return getInternalGraphLookup().findEnclosingBlockStart(node);
+    }
+
+    /** Return all enclosing block start nodes, as with {@link #findEnclosingBlockStart(FlowNode)}.
+     *  @param node Node to find enclosing blocks for, which must belong to this FlowExecution
+     *  @return All enclosing block starts in no particular sort order, or EMPTY_LIST if this is a start or end node
```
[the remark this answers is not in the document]
svanoort: I like it, good point.

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
svanoort: @jglick Unfortunately that's not terribly helpful since this is the meat of the PR and you're the only other person familiar enough with this to really review it.   So, could I please ask you to give this a review -- seems only fair after I reviewed #45

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
svanoort: @jglick As they say around here "if it ain't broke don't fix it."

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -267,15 +178,65 @@ public final boolean isActive() {
 
     @Nonnull
     private List<FlowNode> loadParents(List<String> parentIds) {
-        List<FlowNode> _parents = new ArrayList<>(parentIds.size());
-        for (String parentId : parentIds) {
+        List<FlowNode> _parents;
```
abayer: Why not move declaring `_parents` into the `else` block, since it's not referenced elsewhere?

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -267,15 +178,65 @@ public final boolean isActive() {
...
+                } catch (IOException x) {
+                    LOGGER.log(Level.WARNING, "failed to load parents of " + id, x);
+                }
+            }
+            return _parents;
+        }
+    }
+
+    /**
+     * Get the first enclosing block ID for this node. Can be null.
```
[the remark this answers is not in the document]
abayer: That's copy-pasta from my code, so I'm to blame for the crap wording. =)

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -267,15 +178,65 @@ public final boolean isActive() {
...
+     * @return
+     */
+    @CheckForNull
+    public String getEnclosingId() {
+        FlowNode enclosing = this.exec.findEnclosingBlockStart(this);
+        return enclosing != null ? enclosing.getId() : null;
+    }
+
+    /**
+     * Get the list of enclosing blocks, starting from innermost, for this node. May be empty. Loads the flow nodes into
```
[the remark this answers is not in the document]
abayer: Also my crap wording.

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
```diff
@@ -0,0 +1,237 @@
...
+    static final String INCOMPLETE = "";
+
+    /** Map the blockStartNode to its endNode, to accellerate a range of operations */
+    HashMap<String, String> blockStartToEnd = new HashMap<String, String>();
+
+    /** Map a node to its nearest enclosing block */
+    HashMap<String, String> nearestEnclosingBlock = new HashMap<String, String>();
+
+    /** Update with a new node added to the flowgraph */
+    public void onNewHead(@Nonnull FlowNode newHead) {
```
abayer: Nitpicky, but `@Override`?

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
abayer: Do we have any examples of such special cases?

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/NodeStepNamePredicate.java
```diff
@@ -41,7 +41,7 @@
 import java.util.logging.Logger;
 
 /** Predicate that matches {@link FlowNode}s (specifically {@link StepNode}s) with a specific {@link StepDescriptor} name.
- *  May be used in preference to {@link NodeStepTypePredicate} in cases whern dependencie structures prevent import
+ *  May be used in preference to {@link NodeStepTypePredicate} in cases whern dependency structures prevent import
```
abayer: ...I'll admit it, I'm entertained that you fixed `dependencie` but not `whern`. =)

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
abayer: Yeah, I've got some tooling along those lines somewhere - of course now I can't remember where. I'll find it and write a follow-up PR for a harness.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/NodeStepNamePredicate.java
svanoort: I kind of like picturing the javadocs written by a crusty old gold-miner

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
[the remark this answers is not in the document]
abayer: Details, details. =) The main point is - it's my fault!

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
svanoort: Fragile how?  At this point we have so many things tied to the current ID format (including community extensions) that we can't realistically change it.   I actually think the "assertActiveSteps" is quite a bit more fragile because it's reliant on user display strings being immutable and we do have legitimate cases where that might change.

On the other hand, if we want to provide an XPath-like query language for the flowgraph that's something that might have usefulness beyond just testing (though it would require some fairly extensive testcases on its own, which of necessity would have to use something like this).

jglick:
```
java.lang.AssertionError: expected:<[]> but was:<[End of Pipeline]>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.jenkinsci.plugins.workflow.graph.FlowNodeTest.assertActiveSteps(FlowNodeTest.java:127)
	at org.jenkinsci.plugins.workflow.graph.FlowNodeTest.access$000(FlowNodeTest.java:51)
	at org.jenkinsci.plugins.workflow.graph.FlowNodeTest$3.evaluate(FlowNodeTest.java:113)
```

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/NodeStepNamePredicate.java
svanoort: ... or for some reason "dependencie" grated enough to be noticeable but "whern" didn't

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
```diff
@@ -0,0 +1,219 @@
...
+        DepthFirstScanner scan = new DepthFirstScanner();
+        scan.setup(start.getExecution().getCurrentHeads());
+        for (FlowNode f : scan) {
+            if (f instanceof BlockEndNode) {
+                BlockEndNode end = (BlockEndNode)f;
+                BlockStartNode maybeStart = end.getStartNode();
+                // Cache start in case we need to scan again in the future
+                blockStartToEnd.put(maybeStart.getId(), end.getId());
+                if (start.equals(maybeStart)) {
+                    return end;
```
svanoort: Needs to be covered by a testcase (negative for coverage).

svanoort: Specifically: 
```
if (start.equals(maybeStart)) {
                    return end;
}
```

Other than that, line coverage seems okay (not 100% for StandardGraphLookupView but the cases uncovered are things like rethrowing exceptions and).

svanoort:
Kicking back to WIP while I figure out whether or not a test failure from extended coverage is real or the test being overly finnicky.

svanoort:
Okay that's better.  Testcase whining = not a real problem, and better coverage.

PR #53, opened 2017-09-22
[Jenkins-27395] block structure lookup Mk2 

svanoort:
Replaces https://github.com/jenkinsci/workflow-api-plugin/pull/50

Diff vs the last reviewed version: https://github.com/jenkinsci/workflow-api-plugin/compare/65759d3f8aa61192913c49d4029207b5ebda097e...svanoort:jenkins-27395-block-structure-lookup


src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
```diff
@@ -63,7 +69,19 @@
  * @author Kohsuke Kawaguchi
  * @author Jesse Glick
  */
-public abstract class FlowExecution implements FlowActionStorage {
+public abstract class FlowExecution implements FlowActionStorage, GraphLookupView {  // Implements GraphLookupView because FlowNode lives in another package
+
+    protected transient GraphLookupView internalGraphLookup = null;
+
+    /** Eventually this may be overridden if the FlowExecution has a better source of structural information, such as the {@link FlowNode} storage. */
+    protected synchronized GraphLookupView getInternalGraphLookup() {
```
jglick: Still recommend this be `private` or `@Restricted` unless there is a specific override in hand now.

src/main/java/org/jenkinsci/plugins/workflow/graph/GraphLookupView.java
```diff
@@ -0,0 +1,105 @@
...
+import javax.annotation.Nonnull;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+
+/**
+ * Interface that can be exposed by objects providing means to easily look up information about the structure of a pipeline run
+ * Usually this is scoped to a specific {@link org.jenkinsci.plugins.workflow.flow.FlowExecution}.
+ *
+ * Exists because we do not want to ourselves to only using the standard implementation in {@link StandardGraphLookupView}.
```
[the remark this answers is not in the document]
svanoort: Er, yes.

src/main/java/org/jenkinsci/plugins/workflow/graph/GraphLookupView.java
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * <strong>Implementation note:</strong>
+ * <p>Normally this should only be used internally to power APIs, but if exposed publicly remember that {@link FlowNode}s
```
jglick: Again recommend this be restricted unless and until there is a prepared patch using it from another plugin. Otherwise you are stuck with a possibly misdesigned API, forever.

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
```diff
@@ -108,17 +117,338 @@
...
         }
-        Collections.reverse(actual); // more readable
-        assertEquals(Arrays.asList(expected), actual);
+        Collections.sort(actual);
+        assertEquals(expectedList, actual);
+
+        // Now we clear the cache and try this again
+        Method m = FlowExecution.class.getDeclaredMethod("getInternalGraphLookup", null);
+        m.setAccessible(true);
+        Object ob = m.invoke((FlowExecution)(b.getExecution()), null);
```
jglick: If `@Restricted(NoExternalUse.class) public`, could be called directly.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java
```diff
@@ -113,13 +114,14 @@ public int compare(@CheckForNull FlowNode first, @CheckForNull FlowNode second)
      * Returns all {@link BlockStartNode}s enclosing the given FlowNode, starting from the inside out.
      * This is useful if we want to obtain information about its scope, such as the workspace, parallel branch, or label.
      * Warning: while this is efficient for one node, batch operations are far more efficient when handling many nodes.
+     * <p>Deprecated, prefer {@link FlowNode#iterateEnclosingBlocks()} instead, possibly using {@link FilteratorImpl#FilteratorImpl(Iterator, Predicate)}
```
jglick: Consider prototyping such a change against [this](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/044bbd653de6077b9d1b7871a618444253716647/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy#L254) and/or [this](https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/ce319baddf659aca192aec673f53f40c3ee51755/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L478).

svanoort: @jglick Actually, fetchEnclosingBlocks uses the new APIs now :)

svanoort: So basically you get the goodies without extra hassle.

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
svanoort: Yeah, I'd considered that option

PR #55, opened 2017-09-28
[JENKINS-47300] Provide APIs to track and advertise pipeline persistence guarantees

svanoort:
Provides [JENKINS-47300](https://issues.jenkins-ci.org/browse/JENKINS-47300) or durability levels and tracking for pipeline. 

This is allows the performance end of [JENKINS-33761](https://issues.jenkins-ci.org/browse/JENKINS-33761) by providing a way to track just how hard we're trying to protect our pipelines from failures on the master.   

Otherwise we have no way to know if it's safe to skip some optimizations or not. 

This is also a necessary precursor for [JENKINS-47173](https://issues.jenkins-ci.org/browse/JENKINS-47173) because we need a way to track durability requirements so we know when it's okay to take shortcuts for speed.

screen shot 2017-12-12 at 4 24 48 pm

# Features here
- [x] Add a Global setting UI under Manage Jenkins > Configure System
- [x] Provide an initial set of durability levels (performance-optimized AKA fast mode, max_survivability, aka current write-all-the-time mode, and one that is survivable but avoids the atomic write semantics for some filesystems)
- [x] Provide help text for durability modes and readable descriptions (plus **tooltips** for each mode)
- [x] Provide bindings to pass durability hints via FlowDefinition and retain them as  a field in the FlowExecution so they're set in stone once a run begins

# Downstreams:

* https://github.com/jenkinsci/workflow-support-plugin/pull/47
* https://github.com/jenkinsci/workflow-cps-plugin/pull/188
* https://github.com/jenkinsci/workflow-job-plugin/pull/75
* https://github.com/jenkinsci/workflow-multibranch-plugin/pull/68

# Snapshots
- Latest: 2.25-durability-v4-20171220.174016-1
- Deployed as: 2.25-durability-v3-20171212.214047-1
- Temp-V: 2.25-durability-v3-20171214.022350-2


svanoort:
@reviewbybees

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowDurabilityHint.java
```diff
@@ -0,0 +1,82 @@
...
+package org.jenkinsci.plugins.workflow.flow;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Provides hints about just how hard we should try to protect our workflow from failures of the master.
+ * There is a trade-off between durability and performance, with higher levels carrying much higher overheads to execution.
+ * Storage and persistence of data should try to provide at least the specified level (may offer more).
+ *
+ * <p> Implementation note: all implementations should be immutable - this is only an extension point rather than enum because
```
dwnusbaum: Exciting changes! Was reading through and noticed:
> this is only an extension point rather than enum because

which appears mismatched with the current code, which is an enum.

svanoort: Thanks @dwnusbaum - dangling remnant from a refactor, fixed now.

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/help-durabilityHint.html
```diff
@@ -0,0 +1,20 @@
+<div>
+    <p>
+        This setting provides a trade-off between how fast pipelines run and their "durability", or ability to survive unexpected failures while running.
```
abayer: This and all other doc/help stuff should get reviewed by @rtyler and @bitwiseman at a minimum.

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/help-durabilityHint.html
```diff
@@ -0,0 +1,20 @@
...
+
+        <br><strong>Facts:</strong>
+        <ul>
+            <li>The lowest durability modes (suggested) can be MUCH faster and write to disk FAR less often.</li>
+            <li>All pipelines persist their data when they complete/fail, are paused, or the master restarts gracefully.</li>
+            <li>Running pipelines that do NOT persist data with every step (lower durability settings) may lose data if Jenkins is not shut down gracefully.</li>
+            <li>Pipelines may not be able to resume or be visualized if they cannot persist data.</li>
+            <li>If a running pipeline loses data, you can always run the pipeline again.</li>
```
rtyler: Heh, this one isn't the most actionable or useful bit of information for an end-user. I suggest dropping it.

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/help-durabilityHint.html
```diff
@@ -0,0 +1,20 @@
+<div>
+    <p>
+        This setting provides a trade-off between how fast pipelines run and their "durability", or ability to survive unexpected failures while running.
+
+        <br><strong>Facts:</strong>
+        <ul>
+            <li>The lowest durability modes (suggested) can be MUCH faster and write to disk FAR less often.</li>
+            <li>All pipelines persist their data when they complete/fail, are paused, or the master restarts gracefully.</li>
```
rtyler: There's some verbiage in the setting description strings and this which refer to a "clean shutdown" or "graceful restart" which are somewhat ambiguous terms in my book. Is sending the JVM a SIGTERM considered a "clean shutdown" for the purposes of Pipeline durability?

Assuming it is, I suggest defining that somewhere in this help document (or hyperlinked elsewhere).

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/global.jelly
```diff
@@ -0,0 +1,17 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
+    <f:section title="Pipeline Speed/Durability Settings">
+        <f:entry title="${%Pipeline Default Speed/Durability Level}" field="durabilityHint">
+            <select class="setting-input" name="durabilityHint">
+                <f:option value="null" selected="${instance[field] == null}">None: use pipeline default (<j:out value="${descriptor.suggestedDurabilityHint.name()}" />)</f:option>
```
rtyler: Shouldn't the "None: use pipeline default" be in a messages.properties or whatever file?

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/help-durabilityHint.html
```diff
@@ -0,0 +1,20 @@
+<div>
```
rtyler: Please use a Title-cased "Pipeline" or "Pipelines" in this document to maintain consistency with other aspects of Jenkins Pipeline in the application.

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/help-durabilityHint.html
```diff
@@ -0,0 +1,20 @@
...
+    <p>
+        This setting provides a trade-off between how fast pipelines run and their "durability", or ability to survive unexpected failures while running.
+
+        <br><strong>Facts:</strong>
```
rtyler: This is a very Dwight Schrute way to explain this to the end-user 👿

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/global.jelly
```diff
@@ -0,0 +1,17 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
+    <f:section title="Pipeline Speed/Durability Settings">
```
rtyler: I strongly recommend _against_ using "Speed" in this configuration. I believe including "Speed" will encourage otherwise uneducated administrators and users to make a durability sacrifice, thinking their Pipelines will somehow run magically faster, with the "speedier" setting

If this section is simply "Pipeline Durability Settings" then I think we (CloudBees and Jenkins docs folk) can do the appropriate education in knowledgebase articles for fine-tuning Jenkins master performance to those who will actually benefit from the durability/speed trade-off..

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/help-durabilityHint.html
svanoort: So, @rtyler IIUC basically you're saying to overall just remove the help section?

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/global.jelly
svanoort: @rtyler That's actually what they *should* be doing and what we *want* them to do.  As long as you aren't randomly SIGKILL-ing Jenkins, you should be using PERFORMANCE_OPTIMIZED mode except for critical production deploys.  And there's nothing "magical" about things going exponentially faster when you're not writing a ton of redundant data to disk. 😉

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/help-durabilityHint.html
rtyler: No, this bullet point in particularly.

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/help-durabilityHint.html
svanoort: @abayer I don't see why, since we've never done this before for simple inline text?   I was planning to get them to review more fleshed out documentation content anyway.

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/global.jelly
rtyler: Basically all Jenkins instances are critical production deploys in my opinion. Jenkins is the life-blood of many organizations's delivery pipeline. 

My concern with this phrasing is that if the administrator is not well versed in the implications of these settings (read: everybody), conveying that it will change "Pipeline speed" is very likely going to be understood from the sense of Pipeline execution speed (wall-clock time): "Oh all my Pipelines will finish sooner now, this is like pushing in the Turbo button on my PC, great!" Sure that may occur, but the thrust of this setting's point as it should be conveyed to the user, is that this will change the Durability contract.
 
The vast majority of a Pipeline's wall-clock time in ci.jenkins.io is running Maven, enticing derpy admins like me to change this Durability setting for improved "Pipeline speed" is misleading.

svanoort: @rtyler Ah!  I think my phrasing was unclear -- what I mean by "production deploy" is a pipeline that is deploying code to a production system -- I consider *all* Jenkins instances to be production servers themselves and prioritize stability accordingly.   Downstream PRs add the ability to customize the durability mode on a per-pipeline or per-branch basis.   This means you can set up normal PR builds to just run as fast and light as possible to ensure that you don't overload a master (improving stability), but ensure that the most critical jobs *which deploy apps to prod* get the utmost level of protection.

The downstreams also explicitly test for the worst-case scenario to ensure that we gracefully handle it -- that scenario is an **unplanned** outage of a master while a pipeline is running in the performance_optimized mode.  In that scenario, the worst-case behavior is that the pipeline behaves like a freestyle build -- fails cleanly, retaining the build record and logs.

Only the transitory pipeline data is impacted by durability settings (persistence of the flowgraph, run itself, and program.dat) -- and we have mechanisms (with testing) to ensure they gracefully fail when data isn't persisted. 

Basically, I'm signing off that the chain of PRs should at no point reduce the stability of masters and should improve it overall, because it will prevent a class of outage caused by less-than-optimal pipeline code demanding excessive IO.

Is there a better way to communicate all the above?  I can understand why it's confusing.

> Oh all my Pipelines will finish sooner now, this is like pushing in the Turbo button on my PC, great

That's pretty much the behavior I expect for a lot of enterprise masters, and have a short list that will see that outcome.  

The difference between those and the ci.jenkins.io masters is that we have some of the world's greatest experts in Jenkins and Pipeline tuning our pipelines.  This allows us to easily minimize the overheads so that most of our time actually *is* running sh/bat steps -- but most organizations don't have easy access to the expertise in Jenkins that the Jenkins project does (naturally enough).  The results unfortunately speak for themselves. 

Basically: my goal is to make pipelines easier for people by removing some nasty performance surprises that hang people up -- and generally make them much lighter on the masters (reducing performance-related outages).

abayer:
I should mention that my biggest concerns around this change are users not understanding what they’re doing well enough and unexpectedly shooting themselves in the foot. Given the potential ramifications of changing their durability guarantees, I want to be as sure as we can be that we’re explaining the upsides and downsides of the options well, so that the users can be informed without having to refer to external documentation.

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/global.jelly
svanoort: Theoretically yes, but something like 90% of pipeline doesn't do this, so it's hardly worthwhile to add it in just this one single place.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowDefinition.java
```diff
@@ -42,6 +45,21 @@
  * @author Jesse Glick
  */
 public abstract class FlowDefinition extends AbstractDescribableImpl<FlowDefinition> implements ExtensionPoint {
+
+    protected FlowDurabilityHint myHint = null;
```
jglick: Delete; this does not belong here.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionOwner.java
```diff
@@ -52,6 +52,14 @@
     @Nonnull
     public abstract FlowExecution get() throws IOException;
 
+    /** Invoked in {@link FlowExecutionList#saveAll()} to notify that execution has been suspended */
+    protected void notifyShutdown() {
```
jglick: Delete `protected`; can use package access.

src/main/java/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel.java
```diff
@@ -0,0 +1,78 @@
...
+import net.sf.json.JSONObject;
+import org.kohsuke.stapler.StaplerRequest;
+
+import javax.annotation.CheckForNull;
+
+/**
+ * Supports a global default durability level for users
+ * @author Sam Van Oort
+ */
+public class GlobalDefaultFlowDurabilityLevel extends AbstractDescribableImpl<GlobalDefaultFlowDurabilityLevel>  {
```
jglick: Can be `@Restricted` if using an extension point as recommended elsewhere.

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/global.jelly
```diff
@@ -0,0 +1,17 @@
...
+                <j:forEach var="it" items="${descriptor.durabilityHintValues}">
+                    <j:set var="optionBody" encode="false">${it.description}</j:set>
+                    <option value="${it.name()}"
+                            selected="${(it==instance[field]) ? 'true':null}"
+                            tooltip="${it.tooltip}"><j:out value="${optionBody}"/></option>
+                </j:forEach>
+            </select>
+        </f:entry>
+    </f:section>
+    <f:section title=""></f:section>  <!-- Visual boundary -->
```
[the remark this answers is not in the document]
svanoort: Otherwise it smacks right up against an unrelated setting and produces horrible horrible UX because they look related.

src/main/java/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel.java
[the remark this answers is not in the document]
svanoort: @jglick Explain please.

src/main/java/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel.java
```diff
@@ -0,0 +1,78 @@
...
+
+        public void setDurabilityHint(FlowDurabilityHint hint){
+            this.durabilityHint = hint;
+            save();
+        }
+
+        @Override
+        public boolean configure(StaplerRequest req, JSONObject json) throws FormException {
+            // TODO verify if this is covered by permissions checks or we need an explicit check here.
+            Object ob = json.opt("durabilityHint");
```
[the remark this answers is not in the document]
svanoort: @jglick Does that actually work when we're mutating a field on the Descriptor?

src/main/java/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel.java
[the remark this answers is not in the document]
jglick: Extend `GlobalConfiguration` for global settings rather than rolling your own persistence & UI. See the Jenkins plugin archetype demonstrating its usage.

src/main/java/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel.java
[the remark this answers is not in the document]
jglick: If using `GlobalConfiguration`, yes; see stock sample.

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/global.jelly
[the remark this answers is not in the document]
jglick: Huh, I wonder whether the “unrelated setting” whichever that is needs its own `f:section`.

src/main/java/org/jenkinsci/plugins/workflow/flow/BlockableResume.java
```diff
@@ -0,0 +1,15 @@
+package org.jenkinsci.plugins.workflow.flow;
+
+/**
+ * Can be added to advertise the ability to mark pipeline components which prevent pipelines from being able to resume
+ * after restart or after pause.
+ *
```
jglick:
```html
<p>
```

?

src/main/java/org/jenkinsci/plugins/workflow/flow/BlockableResume.java
```diff
@@ -0,0 +1,15 @@
+package org.jenkinsci.plugins.workflow.flow;
+
+/**
+ * Can be added to advertise the ability to mark pipeline components which prevent pipelines from being able to resume
```
jglick: Unclear from this wording—added to what? What kinds of “components” are being discussed? Is there a specific `add` method you could `link` to in order to help readers understand?

src/main/java/org/jenkinsci/plugins/workflow/flow/BlockableResume.java
```diff
@@ -0,0 +1,15 @@
...
+ * after restart or after pause.
+ *
+ * Pipelines which cannot resume will simply fail.
+ */
+public interface BlockableResume {
+    /** Return true if we prevent the abiity to resume. */
+    public boolean isResumeBlocked();
+
+    /** Set resume on or off - may throw an {@link IllegalArgumentException} if trying to illegally toggle. */
```
jglick: `IllegalStateException` is probably what you meant.

Under what conditions would it be “illegal” to toggle this? As discussed elsewhere, I suspect this could be deleted.

src/main/java/org/jenkinsci/plugins/workflow/flow/DurabilityHintProvider.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+
+/**
+ * Provides a way to indirectly register durability settings to apply to pipelines.
+ */
+public interface DurabilityHintProvider extends ExtensionPoint {
+
+
+    public int ordinal();
```
jglick: Unnecessary—this is defined in `Extension`.

src/main/java/org/jenkinsci/plugins/workflow/flow/DurabilityHintProvider.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @CheckForNull
+    FlowDurabilityHint suggestFor(@Nonnull Item x);
+
+    public static @Nonnull FlowDurabilityHint suggestedFor(@Nonnull Item x) {
+        int ordinal = Integer.MAX_VALUE;
+        FlowDurabilityHint hint = GlobalDefaultFlowDurabilityLevel.getDefaultDurabilityHint();
```
jglick: `GlobalDefaultFlowDurabilityLevel` could be made to have a `DurabilityHintProvider` for consistency.

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/global.jelly
```diff
@@ -13,5 +13,5 @@
             </select>
         </f:entry>
     </f:section>
-    <f:section title=""></f:section>  <!-- Visual boundary -->
+    <br/><br/>
```
jglick: Be careful—you might be inside a `<tr>` here, in which case `<br/>` is illegal content.

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/help-durabilityHint.html
```diff
@@ -1,20 +1,27 @@
 <div>
     <p>
-        This setting provides a trade-off between how fast pipelines run and their "durability", or ability to survive unexpected failures while running.
+        This setting allows users to change the default durability mode for running Pipelines.  In most cases this is a
+        trade-off between performance and the ability for running pipelines to resume after unplanned Jenkins outages.
 
-        <br><strong>Facts:</strong>
+        <br><br><strong>What does this do?</strong>
```
jglick: Prefer semantic `<p>` BTW.

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/help-durabilityHint.html
```diff
@@ -1,20 +1,27 @@
...
+            <li>Previously, running pipelines wrote data constantly, so that they could resume even if Jenkins fails.
+                This setting gives the user the ability to adjust this write behavior.</li>
+            <li>Higher-performance/lower-durability modes write data to disk <strong>much</strong> less often for running pipelines.</li>
+            <li>Writing data less often can massively reduce build times for Pipelines with many steps or complex logic.
+                For pipelines which spend most of their time waiting for a shell/batch script to run, the difference will be less visible.</li>
+            <li>Running pipelines with lower durability settings may lose data <strong>if</strong> they do not finish and Jenkins is not shut down gracefully:</li>
+                <ul>
+                    <li>A "graceful" shutdown where Jenkins goes through a full shutdown process, such as visiting <code>http://[jenkins-server]/exit</code></li>
+                    <li>A "dirty" shutdown, such as using <code>kill -9</code> to terminate the Jenkins process, may prevent incomplete pipelines from persisting data</li>
```
jglick: Note that `SIGTERM` might be graceful or dirty, depending on the servlet container.

src/main/java/org/jenkinsci/plugins/workflow/flow/DurabilityHintProvider.java
svanoort: Yes, I'm aware -- I want this to be separate from extension loading.

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/help-durabilityHint.html
svanoort: I agree it needs to be mentioned, but this isn't the place -- too much overcrowding and a fairly small detail.

The place for that is more detailed documentation that will go on Jenkins.io, where I plan to explain in more depth how to guarantee clean shutdowns and what may cause dirty ones.

src/main/java/org/jenkinsci/plugins/workflow/flow/DurabilityHintProvider.java
svanoort: Implementing it this way allows delegating to GlobalDefaultFlowDurabilityLevel if other implementations do not provide a value - no real benefit to changing it here.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
```diff
@@ -73,6 +73,9 @@
 
     protected transient GraphLookupView internalGraphLookup = null;
 
+    @CheckForNull /** CheckForNull due to loading pre-durability runs. */
+    protected FlowDurabilityHint durabilityHint = null;
```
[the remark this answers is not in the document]
svanoort: CpsFlowExecution needs access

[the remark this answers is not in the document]
jglick: Then make an accessor. Do not use `public` or `protected` fields.

src/main/java/org/jenkinsci/plugins/workflow/flow/DurabilityHintProvider.java
jglick: Huh? Why?

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowDurabilityHint.java
```diff
@@ -33,14 +33,11 @@
  * @author Sam Van Oort
  */
 public enum FlowDurabilityHint {
-    PERFORMANCE_OPTIMIZED(false, false, "Performance-optimized: much faster (requires clean shutdown to save running pipelines)",
-            "Avoids writing data with every step, avoids atomic writes of data.  Pipelines can resume if Jenkins shuts down cleanly, but running pipelines lose step information and can't resume if Jenkins unexpectedly fails."),
+    PERFORMANCE_OPTIMIZED(false, false, Messages.FlowDurabilityHint_PERFORMANCE_OPTIMIZED_description(), Messages.FlowDurabilityHint_PERFORMANCE_OPTIMIZED_tooltip()),
```
jglick: These should be `Localizable`s: `Messages._FlowDurabilityHint_PERFORMANCE_OPTIMIZED_description()`, etc.

svanoort: @jglick  Huh?  They are Localizables, using Messages.xxxx

jglick: Nope, you did not see or understand my earlier review comment I guess—these fields in the enum should be of type `Localizable` not `String`, rendered if and when displayed, which is why I said `Messages._FlowDurabilityHint_PERFORMANCE_OPTIMIZED_description()` not `Messages.FlowDurabilityHint_PERFORMANCE_OPTIMIZED_description()`.

Anyway, not important to change now, as it can be fixed later without breaking API compatibility.

svanoort: Ah, I see, I missed your underscore -- yeah, it's a subtle difference (this one is localized to the master, the other might be localized differently). 

Definitely worth a fix later though.

PR #56, opened 2017-10-03
[JENKINS-40912] Add StashManager.fileNamesForStash(Run,String)

abayer:
[JENKINS-40912](https://issues.jenkins-ci.org/browse/JENKINS-40912)

upstream of https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/52

Might be a better way to do this - I don't love it, but I also don't know of a better way to get this information.

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
@@ -24,6 +24,7 @@
 
 package org.jenkinsci.plugins.workflow.flow;
 
+import com.thoughtworks.xstream.XStream;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.AbortException;
 import hudson.Extension;
@@ -32,10 +33,14 @@ import hudson.ExtensionPoint;
 import hudson.FilePath;
 import hudson.Launcher.LocalLauncher;
 import hudson.Util;
+import hudson.XmlFile;
 import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.org.apache.tools.tar.TarInputStream;
+import hudson.remoting.VirtualChannel;
 import hudson.util.DirScanner;
+import hudson.util.FileVisitor;
+import hudson.util.XStream2;
 import hudson.util.io.ArchiverFactory;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
@@ -45,12 +50,16 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+
+import jenkins.FilePathFilter;
+import jenkins.MasterToSlaveFileCallable;
 import jenkins.model.ArtifactManager;
 import jenkins.model.Jenkins;
 import jenkins.util.BuildListenerAdapter;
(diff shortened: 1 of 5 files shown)


src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -165,6 +167,38 @@ public static void copyAll(@Nonnull Run<?,?> from, @Nonnull Run<?,?> to) throws
...
+     * Get a list of files included in a particular stash.
+     *
+     * @param build The build to look in
+     * @param name The stash name to look for
+     * @return A list of file names included in that stash, if it exists. Otherwise, an empty list.
+     * @throws IOException If errors occur while reading from the archived stash tarball.
+     */
+    public static List<String> fileNamesForStash(@Nonnull Run<?,?> build, @Nonnull String name) throws IOException {
+        List<String> fileNames = new ArrayList<>();
+        File[] kids = storage(build).listFiles();
```
michaelneale: nice! So this is the sauce... very cool

svanoort:
Looking more closely at how this is used in the downstream API, I'm not a fan of double-extracting files this way -- is there not a way we could accomplish this without reading those files twice?

I'd say "sure why not" if it were something were only doing on demand, but stash can already be kind of painful for oversized files.

michaelneale:
@svanoort deeper in core when things are passed in to TarArchiver perhaps (but bigger deal to change)

abayer:
@svanoort Yeah, I don't love it either, but can't figure out a way to get this without making non-trivial changes to `TarArchiver` in core. fwiw, the downstream does test paths.

michaelneale:
I don't think the overhead itself is the problem, as long as it is accurate. When people are doing this they are "debugging" effectively (and over time, I hope no one needs to use it)

abayer:
fwiw, latest change switches to using a `DirScanner` to get the list of files at `stash` time, which is then preserved in an XML file alongside the stashed tarball for easy querying at `unstash` time. No more duplication of decompression.

abayer:
@svanoort See, I'd say the problem is that `StashManager` is here and not in `workflow-basic-steps` - nothing else, so far as I can tell, in here or anywhere other than `workflow-basic-steps` does anything with `StashManager`, so `workflow-basic-steps` is the logical place for testing to be.

svanoort:
@abayer  The tests ought to live alongside the main logic so that we don't have to do some contortions to test changes.  

Even though the StashManger lives here (annoyingly enough), you can always pull the interesting & fiddly bits here into a package-visibility helper class and write tests against that. 

If there's not a workable way to do that, maybe this logic needs to be in the workflow-basic-steps in the first place, and just pass the results into the StashManger?

abayer:
@svanoort So here's why I say to just keep the tests in `workflow-basic-steps`: because that's where we're already testing everything `stash`-related. Why split the tests up? What value would that provide? Just feels like busywork to me.

svanoort:
@abayer If I can't make changes without running tests in a downstream plugin, that's unmaintainable code -- fixing that is hardly "busywork" -- it's simple maintainability.  I honestly don't care if this logic lives here or there, just as long as  its tests are in the same plugin it is.  

See also: lessons learned about why groovy-cps is annoying to change in practice (because you have to go back to workflow-cps to see if things actually work right).

abayer:
...ok then. How about we verify that `StashManager` isn't used by anything else, and assuming that's the case, just move it to `workflow-basic-steps`?

abayer:
aw crap, it's used in `WorkflowRun`.

abayer:
Tried a test using `stash` in `workflow-api`, since we've got a test dependency on `workflow-basic-steps`, and that hung due to classloader stuff. I really see no way to test `StashManager` as a whole, not just this one part, without rewriting `stash`.

abayer:
Well, fwiw, it works when I use the `workflow-basic-steps` SNAPSHOT.

abayer:
@svanoort Added `fakeStash` and `fakeUnstash` steps in test. That's really the best I can come up with, so if that doesn't suffice, I'll table this change until someone else has a better idea.

jglick:
> Tried a test using `stash` in `workflow-api`, since we've got a test dependency on `workflow-basic-steps`

Sure, that should work.

> that hung due to classloader stuff.

Huh? What “class loader stuff”?

svanoort:
@abayer This looks reasonable to me as it stands, but am still waiting for you to fix the tests in order to be able to approve it.

svanoort:
Mergeable/releaseable whenever y'like, @abayer

michaelneale:
very nice

abayer:
Thanks, @svanoort!

PR #60, opened 2017-11-15
getEnclosingBlocks should return a List<? extends BlockStartNode>, not a List<FlowNode>

jglick:
Silly to use an unchecked cast to widen the return type, when the method’s very Javadoc is explicit that it will return `BlockStartNode`s specifically.

Also using a wildcard to emphasize that the result is read-only.

Not necessarily source-compatible, but is binary-compatible.

@reviewbybees


jglick:
@reviewbybees done

PR #62, opened 2017-12-06
[FIXED JENKINS-40912] Add method for getting stash file list

abayer:
[JENKINS-40912](https://issues.jenkins-ci.org/browse/JENKINS-40912)

Replaces busted #56 

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
@@ -171,7 +171,7 @@ public class StashManager {
      * Get the filenames included in a particular stash.
      * @param build The build to look in.
      * @param stashName The stash name to look up.
-     * @return A possibly empty list of filenames included in the given stash name in the given build.
+     * @return A possibly empty list of filenames included in the given stash name in the given build. Unix separators will be used.
      * @throws IOException If unable to interact with the stash files.
      */
     @Nonnull
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
@@ -44,7 +44,9 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
 import javax.annotation.CheckForNull;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -165,6 +167,35 @@ public static void copyAll(@Nonnull Run<?,?> from, @Nonnull Run<?,?> to) throws
...
+     * @throws IOException If unable to interact with the stash files.
+     */
+    @Nonnull
+    public static List<String> stashFileList(@Nonnull Run<?,?> build, @Nonnull String stashName) throws IOException {
+        List<String> contents = new ArrayList<>();
+        File[] kids = storage(build).listFiles();
+        if (kids != null) {
+            for (File kid : kids) {
+                String n = kid.getName();
+                if (n.endsWith(SUFFIX) && stashName.equals(n.substring(0, n.length() - SUFFIX.length()))) {
```
jglick: Would be easier to look up the expected child name directly: `storage(build, stashName)`

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -165,6 +167,30 @@ public static void copyAll(@Nonnull Run<?,?> from, @Nonnull Run<?,?> to) throws
...
+    /**
+     * Get the filenames included in a particular stash.
+     * @param build The build to look in.
+     * @param stashName The stash name to look up.
+     * @return A possibly empty list of filenames included in the given stash name in the given build. Unix separators will be used.
+     * @throws IOException If unable to interact with the stash files.
+     */
+    @Nonnull
+    public static List<String> stashFileList(@Nonnull Run<?,?> build, @Nonnull String stashName) throws IOException {
+        List<String> contents = new ArrayList<>();
```
jglick: Why not `Set<String>`?

src/test/java/org/jenkinsci/plugins/workflow/flow/StashManagerTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+                List<String> secondList = StashManager.stashFileList(b, "second");
+                assertFalse(secondList.isEmpty());
+                assertEquals(1, secondList.size());
+                assertTrue(secondList.contains("elsewhere/nested"));
+            }
+        });
+    }
+
+    private static final class HasWorkspacePredicate implements Predicate<FlowNode> {
```
jglick: There is a utility for this in `workflow-api`.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
svanoort: `Set` is probably more technically correct, but I think `List` will be easier to work with due to positional access.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -165,6 +167,30 @@ public static void copyAll(@Nonnull Run<?,?> from, @Nonnull Run<?,?> to) throws
         FileUtils.copyDirectory(fromStorage, storage(to));
     }
 
+    /**
+     * Get the filenames included in a particular stash.
```
svanoort: Should include a hard warning that this extracts the whole stash -- this is not a free or cheap operation.

src/test/java/org/jenkinsci/plugins/workflow/flow/StashManagerTest.java
svanoort: Yup, `FlowScanningUtils.hasActionPredicate(Class<? extends Action)` IIRC  🐜

PR #63, opened 2017-12-07
[JENKINS-43995] Add status to FlowNodes

abayer:
[JENKINS-43995](https://issues.jenkins-ci.org/browse/JENKINS-43995)
This is tracked via the new `FlowNodeStatusAction`, but could be
switched to using `TagsAction` if desired. It's actually set over in
workflow-cps.

`AtomNode`s return their own status for `getStatus()`. `BlockEndNode`s also
return their own status. `BlockStartNode`s look to their `BlockEndNode`s,
or return null.

Also added caching for looking up the immediate children of a
`BlockStartNode`. That logic almost certainly can use improvement.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/FlowNodeStatusAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/FlowNodeStatusAction.java
@@ -1,0 +1,34 @@
+package org.jenkinsci.plugins.workflow.actions;
+
+import hudson.model.Result;
+
+import javax.annotation.Nonnull;
+
+public final class FlowNodeStatusAction implements PersistentAction {
+    private final Result result;
+
+    public FlowNodeStatusAction(@Nonnull Result result) {
+        this.result = result;
+    }
+
+    @Nonnull
+    public Result getResult() {
+        return result;
+    }
+
+    @Override
+    public String getIconFileName() {
+        return null;  // If we add one then we can easily use this with UI renderings
+    }
+
+    @Override
+    public String getDisplayName() {
+        return "Status";
+    }
+
+    @Override
+    public String getUrlName() {
+        return null;
+    }
+
+}
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/BlockEndNode.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/BlockEndNode.java
@@ -25,11 +25,17 @@
 package org.jenkinsci.plugins.workflow.graph;
 
 import java.io.IOException;
+
+import hudson.model.Result;
+import org.jenkinsci.plugins.workflow.actions.ErrorAction;
+import org.jenkinsci.plugins.workflow.actions.FlowNodeStatusAction;
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
+import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
 
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 
 /**
@@ -69,6 +75,26 @@ public abstract class BlockEndNode<START extends BlockStartNode> extends FlowNod
             }
         }
         return start;
+    }
+
+    @CheckForNull
+    @Override
+    public Result getStatus() {
+        ErrorAction errorAction = getError();
+        if (errorAction != null) {
+            if (errorAction.getError() instanceof FlowInterruptedException) {
+                return Result.ABORTED;
+            } else {
+                return Result.FAILURE;
+            }
+        } else {
+            FlowNodeStatusAction statusAction = getPersistentAction(FlowNodeStatusAction.class);
+            if (statusAction != null) {
+                return statusAction.getResult();
+            }
+        }
+        // If there's no error on the end node or status action on the end node, return success.
+        return Result.SUCCESS;
     }
 
 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/BlockEndNode.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/BlockEndNode.java
@@ -26,16 +26,9 @@ package org.jenkinsci.plugins.workflow.graph;
 
 import java.io.IOException;
 
-import hudson.model.Result;
-import org.jenkinsci.plugins.workflow.actions.ErrorAction;
-import org.jenkinsci.plugins.workflow.actions.FlowNodeStatusAction;
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
-import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
 
 import java.util.List;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 
 /**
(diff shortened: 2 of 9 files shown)


src/main/java/org/jenkinsci/plugins/workflow/actions/FlowNodeStatusAction.java
```diff
@@ -0,0 +1,34 @@
+package org.jenkinsci.plugins.workflow.actions;
+
+import hudson.model.Result;
+
+import javax.annotation.Nonnull;
+
+public final class FlowNodeStatusAction implements PersistentAction {
+    private final Result result;
```
abayer: fwiw, we might want to actually go with moving [`GenericStatus`](https://github.com/jenkinsci/pipeline-graph-analysis-plugin/blob/master/src/main/java/org/jenkinsci/plugins/workflow/pipelinegraphanalysis/GenericStatus.java) into `workflow-api` and using that instead of `Result`.

svanoort: I kind of like that idea actually.

svanoort: Yeah, that might make some sense. Not sure what @jglick would think though, maybe there's a concrete reason not to -- we might need to provide a mapping back and forth from Result though.

src/main/java/org/jenkinsci/plugins/workflow/graph/BlockEndNode.java
```diff
@@ -71,4 +77,24 @@ public BlockEndNode(FlowExecution exec, String id, START start, List<FlowNode> p
...
+                return Result.FAILURE;
+            }
+        } else {
+            FlowNodeStatusAction statusAction = getPersistentAction(FlowNodeStatusAction.class);
+            if (statusAction != null) {
+                return statusAction.getResult();
+            }
+        }
+        // If there's no error on the end node or status action on the end node, return success.
+        return Result.SUCCESS;
```
svanoort: May be better to return null (or GenericStatus.IN_PROGRESS if we swap the type) if node is not completed.

abayer: Since this is the end node, I'm working under the assumption that it's complete if it exists at all.

src/main/java/org/jenkinsci/plugins/workflow/actions/FlowNodeStatusAction.java
abayer: Oh, we'll definitely need that mapping. I'm assuming that if we move `GenericStatus` here, we'll also be adding logic to do that mapping.

src/main/java/org/jenkinsci/plugins/workflow/graph/BlockEndNode.java
svanoort: That's not true -- remember there's a window of time when a FlowNode is being configured before it's completed, and we also have nested block structures doing various things.

abayer: Okiedokie, then I'll change it to return null (for now, eventually `GenericStatus.IN_PROGRESS`) if `isActive()` is true and `Result.SUCCESS` (until `GenericStatus` blah blah etc) if not.

src/main/java/org/jenkinsci/plugins/workflow/graph/BlockStartNode.java
```diff
@@ -51,4 +56,23 @@ protected BlockStartNode(FlowExecution exec, String id, List<FlowNode> parents)
...
+        return this.getExecution().getImmediateChildrenForBlockStart(this);
+    }
+
+    @CheckForNull
+    @Override
+    public Result getStatus() {
+        BlockEndNode endNode = getEndNode();
+        // Only check for errors or explicit status if this block is complete.
+        if (endNode != null) {
+            return endNode.getStatus();
```
svanoort: Will this handle a block that may itself may fail when run?  I.E. Dir step that throws IOException?

abayer: Good question! Needs a test.

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
```diff
@@ -33,9 +30,22 @@
     /** Map a node to its nearest enclosing block */
     HashMap<String, String> nearestEnclosingBlock = new HashMap<String, String>();
 
+    /** Map a block start node to its immediate children nodes, including AtomNode, BlockStartNode, and BlockEndNode */
+    HashMap<String, List<String>> blockChildren = new HashMap<>();
```
svanoort: Not a fan of caching this, since it's fairly easy and fast to do a quick iteration pass once we have the start and end -- that can even be exposed as an Iterator<FlowNode> and never materialized into a concrete list, due to the magic of FlowScanners -- they're Iterables so you can do a for :: each over the flownodes in a scanner, directly. 

Yes, that scan is O(n), but you're going to be iterating over children anyway.

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
```diff
@@ -33,9 +30,22 @@
...
+
     public void clearCache() {
         blockStartToEnd.clear();
         nearestEnclosingBlock.clear();
+        blockChildren.clear();
+    }
+
+    private void addToBlockChildren(@Nonnull String blockId, @Nonnull String childId) {
```
svanoort: If we just directly expose an iterable/iterator over children, no need for this.

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
```diff
@@ -221,4 +228,34 @@ public BlockStartNode findEnclosingBlockStart(@Nonnull FlowNode node) {
         }
         return starts;
     }
+
+    @Nonnull
+    @Override
+    public List<FlowNode> getImmediateChildrenForBlockStart(@Nonnull BlockStartNode start) {
```
svanoort: Questions I have here: 

- What about the ordering of children?
- This feels needlessly complicated.  I think there's a need for tree navigation APIs in some sense, but (see downstream PR comments) I don't see it as valid for this one. 

Long-term I kind of lean to storing tree/DAG structural information in the Storage engine itself, with these APIs just delegating to that (that's part of why it is implemented the way it is) -- one reason I don't want to go too much further into this graphLookupView implementation.    It's a placeholder to tide us over for future storage work, ideally.

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
abayer: Yeah, I honestly wasn't sure what the right approach would be, but decided to bang out the prototype just in case.

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
abayer: Ah, yeah, the ordering would be inverse here, wouldn't it?

src/main/java/org/jenkinsci/plugins/workflow/graph/BlockStartNode.java
abayer: And now has one.

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
svanoort: Well, back-to-forward in linear flows, but parallels don't have a defined ordering (that's one reason for different scanners).

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -319,7 +321,26 @@ public String getUrl() throws IOException {
     }
 
     @CheckForNull
-    public abstract Result getStatus();
+    public Result getStatus() {
```
svanoort: Yeah this is a better place for it IMO

dwnusbaum:
Subsumed by https://github.com/jenkinsci/workflow-api-plugin/pull/91 and related PRs.

PR #64, opened 2018-01-18
[JENKINS-49025] Extending JENKINS-34488 fix yet again

jglick:
[JENKINS-49025](https://issues.jenkins-ci.org/browse/JENKINS-49025)

#32 only considered causes, but other Throwable fields could also be unserializable. https://github.com/jenkinsci/pipeline-aws-plugin/pull/37 (after suppressing whitelist and running on master node) demonstrates that other things can go wrong and cause haywire. The only way to find out for sure if we can serialize a given exception is to try it. Adds a bit of overhead but better than the alternative.

@reviewbybees


jglick:
@reviewbybees done

oleg-nenashev:
👍 . I will update the Wiki Page

oleg-nenashev:
@abayer @svanoort Could you please release it? Due to the internal testing process in my company we need it as soon as possible

jglick:
cf. https://github.com/jenkinsci/jenkins/pull/3358#pullrequestreview-105050031

PR #66, opened 2018-02-21
Javadocs: document iteration order for enclosing blocks APIs

svanoort:
@reviewbybees


svanoort:
Requested by @vivek

vivek:
Thanks @svanoort!

PR #67, opened 2018-02-28
[JENKINS-49635] VirtualFile support

jglick:
Downstream of https://github.com/jenkinsci/jenkins/pull/3302 & https://github.com/jenkinsci/plugin-pom/pull/98. Subsumes #74.

- [X] reusable test facility
- [x] `StashManager` integration
- [x] stronger tests

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.7-20180326.223441-1</version> <!-- TODO https://github.com/jenkinsci/plugin-pom/pull/98 -->
+        <version>3.7</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.9</version>
+        <version>3.10-20180509.203140-1</version> <!-- TODO https://github.com/jenkinsci/plugin-pom/pull/105 -->
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -159,5 +159,11 @@
             <version>1.0</version>
             <scope>test</scope>
         </dependency>
+      <dependency>
+          <groupId>org.jenkins-ci.plugins</groupId>
+          <artifactId>apache-httpcomponents-client-4-api</artifactId>
+          <version>4.5.5-2.1</version>
+          <scope>test</scope>
+      </dependency>
     </dependencies>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.10-20180509.203140-1</version> <!-- TODO https://github.com/jenkinsci/plugin-pom/pull/105 -->
+        <version>3.10</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -144,7 +144,7 @@
         <dependency>
             <groupId>org.jenkins-ci.test</groupId>
             <artifactId>docker-fixtures</artifactId>
-            <version>1.7</version>
+            <version>1.8</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.5</version>
+        <version>3.6</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -62,8 +62,8 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins-core.version>2.110-20180228.232138-2</jenkins-core.version> <!-- TODO https://github.com/jenkinsci/jenkins/pull/3302 -->
-        <jenkins-war.version>2.110-20180228.232235-2</jenkins-war.version>
+        <jenkins-core.version>2.112-20180314.180003-1</jenkins-core.version> <!-- TODO https://github.com/jenkinsci/jenkins/pull/3302 -->
+        <jenkins-war.version>2.112-20180314.180029-1</jenkins-war.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.2</version>
+        <version>3.5</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -62,7 +62,8 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>2.60.3</jenkins.version>
+        <jenkins-core.version>2.110-20180228.232138-2</jenkins-core.version> <!-- TODO https://github.com/jenkinsci/jenkins/pull/3302 -->
+        <jenkins-war.version>2.110-20180228.232235-2</jenkins-war.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.10</version>
+        <version>3.12</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.28</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.118</jenkins.version>
+        <jenkins.version>2.121</jenkins.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
--- a/pom.xml
+++ b/pom.xml
@@ -143,7 +143,7 @@
         <dependency>
             <groupId>org.jenkins-ci.test</groupId>
             <artifactId>docker-fixtures</artifactId>
-            <version>1.7-20180404.203821-2</version> <!-- TODO https://github.com/jenkinsci/docker-fixtures/pull/19 -->
+            <version>1.7</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-api</artifactId>
-    <version>2.26</version>
+    <version>2.27-SNAPSHOT</version>
     <packaging>hpi</packaging>
     <name>Pipeline: API</name>
     <url>https://wiki.jenkins.io/display/JENKINS/Pipeline+API+Plugin</url>
@@ -47,7 +47,7 @@
         <connection>scm:git:git://github.com/jenkinsci/${project.artifactId}-plugin.git</connection>
         <developerConnection>scm:git:git@github.com:jenkinsci/${project.artifactId}-plugin.git</developerConnection>
         <url>https://github.com/jenkinsci/${project.artifactId}-plugin</url>
-        <tag>workflow-api-2.26</tag>
+        <tag>HEAD</tag>
     </scm>
     <repositories>
         <repository>
--- a/pom.xml
+++ b/pom.xml
@@ -62,8 +62,7 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins-core.version>2.118-20180416.205358-1</jenkins-core.version> <!-- TODO https://github.com/jenkinsci/jenkins/pull/3302 -->
-        <jenkins-war.version>2.118-20180416.205448-1</jenkins-war.version>
+        <jenkins.version>2.118</jenkins.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
--- a/pom.xml
+++ b/pom.xml
@@ -140,5 +140,23 @@
             <artifactId>structs</artifactId>
             <version>1.7</version>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.test</groupId>
+            <artifactId>docker-fixtures</artifactId>
+            <version>1.7-20180404.162022-1</version> <!-- TODO https://github.com/jenkinsci/docker-fixtures/pull/19 -->
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>ssh-slaves</artifactId>
+            <version>1.26</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>jdk-tool</artifactId>
+            <version>1.0</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -62,8 +62,8 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins-core.version>2.114-20180326.224001-1</jenkins-core.version> <!-- TODO https://github.com/jenkinsci/jenkins/pull/3302 -->
-        <jenkins-war.version>2.114-20180326.224035-1</jenkins-war.version>
+        <jenkins-core.version>2.115-20180402.200252-1</jenkins-core.version> <!-- TODO https://github.com/jenkinsci/jenkins/pull/3302 -->
+        <jenkins-war.version>2.115-20180402.200314-1</jenkins-war.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
--- a/pom.xml
+++ b/pom.xml
@@ -62,8 +62,8 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins-core.version>2.115-20180402.200252-1</jenkins-core.version> <!-- TODO https://github.com/jenkinsci/jenkins/pull/3302 -->
-        <jenkins-war.version>2.115-20180402.200314-1</jenkins-war.version>
+        <jenkins-core.version>2.118-20180416.205358-1</jenkins-core.version> <!-- TODO https://github.com/jenkinsci/jenkins/pull/3302 -->
+        <jenkins-war.version>2.118-20180416.205448-1</jenkins-war.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
--- a/pom.xml
+++ b/pom.xml
@@ -143,7 +143,7 @@
         <dependency>
             <groupId>org.jenkins-ci.test</groupId>
             <artifactId>docker-fixtures</artifactId>
-            <version>1.7-20180404.162022-1</version> <!-- TODO https://github.com/jenkinsci/docker-fixtures/pull/19 -->
+            <version>1.7-20180404.203821-2</version> <!-- TODO https://github.com/jenkinsci/docker-fixtures/pull/19 -->
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -28,12 +28,12 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.7</version>
+        <version>3.9</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-api</artifactId>
-    <version>2.28-SNAPSHOT</version>
+    <version>${revision}${changelist}</version>
     <packaging>hpi</packaging>
     <name>Pipeline: API</name>
     <url>https://wiki.jenkins.io/display/JENKINS/Pipeline+API+Plugin</url>
@@ -47,7 +47,7 @@
         <connection>scm:git:git://github.com/jenkinsci/${project.artifactId}-plugin.git</connection>
         <developerConnection>scm:git:git@github.com:jenkinsci/${project.artifactId}-plugin.git</developerConnection>
         <url>https://github.com/jenkinsci/${project.artifactId}-plugin</url>
-        <tag>HEAD</tag>
+        <tag>${scmTag}</tag>
     </scm>
     <repositories>
         <repository>
@@ -62,6 +62,8 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
+        <revision>2.28</revision>
+        <changelist>-SNAPSHOT</changelist>
         <jenkins.version>2.118</jenkins.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.6</version>
+        <version>3.7-20180326.223441-1</version> <!-- TODO https://github.com/jenkinsci/plugin-pom/pull/98 -->
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -62,11 +62,12 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins-core.version>2.112-20180314.180003-1</jenkins-core.version> <!-- TODO https://github.com/jenkinsci/jenkins/pull/3302 -->
-        <jenkins-war.version>2.112-20180314.180029-1</jenkins-war.version>
+        <jenkins-core.version>2.114-20180326.224001-1</jenkins-core.version> <!-- TODO https://github.com/jenkinsci/jenkins/pull/3302 -->
+        <jenkins-war.version>2.114-20180326.224035-1</jenkins-war.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
+        <useBeta>true</useBeta>
     </properties>
     <dependencies>
         <dependency>
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
@@ -291,6 +291,9 @@ public class StashManager {
 
     /**
      * Mixin interface for an {@link ArtifactManager} which supports specialized stash behavior as well.
+     * The recommended standard implementation is in the plugin currently named {@code artifact-manager-s3},
+     * which in turn supports extensibility to various cloud providers.
+     * @see <a href="https://github.com/jenkinsci/jep/blob/master/jep/202/README.adoc">JEP-202</a>
      */
     @Restricted(Beta.class)
     public interface StashAwareArtifactManager /* extends ArtifactManager */ {
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
@@ -26,12 +26,16 @@ package org.jenkinsci.plugins.workflow.flow;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.AbortException;
+import hudson.EnvVars;
 import hudson.Extension;
 import hudson.ExtensionList;
 import hudson.ExtensionPoint;
 import hudson.FilePath;
+import hudson.Launcher;
 import hudson.Launcher.LocalLauncher;
 import hudson.Util;
+import hudson.model.Computer;
+import hudson.model.Node;
 import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.org.apache.tools.tar.TarInputStream;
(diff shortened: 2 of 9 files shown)


pom.xml
```diff
@@ -62,7 +62,8 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>2.60.3</jenkins.version>
+        <jenkins-core.version>2.110-20180228.232138-2</jenkins-core.version> <!-- TODO https://github.com/jenkinsci/jenkins/pull/3302 -->
```
svanoort: 🐛 Hard nope for me right here.   We do not accept gratuitous core bumps especially to such a bleeding edge core versions because it prevents anybody from being able to consume new features and bugfixes to the pipeline plugins.

abayer: I agree. At a minimum this needs to hit LTS first.

svanoort: Prefer 2 LTS cycles because otherwise it's hard to ship fixes to many of the users (especially important for urgent issues) that tend to run an LTS behind but yeah, 1 LTS is the bare minimum for me too -- should be reserved for critical fixes and super high-value features generally though.

svanoort:
Note: that's only a time-sensitive merge-block, not a permanent one.  I'm assuming that this is a speculative WIP PR so that won't be an issue and by the time it's release-ready the dependency will have been in an LTS or two.

jglick:
Well, no. See [JEP-300](https://github.com/jenkinsci/jep/blob/master/jep/300/README.adoc#automatically-updated-distribution).

Anyway that is parenthetical—obviously no one was proposing a PR like this be merged today. Please reserve PR reviews to substantive matters of the code change itself to avoid visual clutter. It is helpful to create (I think conventionally orange?) `on-hold` labels as a marker for PRs which are not eligible for immediate merging to their target branch for logistical reasons, at least until Essentials is live.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -228,16 +286,77 @@ public boolean shouldClearAll(@Nonnull Run<?,?> build) {
...
+                listener.error(c.getDisplayName() + " seems to be offline");
+                return new LocalLauncher(listener);
+            }
+        } else {
+            listener.error(workspace + " seems to be offline");
+            return new LocalLauncher(listener);
+        }
+    }
+
+    private static @Nonnull EnvVars envFor(@Nonnull FilePath workspace, @Nonnull TaskListener listener) throws IOException, InterruptedException {
```
oleg-nenashev: Not sure about the current implementation. It builds computer environment variables, but ignores possible overrides by the `Run`'s logic. I hope it does not impact artifact managers, but I can imagine `withEnvVars()` closure overriding some variables for the external artifact manager. E.g. just `tool()` installation for the artifact management binary.

Maybe an edge case

jglick: Yes this could be improved I think—will check. Anyway this method is only called when using the deprecated overloads. As of https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/60 the actual `EnvVars` is loaded from the step context, for example taking into account `withEnv` blocks.

src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
```diff
@@ -0,0 +1,238 @@
...
+
+/**
+ * {@link #run} allows an implementation of {@link ArtifactManager} plus {@link VirtualFile} to be run through a standard gantlet of tests.
+ */
+public class ArtifactManagerTest {
+
+    @Rule public JenkinsRule r = new JenkinsRule();
+    @Rule public LoggerRule logging = new LoggerRule();
+
+    public static void run(@Nonnull JenkinsRule r, @CheckForNull ArtifactManagerFactory factory) throws Exception {
```
carlossg: I would split this, at least another for StashAwareArtifactManager, otherwise we only get the first failure

jglick:
> I would split this, at least another for `StashAwareArtifactManager`, otherwise we only get the first failure

Considered it, though it sounds like a bit of work to set up. The standard idiom that I know of for JUnit 4 would be to use `Parameterized` as in https://github.com/jenkinsci/workflow-cps-plugin/pull/139, and I suppose the parameters would be instances of something like

```java
@FunctionalInterface
interface TestArtifactManager {
    void test(ArtifactManagerFactory amf) throws Exception;
}
```

That would still wind up running a lot slower, though, because it would be hard to share setup such as the mock agents. All in all, did not seem worth the bother.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -83,20 +89,35 @@ public static void stash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull
...
+     * @param listener a way to report progress or problems
      * @param includes a set of Ant-style file includes, separated by commas; null/blank is allowed as a synonym for {@code **} (i.e., everything)
      * @param excludes an optional set of Ant-style file excludes
      * @param useDefaultExcludes whether to use Ant default excludes
      * @param allowEmpty whether to allow an empty stash
+     * @see StashAwareArtifactManager#stash
      */
     @SuppressFBWarnings(value="RV_RETURN_VALUE_IGNORED_BAD_PRACTICE", justification="fine if mkdirs returns false")
-    public static void stash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull FilePath workspace, @Nonnull TaskListener listener,
+    public static void stash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull FilePath workspace, @Nonnull Launcher launcher, @Nonnull EnvVars env, @Nonnull TaskListener listener,
```
jglick: @abayer FYI, this is the new overload which could plausibly be combined with a `boolean verbose` option to address [JENKINS-40912](https://issues.jenkins-ci.org/browse/JENKINS-40912) in lieu of #62. A minor API change like that could be done in trunk at low risk, leaving the introduction of `StashAwareArtifactManager` and core baseline changes to this PR for later.

abayer: Not entirely clear on what you mean? I made a comment over at https://issues.jenkins-ci.org/browse/JENKINS-40912 wondering whether building support for listing stash files into `StashAwareArtifactManager` makes sense, since anything we do here like #62 would break horribly the moment the `StashAwareArtifactManager` change came in... I dunno. Everything about JENKINS-40912 makes me want to light myself on fire at this point. =)

jglick: Responded in JIRA.

jglick:
No longer any particular reason to hold this back from my PoV, unless @abayer has made a decision w.r.t. #62 that would require modifications to the signatures of the new `stash` / `unstash` overloads.

(Note that there is still a dependency on merged yet unreleased changes in `workflow-step-api`.)

carlossg:
Agree, this is ready to be merged now that core changes are released

oleg-nenashev:
Deployed as 2.28-20180427.142849-2 to do some testing

src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
```diff
@@ -0,0 +1,395 @@
...
+            StashManager.clearAll(b, listener);
+            // Stashes should have been deleted, but not artifacts:
+            assertTrue(b.getArtifactManager().root().child("file").isFile());
+            ws.deleteContents();
+            assertFalse(ws.child("file").exists());
+            f.apply(p, b, ws, launcher, env, listener);
+        }
+    }
+
+    private static class Verify {
```
svanoort: This usage feels... very nonintuitive and hard to follow.  Please, could we do something to clean it up?

src/test/java/org/jenkinsci/plugins/workflow/DirectArtifactManagerFactory.java
```diff
@@ -0,0 +1,209 @@
...
+import org.apache.commons.io.FileUtils;
+import org.jvnet.hudson.test.JenkinsRule;
+
+/**
+ * Exercises direct download of artifacts.
+ * Whereas {@link ArtifactManagerTest} allows you to test an implementation, this allows you to test a caller.
+ * Use {@link #whileBlockingOpen} to exercise the behavior.
+ * @see <a href="https://issues.jenkins-ci.org/browse/JENKINS-49635">JENKINS-49635</a>
+ */
+public final class DirectArtifactManagerFactory extends ArtifactManagerFactory {
```
svanoort: Future note: this would be a good place to inject testing of different failure modes (hangs, IOExceptions etc).

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
jglick: Possible, though six are mandatory and only five optional, and there is only one caller so we are not really simplifying anything that way. Option builders are useful when you have an open-ended list of callers and most of them do not need to specify most of the options.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -228,16 +289,81 @@ public boolean shouldClearAll(@Nonnull Run<?,?> build) {
 
     }
 
-    @Restricted(DoNotUse.class)
+    /**
+     * Mixin interface for an {@link ArtifactManager} which supports specialized stash behavior as well.
+     */
+    @Restricted(Beta.class)
+    public interface StashAwareArtifactManager /* extends ArtifactManager */ {
```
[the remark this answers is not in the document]
jglick: There is only one foreseen implementation of this beta SPI in Essentials: `JCloudsArtifactManager`. This in turn delegates to a small amount of provider-specific code to handle S3, etc.

Nor is baking protections into the API feasible. The specifics of JClouds semantics are interleaved with Remoting boundaries.

Modifying behavior of local artifact storage is out of scope. No one that I know of has complained about such issues with the current implementation, this PR does not touch it, and our policy moving forward will be to recommend external storage anyway.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -228,16 +289,81 @@ public boolean shouldClearAll(@Nonnull Run<?,?> build) {
...
+
+        /** @see StashManager#stash(Run, String, FilePath, Launcher, EnvVars, TaskListener, String, String, boolean, boolean) */
+        void stash(@Nonnull String name, @Nonnull FilePath workspace, @Nonnull Launcher launcher, @Nonnull EnvVars env, @Nonnull TaskListener listener, @CheckForNull String includes, @CheckForNull String excludes, boolean useDefaultExcludes, boolean allowEmpty) throws IOException, InterruptedException;
```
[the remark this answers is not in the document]
jglick: Ditto, only one caller.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -228,16 +289,81 @@ public boolean shouldClearAll(@Nonnull Run<?,?> build) {
...
+
+    }
+
+    private static @CheckForNull StashAwareArtifactManager stashAwareArtifactManager(@Nonnull Run<?, ?> build) throws IOException {
+        ArtifactManager am = build.pickArtifactManager();
+        return am instanceof StashAwareArtifactManager ? (StashAwareArtifactManager) am : null;
+    }
+
+    @Deprecated
+    private static @Nonnull Launcher launcherFor(@Nonnull FilePath workspace, @Nonnull TaskListener listener) {
```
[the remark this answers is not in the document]
jglick: Yes, handled by `Node.createLauncher`. It does not matter since this is dead code unless the user did not also update `workflow-basic-steps`.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -228,16 +289,81 @@ public boolean shouldClearAll(@Nonnull Run<?,?> build) {
...
+                return new LocalLauncher(listener);
+            }
+        } else {
+            listener.error(workspace + " seems to be offline");
+            return new LocalLauncher(listener);
+        }
+    }
+
+    @Deprecated
+    private static @Nonnull EnvVars envFor(@Nonnull Run<?, ?> build, @Nonnull FilePath workspace, @Nonnull TaskListener listener) throws IOException, InterruptedException {
```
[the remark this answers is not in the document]
jglick: See above, this is dead code for binary compatibility.

src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
jglick: It is somewhat hard to follow, but I do not see any particular simplification.

src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
```diff
@@ -0,0 +1,395 @@
...
+                } catch (RuntimeException | IOException x) {
+                    throw x;
+                } catch (Exception x) {
+                    throw new IOException(x);
+                }
+                return null;
+            }
+        }
+
+        private void test() throws Exception {
```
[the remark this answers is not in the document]
jglick: ack

src/test/java/org/jenkinsci/plugins/workflow/DirectArtifactManagerFactory.java
jglick: All these are tested in `artifact-manager-s3`. This API does not define anything about the service handling the artifacts, so it is logically impossible to have such tests here.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
svanoort: I'll let it go for now -- you're right that the set of callers is currently pretty limited.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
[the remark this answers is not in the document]
svanoort: 👍

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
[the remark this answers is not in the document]
svanoort: @jglick  Some points:

1. Whatever API you're expecting users to extend when writing new External Storage implementations needs to include hooks for JENKINS-50597 protections.  As we've already discussed a "one size fits all" solution isn't appropriate for handling network failures, and handling network failures is non-optional. We already know from durable-task-step and remoting failures that network failures come in many flavors, are common, and if we do not plan for them they can cause catastrophic stability problems.

2. If **this** isn't the API users are expected to implement/extend for external storage implementations, what is?  How do we ensure plugin authors base off that rather than this API? (Restrictions? Documentation?)  How do we ensure failures are appropriately propagated to the top level? I think IOException subtypes are generally appropriate for purposes of keeping API compatibility on current APIs, but would like the optional ability to inject a handler where practical.

> There is only one foreseen implementation of this beta SPI

This assumption raises a red flag, because if feels like we're getting tunnel vision about shipping one implementation and setting up for future problems.  I can totally understand the desire to just *release*, but if you want external storage to be broadly used, the APIs for it should be built for the future uses.

If you only want one API consumer for **this** API, it should be locked down as much as possible to prevent others and at the very least documented that "you should extend this other thing here, extending this is dangerous".  Otherwise the API should be built in a way that is appropriate to handle extension. 

Now, if the expected behavior is that plugin authors will *always* extend off JCloud APIs for external storage implementations, I have no problem with making that the case, but again we need to make that clear and do what we can to keep this locked-down -- and ensure again that failures are propagated.

> Nor is baking protections into the API feasible. The specifics of JClouds semantics are interleaved with Remoting boundaries.

That is going to need a proper explanation, I'm afraid -- on the surface it smells to me like we're letting an implementation detail dictate API design rather than trying to build for robustness and then figure out implementation.

> No one that I know of has complained about such issues with the current implementation

You mean, aside from general broken-ness with IO operations when an NFS server gets overloaded?  I don't think anybody is saying that's a hard requirement in this work, just that if our protections extend to that as a side-effect it won't be a problem and in fact may be beneficial.

> our policy moving forward will be to recommend external storage anyway

Which will require a rich ecosystem of implementations for different technologies, given how many different Jenkins deployment architectures there are.  Which again means that we need to build those base APIs to be reusable and safe for extension.

src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
```diff
@@ -208,20 +214,32 @@ private static void setUpWorkspace(FilePath workspace, boolean weirdCharacters)
...
+     */
     @FunctionalInterface
-    interface TestStashFunction {
+    private interface TestStashFunction {
         void apply(FreeStyleProject p, FreeStyleBuild b, FilePath ws, Launcher launcher, EnvVars env,
                 TaskListener listener) throws Exception;
     }
 
+    /**
+     * Verfies behaviors of stash and unstash operations.
```
svanoort: 🐜  Verifies, not verfies

src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
```diff
@@ -0,0 +1,438 @@
...
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.LoggerRule;
+
+/**
+ * {@link #artifactArchiveAndDelete} and variants allow an implementation of {@link ArtifactManager} plus {@link VirtualFile} to be run through a standard gantlet of tests.
+ */
+public class ArtifactManagerTest {
```
svanoort: 👍 The Javadocs make this a fair bit clearer.

src/test/java/org/jenkinsci/plugins/workflow/DirectArtifactManagerFactory.java
```diff
@@ -58,10 +70,44 @@
  */
 public final class DirectArtifactManagerFactory extends ArtifactManagerFactory {
 
+    private static final Logger LOGGER = Logger.getLogger(DirectArtifactManagerFactory.class.getName());
     private static final AtomicInteger blockOpen = new AtomicInteger();
 
+    private final transient URL baseURL;
+
+    public DirectArtifactManagerFactory() throws Exception {
+        HttpServer server = ServerBootstrap.bootstrap().
```
svanoort: This is a pretty nifty change.   I dig it.

src/test/java/org/jenkinsci/plugins/workflow/DirectArtifactManagerFactory.java
```diff
@@ -0,0 +1,257 @@
...
+import org.apache.http.HttpRequest;
+import org.apache.http.HttpResponse;
+import org.apache.http.entity.StringEntity;
+import org.apache.http.impl.bootstrap.HttpServer;
+import org.apache.http.impl.bootstrap.ServerBootstrap;
+import org.apache.http.protocol.HttpContext;
+import org.jvnet.hudson.test.JenkinsRule;
+
+/**
+ * Exercises direct download of artifacts.
```
svanoort: Might be better phrased as "Provides a minimal real implementation of an ArtifactManager which supports URL-based operations and has HTTP download." then all the other stuff -- helps clarify what it is as well as how you use it.

src/test/java/org/jenkinsci/plugins/workflow/DirectArtifactManagerFactory.java
jglick: FYI this was necessary because `RobustHTTPClient` works only on, well, HTTP(S) URLs. I have already clarified the core Javadoc to explain that external URLs must in fact be HTTP(S) (other protocols would hardly be sensible), so in order to _test_ handling of them, we need to be running a real HTTP server.

src/test/java/org/jenkinsci/plugins/workflow/DirectArtifactManagerFactory.java
jglick: ack

jglick:
Moving a long thread to a top-level comment, since otherwise it is liable to suddenly disappear behind a diff wall.

> Whatever API you're expecting users to extend when writing new External Storage implementations needs to include hooks for JENKINS-50597 protections.

No, it does not. The “hook” is that when you provide an external artifact manager, you get to decide how `stash` and `unstash` steps should work instead of the traditional implementation. Everything about the quality of the replacement implementation is a topic for different layers.

> a "one size fits all" solution isn't appropriate for handling network failures

Agreed.

> handling network failures is non-optional

Agreed, at least for plugins we include in Essentials.

> If this isn't the API users are expected to implement/extend for external storage implementations, what is?

See details in `artifact-manager-s3`.

> How do we ensure plugin authors base off that rather than this API? (Restrictions? Documentation?)

Documentation is fine for me.

> How do we ensure failures are appropriately propagated to the top level?

Via a number of mechanisms which you can see in `artifact-manager-s3`.

> I think IOException subtypes are generally appropriate for purposes of keeping API compatibility on current APIs, but would like the optional ability to inject a handler where practical.

Test coverage indicates no need for any such “handler”, whatever that would be.

>> There is only one foreseen implementation of this beta SPI
> 
> This assumption raises a red flag, because if feels like we're getting tunnel vision about shipping one implementation and setting up for future problems. I can totally understand the desire to just release, but if you want external storage to be broadly used, the APIs for it should be built for the future uses.

The future uses we can _foresee_ at this point are things like Azure storage. Our current plan is for those to be based on jclouds as well.

> If you only want one API consumer for this API, it should be locked down as much as possible to prevent others

There is not much to lock down. It is merely a way to swap out the stock implementation of `stash`/`unstash`.

> at the very least documented that "you should extend this other thing here, extending this is dangerous".

That can be done.

>> Nor is baking protections into the API feasible. The specifics of JClouds semantics are interleaved with Remoting boundaries.
> 
> That is going to need a proper explanation

JEP-202 explains how the system works.

> we're letting an implementation detail dictate API design

To the contrary, here we are avoiding any unnecessary API design in `workflow-api`: it is kept to the bare minimum necessary for the `stash` and `unstash` steps to operate.

jglick:
W.r.t. “locking down” APIs, this is exactly what the beta mark is for—we do not advertise the API as something to be used by unknown external parties without first consulting with us. If the design turns out to be wrong after broader experience, then we change it.

svanoort:
Discussed with @jglick today out-of-band to try to bridge differences here and move things forward.

In general, the assumptions seems to be that:

1. We will only ever care about storing external artifacts in cloud blob stores (AWS S3 and similar implementations in Azure, Google Cloud, etc) that are supported via the JClouds library. 

2. If community members decide to implement their own external storage solutions on top of the new API here, we are not responsible for any harm they do to Jenkins or the Pipeline ecosystem. (In practice, I doubt this will hold up -- there's a history of Pipeline specifically being blamed for issues caused entirely by third-party plugins.  I can cite several recent examples, but will have to do so privately due to confidentiality concerns.)

As far as handling of network failures per [JENKINS-50597](https://issues.jenkins-ci.org/browse/JENKINS-50597) I am happy to report that the [RobustHttpClient](https://github.com/jenkinsci/apache-httpcomponents-client-4-api-plugin/blob/master/src/main/java/io/jenkins/plugins/httpclient/RobustHTTPClient.java) provides *some* of the protections we need -- and might be retrofitted to provide others such as dead-connection timeouts as opposed to fixed-length timeouts, etc.  The implementation in the artifact-manager-s3 plugin uses these for the bulk upload/download of artifacts/stashes.  JClouds third-party APIs are doing most of the work under the hood for metadata calls -- and they use their own distinct network error/retry handling -- this appears to be mostly treated as a black box, but at least provides some protections.

This question has been answered:

> If this isn't the API users are expected to implement/extend for external storage implementations, what is? 

The answer appears to be the [BlobStoreProvider](https://github.com/jenkinsci/artifact-manager-s3-plugin/blob/master/src/main/java/io/jenkins/plugins/artifact_manager_jclouds/BlobStoreProvider.java) -- which is good in principle, but unfortunately that API is very tightly coupled to JClouds so it isn't usable for more general external storage.  Additionally AFAICT there is no documentation indicating that plugin authors should use that extension point for new external storage implementations. 

I raised a concern that this set of APIs is so tightly coupled to JClouds APIs that it cannot be used for artifact repositories such as Artifactory and Nexus -- even though they were in fact [mentioned in the original JEP-202 and later removed](https://github.com/jenkinsci/jep/pull/117/commits/54c4f056804240f557a2aeec41da38dcf8c137f9#diff-4d18c8dccc6884e2f55b619f41cf0797L129).

The reason given for why we do not ever need to care about these artifact repositories is that they "would need their own logic above and beyond cloud BlobStores".  Although I tried to understand what this was and what ensured we would not want to support them with the same base APIs for external artifact storage, I am sorry to say that no satisfactory answer emerged so we moved on. 

Give that answer, I'm still concerned that this concern raised previously still applies:

> This assumption raises a red flag, because if feels like we're getting tunnel vision about shipping one implementation and setting up for future problems. I can totally understand the desire to just release, but if you want external storage to be broadly used, the APIs for it should be built for the future uses.

I have proposed a compromise, which is to define one additional level of `ExtensionPoint` above the [JCloudsArtifactManager](https://github.com/jenkinsci/artifact-manager-s3-plugin/blob/master/src/main/java/io/jenkins/plugins/artifact_manager_jclouds/JCloudsArtifactManager.java) that plugin authors can extend. This would extend `StashAwareArtifactManager` and provide a minimal set of logic for failure detection and handling - probably hooks to wrap an HTTP call with `RobustHttpClient`, maybe a top-level timeout and retry for operations. `JCloudsArtifactManager` would extend that with the JCloud-specific components.  

This effectively "bakes in" general-purpose network protections and ensures the community can easily provide implementations for external artifact storage, while also allowing us to add extra protections within that extension point as needed.  This is a pattern that has worked well for Pipeline in general.  It's an API strategy encapsulating a piece of functionality including the appropriate error handling associated, rather than relying on a completely robust implementation in every case (and makes those implementations simpler as a result).  One example of this strategy paying off: [JENKINS-47758](https://issues.jenkins-ci.org/browse/JENKINS-47758), where we told all Groovy script users to use Script Security for protection, and could then leverage it to catch and protect against memory leaks.

In fact, I've proposed simply making that proposed extension point trivial for now but clearly documenting it as "you must extend this for safety reasons."  Simply *having* that level of indirection allows us to add robustness features later as needed.

Unfortunately that proposal was flatly rejected without an alternative besides accepting this *in entirety in an as-is state.*

I have clearly voiced my concerns here about the way this coupling into Pipeline is being done.  **I think this *will* set us up for robustness and reliability problems in the face of network misbehavior, especially when community members see the implementation and try to roll their own storage implementation without the benefits of a robust-by-design set of base APIs.**   

Unfortunately nobody else appears to share those concerns -- including others such as @oleg-nenashev and @jglick who have suffered from trying to resolve failures traceable to bad assumptions about network robustness.  So, in the interests of not hard-blocking work, I'll grudgingly accept this as merge-able -- once the "you should extend this other thing here, extending this is dangerous" expectation is clearly spelled out in Javadocs and documentation - both in this PR (warning: "implement `StashAwareArtifactManager` at your own risk if you use network storage") and in the downstream implementation.

CC @vivek since he's watching this PR.

jglick:
We actually went over most of these topics in detail, but to summarize:

> that API is very tightly coupled to JClouds so it isn't usable for more general external storage

jclouds has its own extension SPI, certainly more mature than anything we could invent.

> The reason given for why we do not ever need to care about these artifact repositories is that they "would need their own logic above and beyond cloud BlobStores". Although I tried to understand what this was

Repository managers are services deliberately and explicitly incorporated into the CI/CD workflow. For example, a Maven-based project in an organization intending to use Nexus or Artifactory for storage of “artifacts” would have been configured (in whatever settings file) to deploy and/or fetch those artifacts via `mvn` commands. If you are doing that, then you do not need nor want `archiveArtifacts`, `copyartifact`, etc.; the identification of artifacts would be for example a GAV rather than some ID/URL made up by Jenkins. If you are not doing that, and do not really care where artifacts are stored—you just want to run generic upload/download steps copy-pasted from documentation and be assured that big files are anywhere except on `$JENKINS_HOME`—these services would be both overkill (complex to manage) and inefficient; you would be better off with a fast generic blob store.

> define one additional level of `ExtensionPoint`

This would be an API which no extant codes actually calls, except perhaps `artifact-manager-s3` itself, and so YAGNI. The entire point of marking newly introduced APIs with the beta annotation is so that we can ship working features without needing this kind of discussion of hypotheticals, do development in `master`, and refactor later if and when our initial assumptions are proven wrong.

As of https://github.com/jenkinsci/jep/pull/117 plus recent edits here, Javadoc and JEP both already link to recommended implementation strategies, and all relevant APIs are marked beta. Anyone who deliberately bypasses beta restrictions in a new plugin, does not consult with the authors of the APIs in question, _and_ offers a poor-quality implementation is beyond our help. There is no reason to believe they would choose to implement any other intermediate API we offer anyway.

At any rate, none of this discussion has any direct bearing on _this_ PR; it is just general comments on JEP-202. This PR merely provides the hooks by which that portion of JEP-202 limited to Pipeline projects can operate—for historical reasons `unarchive`, `stash`, and `unstash` are defined in `workflow-*` plugins, while `archiveArtifacts` is defined in core and `copyArtifacts` in another plugin already merged and released.

vivek:
>As of jenkinsci/jep#117 plus recent edits here, Javadoc and JEP both already link to recommended implementation strategies, and all relevant APIs are marked beta. Anyone who deliberately bypasses beta restrictions in a new plugin, does not consult with the authors of the APIs in question, and offers a poor-quality implementation is beyond our help. There is no reason to believe they would choose to implement any other intermediate API we offer anyway.

I think we can move forward with above consideration. @svanoort this sounds good to you? Lets get moving on this PR :)

svanoort:
@vivek  I'm just waiting on some docs from @jglick as promised somewhere way above and then yeah, I'll do the dance to set up the master/stable branches, merge this to master, and set the merge target as stable

vivek:
Thanks @svanoort!

carlossg:
> We will only ever care about storing external artifacts in cloud blob stores (AWS S3 and similar implementations in Azure, Google Cloud, etc) that are supported via the JClouds library.

That is not accurate, implementors may choose their own libraries. JClouds makes it easier but other libraries may be more optimized, so it's up to future implementors to choose, but that's not the API decision

> If community members decide to implement their own external storage solutions on top of the new API here, we are not responsible for any harm they do to Jenkins or the Pipeline ecosystem

correct, we can't predict the future

> I raised a concern that this set of APIs is so tightly coupled to JClouds APIs that it cannot be used for artifact repositories such as Artifactory and Nexus

I don't think it is coupled to JClouds but to blobstores. I'd rather not over engineer a solution for artifact repositories until there is any interest from those parties, and they might as well ignore this as using `mvn deploy` or such would make it unnecessary 

> I have proposed a compromise, which is to define one additional level of ExtensionPoint above the JCloudsArtifactManager that plugin authors can extend. This would extend StashAwareArtifactManager and provide a minimal set of logic for failure detection and handling - probably hooks to wrap an HTTP call with RobustHttpClient, maybe a top-level timeout and retry for operations.

I don't agree that retry logic should live in pipeline. This approach would assume that other cloud libraries don't already have some retry logic or that use a specific http client, which may not hold true. If we see other implementations copy that logic then we can move it up or sideways to a common utilities plugin 

> when community members see the implementation and try to roll their own storage implementation

again I don't think we should over engineer it, the first implementation complies with requirements, there is no timeframe for new implementations and this is considered Beta to bring other interested parties

> The entire point of marking newly introduced APIs with the beta annotation is so that we can ship working features without needing this kind of discussion of hypotheticals, do development in master, and refactor later if and when our initial assumptions are proven wrong.

agreed

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -291,8 +291,12 @@ public boolean shouldClearAll(@Nonnull Run<?,?> build) {
 
     /**
      * Mixin interface for an {@link ArtifactManager} which supports specialized stash behavior as well.
-     * The recommended standard implementation is in the plugin currently named {@code artifact-manager-s3},
-     * which in turn supports extensibility to various cloud providers.
+     * <p>The recommended standard implementation is {@code JCloudsArtifactManager}
```
svanoort: @jglick This is *still* not what we had discussed.

svanoort:
@vivek Still blocked here due to not making the requested and trivial change.  I noted this would be ready to merge:

> once the "you should extend this other thing here, extending this is dangerous" expectation is clearly spelled out in Javadocs and documentation - both in this PR (warning: "implement StashAwareArtifactManager at your own risk if you use network storage") and in the downstream implementation.

Instead a few trivial rewordings were done.  

I've set a very low bar here for merging what I consider to be a risky change and yet still someone felt they did not need to meet it?

vivek:
@svanoort PTAL, I think its good to go.

PR #68, opened 2018-04-11
Add NotStoredReason to act as placeholder for unserializable ArgumentsAction arguments

svanoort:
Part of solution for [JENKINS-50752](https://issues.jenkins-ci.org/browse/JENKINS-50752) which is causing problems in the wild. 

Used downstream in https://github.com/jenkinsci/workflow-cps-plugin/pull/220

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -136,7 +136,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.7</version>
+            <version>1.14</version>
         </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -289,9 +292,9 @@ public boolean isUnmodifiedArguments() {
             StepDescriptor d = ((StepNode) n).getDescriptor();
             if (d != null) {
                 try {
-                    return resolve(new DescribableModel<>(d.clazz), args).getArguments();
+                    return resolve(DescribableModel.of(d.clazz), args).getArguments();
```
svanoort: Just a little cleanup to use the new API to avoid doing needless allocations.

pom.xml
```diff
@@ -136,7 +136,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.7</version>
+            <version>1.14</version>
```
svanoort: Picked up for the optimized DescribableModel.of method

PR #69, opened 2018-04-18
MessageFormat treats ' as a special character  …

jsoref:
https://docs.oracle.com/javase/8/docs/api/java/text/MessageFormat.html

Test case is jenkins/configure
 Expand "Pipeline Default Speed/Durability Level"
 Help for "Performance-optimized: much faster (requires clean shutdown to save running pipelines)"

Without this fix, you will see:
... but running pipelines lose step information and cant resume if Jenkins unexpectedly fails.

You should see:
... but running pipelines lose step information and can't resume if Jenkins unexpectedly fails.

This is equivalent to jenkinsci/jenkins#3203 but for this repo...


svanoort:
Good catch.

PR #70, opened 2018-04-25
avoid printing too long argument lists during FlowNode visualization

ArturHarasimiuk:
In some cases it is possible that node argument will be impossible to
wrap properly and is displayed as single line. In this case pipeline
steps view becomes hard to read because right most columns are moved
beyond the screen. this requires scrolling which is not so convenient.
Also status colum is outside of screen.

This change cuts arguments list to 80 characters which should be fine
for most of usages.


ArturHarasimiuk:
thanks for comment. I created new PR: https://github.com/jenkinsci/workflow-support-plugin/pull/71

PR #71, opened 2018-05-25
Split tests so implementations can choose which ones to call

carlossg:
Subsumes #67.

(originally https://github.com/jglick/workflow-api-plugin/pull/2)

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
@@ -90,11 +90,9 @@ public class ArtifactManagerTest {
             return null;
         }
     }
-
-    /**
-     * @param image use {@link #prepareImage} in a {@link BeforeClass} block
-     */
-    public static void run(@Nonnull JenkinsRule r, @CheckForNull ArtifactManagerFactory factory, boolean weirdCharacters, @CheckForNull DockerImage image) throws Exception {
+    
+    private static void wrapInContainer(@Nonnull JenkinsRule r, @CheckForNull ArtifactManagerFactory factory,
+            boolean weirdCharacters, TestFunction f) throws Exception {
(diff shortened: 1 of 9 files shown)


jglick:
Presumably https://github.com/jenkinsci/compress-artifacts-plugin/pull/7 would need to be amended to call (just) `artifactArchiveAndDelete`.

src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
```diff
@@ -117,52 +115,77 @@ public static void run(@Nonnull JenkinsRule r, @CheckForNull ArtifactManagerFact
...
+                runningContainer.close();
+            }
+        }
+    }
+
+    /**
+     * @param image use {@link #prepareImage} in a {@link BeforeClass} block
+     */
+    public static void artifactArchive(@Nonnull JenkinsRule r, @CheckForNull ArtifactManagerFactory factory, boolean weirdCharacters, @CheckForNull DockerImage image) throws Exception {
+        wrapInContainer(r, factory, weirdCharacters, new TestFunction() {
```
jglick: use a lambda

src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
```diff
@@ -117,52 +115,77 @@ public static void run(@Nonnull JenkinsRule r, @CheckForNull ArtifactManagerFact
...
-                root = b2.getArtifactManager().root();
+            f.apply(agent, p, b, ws);
+        } finally {
+            if (runningContainer != null) {
+                runningContainer.close();
+            }
+        }
+    }
+
+    /**
```
jglick: Need explanation of Javadoc of the differences between these methods.

jglick:
Folded into #67, after fixing a test failure.

PR #77, opened 2018-09-11
[JENKINS-44193] wfapi/describe execNode empty

ArturHarasimiuk:
Adding ExecutorAction which will be added to flowNode actions and
later exposed as execNode by wfapi.


jglick:
I think this is covered by https://github.com/jenkinsci/workflow-api-plugin/blob/61c3180fa03f758198ee97b678ea8adf3cda9818/src/main/java/org/jenkinsci/plugins/workflow/actions/WorkspaceAction.java#L41-L42

jglick:
Not needed.

PR #78, opened 2018-09-14
use ABORTED icon color for interrupted steps

ArturHarasimiuk:
parallel() aborts branch execution when failFast == true and another
branch failed. In this case RED color is used in 'Pipeline Steps' view.
It makes more difficult to find real reason of failure because some of
red dots are misleading.

With this change we use BallColor.ABORTED when FlowInterruptedException
is the reason of step failure. Thanks to this user can easier find root
cause of build failure.


svanoort:
@ArturHarasimiuk Thanks for the contribution, this one is solid work!

PR #81, opened 2018-10-19
[JENKINS-54073] Buffering for FileLogStorage

jglick:
~Checking whether this addresses [JENKINS-54073](https://issues.jenkins-ci.org/browse/JENKINS-54073). It does not seem to make any positive difference in little local tests but I doubt those are realistic anyway.~ Note that as far as I can tell, freestyle builds [do not buffer](https://github.com/jenkinsci/jenkins/blob/b057abc7b84295969b894d06011e250842c6d09c/core/src/main/java/hudson/model/Run.java#L1887-L1889), and neither did [pre-JEP-210 Pipeline builds](https://github.com/jenkinsci/workflow-job-plugin/blob/workflow-job-2.21/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L227-L228).

----

I think I finally tracked down the issue. While this PR still buffers writes to the `log` on master, the main change is ensuring that the `PrintStream` made available remotely (most significantly, to `DurableTaskStep`) refrains from sending every little chunk of output one by one, and also refrains from sending gratuitous `ProxyOutputStream.Flush` packets. More: https://github.com/jenkinsci/workflow-api-plugin/pull/81#discussion_r228196101

Freestyle builds in fact _do_ send little bits of output. I guess nobody cared. After all, it was harder to get lots of agents spewing log output in parallel.

----

The main open question for myself here is the expectation that remote code calls the equivalent of `listener.getLogger().flush()` when it is done with a major task (like a step). If it neglects to do so, any remaining content will get auto-flushed, but out of order in the build log and possibly too late (after the agent has been disconnected or the build completed). This is not exactly compatible (thus the need for https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/85), but on the other hand I cannot think of many cases where code other than `DurableTaskStep` was printing content remotely to begin with—most plugins which print messages do so inside `perform` or similar, on the master.

A possible refinement would be to turn autoflush back on for the remote `PrintStream`, so that `println(String)` and the like are sent immediately, but then modify `DecoratedTaskListener` to extract the underlying `OutputStream` from the `PrintStream delegate.getLogger()`. Thus there would only be one `PrintStream` on top of the stack of delegates, so any calls to it directly would flush, yet `DurableTaskStep.HandlerImpl` would continue to write to the buffered, non-autoflushing `OutputStream` stack, which would be ideal. The problem with this scenario is that decorators then lose access to the `PrintStream` that they sometimes need for synchronization purposes to avoid mangled output under high concurrency, as in https://github.com/jenkinsci/timestamper-plugin/commit/bf1e61c8a57ef434316a41ffbfb4126e88a9f1f9 in https://github.com/jenkinsci/timestamper-plugin/pull/25. I am not sure yet if there is a good resolution to this conflict; if there is, it should be possible to do it compatibly later (since the worst that could happen is that something like https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/85 gratuitously sends one flush packet at the end of a step).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
@@ -67,13 +67,14 @@ final class BufferedBuildListener implements BuildListener, Closeable, Serializa
         private static final long serialVersionUID = 1;
 
         private final RemoteOutputStream ros;
+        private final DelayBufferedOutputStream.Tuning tuning = DelayBufferedOutputStream.Tuning.DEFAULT; // load defaults on master
 
         Replacement(BufferedBuildListener cbl) {
             this.ros = new RemoteOutputStream(new CloseProofOutputStream(cbl.out));
         }
 
         private Object readResolve() throws IOException {
-            return new BufferedBuildListener(new DelayBufferedOutputStream(ros));
+            return new BufferedBuildListener(new DelayBufferedOutputStream(ros, tuning));
         }
 
     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
@@ -97,7 +97,7 @@ final class BufferedBuildListener implements BuildListener, Closeable, Serializa
         private long recurrencePeriod = MIN_RECURRENCE_PERIOD;
 
         DelayBufferedOutputStream(OutputStream out) {
-            super(new FlushProofOutputStream(out)); // default buffer size: 8Kib
+            super(new FlushControlledOutputStream(out)); // default buffer size: 8Kib
             reschedule();
         }
 
@@ -108,13 +108,13 @@ final class BufferedBuildListener implements BuildListener, Closeable, Serializa
 
         /** We can only call {@link BufferedOutputStream#flushBuffer} via {@link #flush}, but we do not wish to flush the underlying stream, only write out the buffer. */
         private void flushBuffer() throws IOException {
-            ThreadLocal<Boolean> fauxFlushing = ((FlushProofOutputStream) out).fauxFlushing;
-            boolean orig = fauxFlushing.get();
-            fauxFlushing.set(true);
+            ThreadLocal<Boolean> enableFlush = ((FlushControlledOutputStream) out).enableFlush;
+            boolean orig = enableFlush.get();
+            enableFlush.set(false);
             try {
                 flush();
             } finally {
-                fauxFlushing.set(orig);
+                enableFlush.set(orig);
             }
         }
         
@@ -151,15 +151,16 @@ final class BufferedBuildListener implements BuildListener, Closeable, Serializa
         
     }
 
-    private static final class FlushProofOutputStream extends FilterOutputStream {
+    /** @see DelayBufferedOutputStream#flushBuffer */
+    private static final class FlushControlledOutputStream extends FilterOutputStream {
 
-        private final ThreadLocal<Boolean> fauxFlushing = new ThreadLocal<Boolean>() {
+        private final ThreadLocal<Boolean> enableFlush = new ThreadLocal<Boolean>() {
             @Override protected Boolean initialValue() {
-                return false;
+                return true;
             }
         };
 
-        FlushProofOutputStream(OutputStream out) {
+        FlushControlledOutputStream(OutputStream out) {
             super(out);
         }
 
@@ -168,7 +169,7 @@ final class BufferedBuildListener implements BuildListener, Closeable, Serializa
         }
         
         @Override public void flush() throws IOException {
-            if (!fauxFlushing.get()) {
+            if (enableFlush.get()) {
                 super.flush();
             }
         }
--- a/src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
@@ -1,0 +1,176 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2018 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.jenkinsci.plugins.workflow.log;
+
+import hudson.model.BuildListener;
+import hudson.remoting.RemoteOutputStream;
+import hudson.util.StreamTaskListener;
+import java.io.BufferedOutputStream;
+import java.io.Closeable;
+import java.io.FilterOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.PrintStream;
+import java.lang.ref.Reference;
+import java.lang.ref.WeakReference;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import jenkins.util.Timer;
+import org.jenkinsci.remoting.SerializableOnlyOverRemoting;
+
+/**
+ * Unlike {@link StreamTaskListener} this does not set {@code autoflush} on the reconstructed {@link PrintStream}.
+ * It also implements buffering on output, flushed (without ProxyOutputStream.Flush) at intervals.
+ */
+final class BufferedBuildListener implements BuildListener, Closeable, SerializableOnlyOverRemoting {
+
+    private static final Logger LOGGER = Logger.getLogger(BufferedBuildListener.class.getName());
+
+    private final OutputStream out;
+    private final PrintStream ps;
+
+    BufferedBuildListener(OutputStream out) throws IOException {
+        this.out = out;
+        ps = new PrintStream(out, false, "UTF-8");
+    }
+    
+    @Override public PrintStream getLogger() {
+        return ps;
+    }
+    
+    @Override public void close() throws IOException {
+        ps.close();
+    }
+
+    private Object writeReplace() {
+        return new Replacement(this);
+    }
+
+    private static final class Replacement implements SerializableOnlyOverRemoting {
+
+        private static final long serialVersionUID = 1;
+
+        private final RemoteOutputStream ros;
+
+        Replacement(BufferedBuildListener cbl) {
+            this.ros = new RemoteOutputStream(cbl.out);
+        }
+
+        private Object readResolve() throws IOException {
+            return new BufferedBuildListener(new DelayBufferedOutputStream(ros));
+        }
+
+    }
+
+    private static final class DelayBufferedOutputStream extends BufferedOutputStream {
+
+        // TODO make these customizable (not trivial since this system properties would need to be loaded on the master side and then remoted)
+        private static final long MIN_RECURRENCE_PERIOD = 250; // ¼s
+        private static final long MAX_RECURRENCE_PERIOD = 10_000; // 10s
+        private static final float RECURRENCE_PERIOD_BACKOFF = 1.05f;
+
+        private long recurrencePeriod = MIN_RECURRENCE_PERIOD;
+
+        DelayBufferedOutputStream(OutputStream out) {
+            super(new FlushProofOutputStream(out)); // default buffer size: 8Kib
+            reschedule();
+        }
+
+        private void reschedule() {
+            Timer.get().schedule(new Flush(this), recurrencePeriod, TimeUnit.MILLISECONDS);
+            recurrencePeriod = Math.min((long) (recurrencePeriod * RECURRENCE_PERIOD_BACKOFF), MAX_RECURRENCE_PERIOD);
+        }
+
+        /** We can only call {@link BufferedOutputStream#flushBuffer} via {@link #flush}, but we do not wish to flush the underlying stream, only write out the buffer. */
+        private void flushBuffer() throws IOException {
+            ThreadLocal<Boolean> fauxFlushing = ((FlushProofOutputStream) out).fauxFlushing;
+            boolean orig = fauxFlushing.get();
+            fauxFlushing.set(true);
+            try {
+                flush();
+            } finally {
+                fauxFlushing.set(orig);
+            }
+        }
+        
+        @Override public void close() throws IOException {
+            // Ignored. We do not allow the stream to be closed from the remote side.
+        }
+        
+        void run() {
+            try {
+                flushBuffer();
+            } catch (IOException x) {
+                LOGGER.log(Level.FINE, null, x);
+            }
+            reschedule();
+        }
+        
+    }
+    
+    private static final class Flush implements Runnable {
+
+        /** Since there is no explicit close event, just keep flushing periodically until the stream is collected. */
+        private final Reference<DelayBufferedOutputStream> osr;
+        
+        Flush(DelayBufferedOutputStream os) {
+            osr = new WeakReference<>(os);
+        }
+        
+        @Override public void run() {
+            DelayBufferedOutputStream os = osr.get();
+            if (os != null) {
+                os.run();
+            }
+        }
+        
+    }
+
+    private static final class FlushProofOutputStream extends FilterOutputStream {
+
+        private final ThreadLocal<Boolean> fauxFlushing = new ThreadLocal<Boolean>() {
+            @Override protected Boolean initialValue() {
+                return false;
+            }
+        };
+
+        FlushProofOutputStream(OutputStream out) {
+            super(out);
+        }
+
+        @Override public void write(byte[] b, int off, int len) throws IOException {
+            out.write(b, off, len); // super method writes one byte at a time!
+        }
+        
+        @Override public void flush() throws IOException {
+            if (!fauxFlushing.get()) {
+                super.flush();
+            }
+        }
+
+    }
+
+}
--- a/src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
@@ -25,30 +25,21 @@
 package org.jenkinsci.plugins.workflow.log;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import hudson.CloseProofOutputStream;
 import hudson.model.BuildListener;
 import hudson.remoting.RemoteOutputStream;
 import hudson.util.StreamTaskListener;
-import java.io.BufferedOutputStream;
 import java.io.Closeable;
-import java.io.FilterOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.PrintStream;
-import java.lang.ref.Reference;
-import java.lang.ref.WeakReference;
-import java.util.concurrent.TimeUnit;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-import jenkins.util.Timer;
 import org.jenkinsci.remoting.SerializableOnlyOverRemoting;
 
 /**
  * Unlike {@link StreamTaskListener} this does not set {@code autoflush} on the reconstructed {@link PrintStream}.
- * It also implements buffering on output, flushed (without ProxyOutputStream.Flush) at intervals.
+ * It also wraps on the remote side in {@link DelayBufferedOutputStream}.
  */
 final class BufferedBuildListener implements BuildListener, Closeable, SerializableOnlyOverRemoting {
-
-    private static final Logger LOGGER = Logger.getLogger(BufferedBuildListener.class.getName());
 
     private final OutputStream out;
     @SuppressFBWarnings(value = "SE_BAD_FIELD", justification = "using Replacement anyway, fields here are irrelevant")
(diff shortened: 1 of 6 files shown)


src/test/java/org/jenkinsci/plugins/workflow/log/LogStorageTestBase.java
```diff
@@ -93,6 +93,7 @@
         step1.getLogger().println("one #3");
         overall.getLogger().println("pausing");
         overallHtmlPos = assertOverallLog(overallHtmlPos, "<span class=\"pipeline-node-1\">one #2\none #3\n</span>pausing\n", true);
+        // TODO if we produce output from the middle of a step, we need new span blocks
```
jglick: An unrelated bug I noticed: the **hide** links do not work reliably for running builds.

src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
```diff
@@ -0,0 +1,176 @@
...
+            this.ros = new RemoteOutputStream(cbl.out);
+        }
+
+        private Object readResolve() throws IOException {
+            return new BufferedBuildListener(new DelayBufferedOutputStream(ros));
+        }
+
+    }
+
+    private static final class DelayBufferedOutputStream extends BufferedOutputStream {
```
jglick: The key part of the fix: buffering.

src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
```diff
@@ -0,0 +1,176 @@
...
+final class BufferedBuildListener implements BuildListener, Closeable, SerializableOnlyOverRemoting {
+
+    private static final Logger LOGGER = Logger.getLogger(BufferedBuildListener.class.getName());
+
+    private final OutputStream out;
+    private final PrintStream ps;
+
+    BufferedBuildListener(OutputStream out) throws IOException {
+        this.out = out;
+        ps = new PrintStream(out, false, "UTF-8");
```
jglick: Also important: the `false`. Compare [this](https://github.com/jenkinsci/jenkins/blob/54f4733cd009d98d912b20c0f546db5e24bc3d66/core/src/main/java/hudson/util/StreamTaskListener.java#L172).

src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
```diff
@@ -0,0 +1,178 @@
...
+        private long recurrencePeriod = MIN_RECURRENCE_PERIOD;
+
+        DelayBufferedOutputStream(OutputStream out) {
+            super(new FlushProofOutputStream(out)); // default buffer size: 8Kib
+            reschedule();
+        }
+
+        private void reschedule() {
+            Timer.get().schedule(new Flush(this), recurrencePeriod, TimeUnit.MILLISECONDS);
+            recurrencePeriod = Math.min((long) (recurrencePeriod * RECURRENCE_PERIOD_BACKOFF), MAX_RECURRENCE_PERIOD);
```
dwnusbaum: Why back the recurrencePeriod off over time unconditionally? My initial thought is to back off only if there were no writes to the buffer since the last call to `#run`.

src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
```diff
@@ -0,0 +1,178 @@
...
+        FlushProofOutputStream(OutputStream out) {
+            super(out);
+        }
+
+        @Override public void write(byte[] b, int off, int len) throws IOException {
+            out.write(b, off, len); // super method writes one byte at a time!
+        }
+        
+        @Override public void flush() throws IOException {
+            if (!fauxFlushing.get()) {
```
dwnusbaum: nit: I find the naming kind of confusing here. This looks to me like a `FlushUnlessFlushingIsTemporarilyDisabledOutputStream` :), and I think `fauxFlushing` could be more clearly named `allowFlushing` and its initial value and uses could be inverted.

src/test/java/org/jenkinsci/plugins/workflow/log/LogStorageTestBase.java
jglick: As an aside: I checked actual `progressiveHtml` HTTP responses and they look sane, and I certainly verified this scenario when developing the feature, so I am not sure why it would not be working now. Possibly a client-side CSS issue. Will look into it separately.

src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
jglick: Ack, can work on the naming a bit.

src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
jglick: Oh I forgot to mention: when wrapping in `timestamps {…}`, this works fine because the buffering is applied right before transmitting over Remoting, whereas `FileMonitoringController.Watcher` still runs every 100ms and `hudson.plugins.timestamper.pipeline.GlobalDecorator` is applied to that immediately. So after the process has been running for a few minutes, you will see a 10s pause, then a big chunk of output appear in the log but with every line marked with a timestamp to ~100ms precision. That is per thread; if there are multiple threads running, you see correspondingly faster output.

(`progressiveText.jelly` displays new build log content once per second. Blue Ocean seems to do something similar up to some cutoff; once the content for a given step gets even slightly long, it just seems to give up and stop showing anything. So it is only really interesting to look at this stuff in the classic UI.)

jglick:
> maybe it would help a little bit in some cases

User thread dumps did show some lock contention writing to `log`, presumably because each write was unbuffered. I expect that this issue should be greatly reduced by minimizing the number of times writes to `log` are actually needed, and avoiding excessive interleaving, as the later parts of this PR do. Still, it seems useful to avoid the need for synchronous file writes.

Now that I think about it, since I have already written the `DelayBufferedOutputStream` for use on the remote side, I may as well use it on the master side too. That would reduce the chance and minimize the extent of text loss in the case of a sudden shutdown while a build is running.

src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
dwnusbaum: Ok, that reasoning makes sense to me, and it seems like a good trade-off.

src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
```diff
@@ -0,0 +1,81 @@
...
+        private static final long serialVersionUID = 1;
+
+        private final RemoteOutputStream ros;
+
+        Replacement(BufferedBuildListener cbl) {
+            this.ros = new RemoteOutputStream(new CloseProofOutputStream(cbl.out));
+        }
+
+        private Object readResolve() throws IOException {
+            return new BufferedBuildListener(new DelayBufferedOutputStream(ros));
```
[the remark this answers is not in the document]
jglick: OK, will check this.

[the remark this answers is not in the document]
jglick: Actually it is not that decoration is taking place on the wrong node (you were perhaps confusing an earlier `Running on remote` message, always printed from master, with the `printed a message on master=false` message being asserted). It is just that with buffering, the remotely printed message is not sent to the build log quickly enough; corrected with

```diff
diff --git a/src/test/java/org/jenkinsci/plugins/workflow/job/console/TaskListenerDecoratorTest.java b/src/test/java/org/jenkinsci/plugins/workflow/job/console/TaskListenerDecoratorTest.java
index 2262458..a53b615 100644
--- a/src/test/java/org/jenkinsci/plugins/workflow/job/console/TaskListenerDecoratorTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/job/console/TaskListenerDecoratorTest.java
@@ -200,6 +200,7 @@ public class TaskListenerDecoratorTest {
             }
             @Override public Void call() throws RuntimeException {
                 listener.getLogger().println("printed a message on master=" + JenkinsJVM.isJenkinsJVM());
+                listener.getLogger().flush();
                 return null;
             }
         }
```

The question is whether this is just a test fix that needs to be made, as I did in `LogStorageTestBase`, or whether this is a real problem and we need to somehow force a flush of any remote streams when a step ends. Let me think this over.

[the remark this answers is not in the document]
jglick: And, some `ExportTable.diagnoseInvalidObjectId` warnings. Remoting :man_shrugging:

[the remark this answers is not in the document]
jglick: > whether it is practical for the master to reach out to agents at the end of a step and request flushes

I cannot think of any reasonable way to do that. You would need to have the agents `Channel.export` some flushing thunk which the `BufferedBuildListener` would collect so that its `close` could call them all. But at that point the channel might not even be alive, or remote calls could hang, and there could be complicated garbage collection issues here, etc.

I think it is more straightforward to just ask the step, or whatever is working with a `TaskListener` on the remote side, to `flush` at some logical conclusion. This is easily done for `DurableTaskStep` (I will file a PR for it), and for `RemoteLauncher` in core, and for this little case in the test, the change I mentioned. On the flip side, cloud implementations of `TaskListener` need to ensure that `flush` works, something I apparently still need to handle in `CloudWatchSender` in https://github.com/jenkinsci/pipeline-log-fluentd-cloudwatch-plugin/pull/8.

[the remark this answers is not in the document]
dwnusbaum: Ok, having the remote side flush when it is done with the stream seems like a reasonable solution to me.

src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
dwnusbaum: For reference: https://github.com/jenkinsci/jenkins/pull/3703

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
```diff
@@ -0,0 +1,134 @@
...
+    private static final float RECURRENCE_PERIOD_BACKOFF = 1.05f;
+
+    private long recurrencePeriod = MIN_RECURRENCE_PERIOD;
+
+    DelayBufferedOutputStream(OutputStream out) {
+        super(new FlushControlledOutputStream(out)); // default buffer size: 8Kib
+        reschedule();
+    }
+
+    private void reschedule() {
```
svanoort: This feels quite over-complicated to me for what we're trying to do, and that's concerning when combined with our overall logging complexity.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * Buffered output stream which is guaranteed to deliver content after some time even if idle and the buffer does not fill up.
+ * The automatic “flushing” does <em>not</em> flush the underlying stream, for example via {@code ProxyOutputStream.Flush}.
+ */
+final class DelayBufferedOutputStream extends BufferedOutputStream {
+
+    private static final Logger LOGGER = Logger.getLogger(DelayBufferedOutputStream.class.getName());
+
+    // TODO make these customizable (not trivial since this system properties would need to be loaded on the master side and then remoted)
+    private static final long MIN_RECURRENCE_PERIOD = 250; // ¼s
```
svanoort: :bug:  250 ms is *way* too often to flush if we want buffering to be useful here -- remember we're still going to be flushing when we hit the end of the buffer anyway, so we're only holding the last $BufferSize bytes.  I would suggest flushing no more than every 10s or so.

Also, I'm wondering if there's something that guarantees the stream will be flushed at Pipeline shutdown?  (Quite possibly, I just can't think of something).

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
```diff
@@ -0,0 +1,134 @@
...
+        boolean orig = enableFlush.get();
+        enableFlush.set(false);
+        try {
+            flush();
+        } finally {
+            enableFlush.set(orig);
+        }
+    }
+
+    void run() {
```
svanoort: :ant: Might be better named as flushSafelyAndReschedule -- I'm so glad to see this approach though, huge :+1:  for making sure we catch errors when rescheduling.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
```diff
@@ -0,0 +1,134 @@
...
+    private long recurrencePeriod = MIN_RECURRENCE_PERIOD;
+
+    DelayBufferedOutputStream(OutputStream out) {
+        super(new FlushControlledOutputStream(out)); // default buffer size: 8Kib
+        reschedule();
+    }
+
+    private void reschedule() {
+        Timer.get().schedule(new Flush(this), recurrencePeriod, TimeUnit.MILLISECONDS);
+        recurrencePeriod = Math.min((long) (recurrencePeriod * RECURRENCE_PERIOD_BACKOFF), MAX_RECURRENCE_PERIOD);
```
svanoort: Why do we need an exponential backoff for simply flushing logs (which we're presumably doing to ensure we don't lose data in the event a master goes down)?  Wouldn't it be enough to just schedule this to run periodically at a pretty long interval (say 10-30s)?  Then it won't run that often, and a few seconds of log data isn't as big a deal.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
```diff
@@ -0,0 +1,134 @@
...
+
+    // TODO make these customizable (not trivial since this system properties would need to be loaded on the master side and then remoted)
+    private static final long MIN_RECURRENCE_PERIOD = 250; // ¼s
+    private static final long MAX_RECURRENCE_PERIOD = 10_000; // 10s
+    private static final float RECURRENCE_PERIOD_BACKOFF = 1.05f;
+
+    private long recurrencePeriod = MIN_RECURRENCE_PERIOD;
+
+    DelayBufferedOutputStream(OutputStream out) {
+        super(new FlushControlledOutputStream(out)); // default buffer size: 8Kib
```
svanoort: We're keeping these streams open as long as the build is running, right?  If not, ignore the following, but if so, we should probably use 64 kB for the buffer size.  Larger buffer sizes give considerably better I/O performance (especially for HDDs).  For SSDs, performance graphs like [this one with much better throughput at larger block sizes](https://img.purch.com/capacity-performance-ssd/w/449/aHR0cDovL21lZGlhLmJlc3RvZm1pY3JvLmNvbS9ILzQvMjk2MTA0L29yaWdpbmFsL2F0dG93cml0ZS5wbmc=) or [this one](https://img.purch.com/r/711x457/aHR0cDovL21lZGlhLmJlc3RvZm1pY3JvLmNvbS9XLzQvNzg5MTI0L29yaWdpbmFsL2ltYWdlMDAxLnBuZw==) are extremely common. 

64 kB is a kind of sweet-spot which maximizes performance without excessive memory use (64 kB per currently-running-build is a fairly small contribution to overall memory use, and we could always add a system property to let people reduce that for low-memory enviornments).

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
```diff
@@ -110,16 +112,17 @@ private synchronized void open() throws IOException {
...
 
     @Override public TaskListener nodeListener(FlowNode node) throws IOException, InterruptedException {
-        return new StreamTaskListener(new IndexOutputStream(node.getId()), StandardCharsets.UTF_8);
+        return new BufferedBuildListener(new IndexOutputStream(node.getId()));
     }
 
     private void checkId(String id) throws IOException {
         assert Thread.holdsLock(this);
         if (!Objects.equals(id, lastId)) {
+            bos.flush();
```
svanoort: Maybe I misunderstand something here (trying to review fast),  but won't it defeat the point if we flush here, since `checkId` is called with interleaved parallels?  Along with that, wouldn't it make sense to buffer the index output stream too?

svanoort: Specifically b/c the index gets a *lot* of tiny writes with parallel interleaved logging.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
jglick: > I would suggest flushing no more than every 10s or so.

I guess you missed the next two constants?

> something that guarantees the stream will be flushed at Pipeline shutdown

See https://github.com/jenkinsci/workflow-api-plugin/pull/81#discussion_r228273640 plus linked PRs.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
jglick: This buffer size seemed to work well in practice. Can always be tuned. (As noted above, it is a bit tricky to allow control via system property when using remotable classes.)

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
jglick: Can you be more specific please about what is more complicated than it could be?

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
jglick: > which we're presumably doing to ensure we don't lose data in the event a master goes down

No, that is a consideration, but not the driver for the choice of `MIN_RECURRENCE_PERIOD`. Consider a user watching a build that is doing different stuff, including a number of quick `sh` steps. You want to see output from those in as close to real time as possible. But then once a given process has been running for a few minutes, the likelihood that it is being watched closely drops off and it is fine to send big chunks at intervals.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
jglick: The naming was just to match that of `Flush.run`. Note that it would seem simpler to just make this class implement `Runnable`, but I needed to use the `WeakReference` link.

jglick: > catch errors when rescheduling

I am actually on the fence about this. It is hard to know if the exception in this case was a fatal connection error, in which case it is pointless to reschedule, or something recoverable (and I struggle to come up with an example). At any rate, rescheduling after an exception seemed harmless.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
jglick: > won't it defeat the point if we flush here, since `checkId` is called with interleaved parallels

Not so often, because

> the index gets a _lot_ of tiny writes with parallel interleaved logging

is no longer true to nearly the same extent, due to the remote buffering.

It seemed unsafe to leave content not yet flushed to `log` when `log-index` already referred to that transition: your index would then contain offsets greater than the length of the file.

> wouldn't it make sense to buffer the index output stream too

Should not be necessary as `log-index` is orders of magnitude smaller than `log`.

src/test/java/org/jenkinsci/plugins/workflow/log/LogStorageTestBase.java
```diff
@@ -149,11 +148,6 @@ protected static void close(TaskListener listener) throws Exception {
         VirtualChannel channel = r.createOnlineSlave().getChannel();
         channel.call(new RemotePrint("overall from agent", overall));
         channel.call(new RemotePrint("step from agent", step));
-        while (!IOUtils.toString(text().readAll()).contains("overall from agent") || !IOUtils.toString(text().readAll()).contains("step from agent")) {
-            // TODO current cloud implementations may be unable to honor the completed flag on remotely printed messages, pending some way to have all affected loggers confirm they have flushed
```
jglick: Plan to address this differently, by having the cloud implementation honor `flush`.

jglick:
> we're still flushing too eagerly

Not really. Addressed point-by-point above.

> we could cut a Beta release

Pointless to mark it `-beta` since

* the affected code is only active when using `workflow-job` 2.26, which is not currently published
* the behavior is demonstrably more efficient than in the published non-beta release
* too few people use the experimental update center; if we want to point specific people to try specific builds, we offer them incremental releases

> [has this] had some serious manual and scaling testing[?]

I have been testing it manually with a local scalability load. That is how the problems in the previous state were identified and how these changes were developed. I do not know if you consider that “serious”.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
dwnusbaum: FWIW: I had a similar question, see https://github.com/jenkinsci/workflow-api-plugin/pull/81#discussion_r228180118 for related discussion.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
```diff
@@ -0,0 +1,148 @@
...
+ * The automatic “flushing” does <em>not</em> flush the underlying stream, for example via {@code ProxyOutputStream.Flush}.
+ */
+final class DelayBufferedOutputStream extends BufferedOutputStream {
+
+    private static final Logger LOGGER = Logger.getLogger(DelayBufferedOutputStream.class.getName());
+
+    static final class Tuning implements SerializableOnlyOverRemoting {
+        private Tuning() {}
+        // nonfinal for Groovy scripting:
+        long minRecurrencePeriod = Long.getLong(DelayBufferedOutputStream.class.getName() + ".minRecurrencePeriod", 250); // ¼s
```
svanoort: :ant: This still seems awfully short -- if it's a fast-completing task that someone would watch, it'll finish fast and get flushed anyway.  Given shell steps already come with 1/4 to 1/3 of a second of overhead, I think we can probably get away with 1-2s as a starting point.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
```diff
@@ -0,0 +1,148 @@
...
+        long maxRecurrencePeriod = Long.getLong(DelayBufferedOutputStream.class.getName() + ".maxRecurrencePeriod", 10_000); // 10s
+        float recurrencePeriodBackoff = Float.parseFloat(System.getProperty(DelayBufferedOutputStream.class.getName() + ".recurrencePeriodBackoff", "1.05"));
+        int bufferSize = Integer.getInteger(DelayBufferedOutputStream.class.getName() + ".bufferSize", 1 << 16); // 64Kib
```
svanoort: :+1: for 64k buffers

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
```diff
@@ -0,0 +1,148 @@
...
+final class DelayBufferedOutputStream extends BufferedOutputStream {
+
+    private static final Logger LOGGER = Logger.getLogger(DelayBufferedOutputStream.class.getName());
+
+    static final class Tuning implements SerializableOnlyOverRemoting {
+        private Tuning() {}
+        // nonfinal for Groovy scripting:
+        long minRecurrencePeriod = Long.getLong(DelayBufferedOutputStream.class.getName() + ".minRecurrencePeriod", 250); // ¼s
+        long maxRecurrencePeriod = Long.getLong(DelayBufferedOutputStream.class.getName() + ".maxRecurrencePeriod", 10_000); // 10s
+        float recurrencePeriodBackoff = Float.parseFloat(System.getProperty(DelayBufferedOutputStream.class.getName() + ".recurrencePeriodBackoff", "1.05"));
```
svanoort: @jglick I think maybe you meant to make the backoff **1.5** rather than **1.05**?   With 1.05 we'll flush 75 times before we hit the maximum backoff (on run 76), whereas a backoff of 1.5 it's about 9 cycles (enough to give you just under 30s of fast-flushing logs). 

The 1.5 constant matches the behavior you seemed to want here, where the 1.05 would still generate a lot of log interleaving (and a big index file) with slow-writing logs.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ * Buffered output stream which is guaranteed to deliver content after some time even if idle and the buffer does not fill up.
+ * The automatic “flushing” does <em>not</em> flush the underlying stream, for example via {@code ProxyOutputStream.Flush}.
+ */
+final class DelayBufferedOutputStream extends BufferedOutputStream {
+
+    private static final Logger LOGGER = Logger.getLogger(DelayBufferedOutputStream.class.getName());
+
+    static final class Tuning implements SerializableOnlyOverRemoting {
```
svanoort: :+1: :+1:  for making it configurable

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
svanoort: > Should not be necessary as log-index is orders of magnitude smaller than log.

Note that this may not *necessarily* be true, to give one example: parallel shell steps that log relatively slowly but constantly will generate a lot of log index entries until they hit the max backoff (one of the better arguments for a somewhat higher max backoff and growth factor).  For example, something like: 

> randomDownloadTool downloading artifact_foo.jar
> .
> .
> .
> .
> .
> download completed

Would generate this result.

However the other argument is compelling enough in general.

svanoort:
@jglick  I didn't miss the next two constants, but the use-case of a user live-tailing the log hadn't been explained (or if so I was too groggy to catch it). Their values make more sense in that light.  I think a few tweaks to their values will help ensure we don't get into an index-abuse situation again.

svanoort:
Sorry, clicked "close and comment" rather than "comment" by accident.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
svanoort: Alternately 0.250s is fine with a higher growth factor (probably the better option, since the first few outputs come fast, but then it quickly slows down to something more sedate so as to avoid spamming the log index with slow-moving logs).

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
jglick: No, 1.05 was meant.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
jglick: Given https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/85 I can try 1s, the same as the delay in `progressiveText.jelly`.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
jglick: Note that link layer packets are much smaller than this typically. However a larger buffer (divided into packets by TCP) means less overhead in the Remoting layer, for example in (de)serialization.

src/test/java/org/jenkinsci/plugins/workflow/log/LogStorageTestBase.java
jglick: (which I did)

PR #82, opened 2018-11-09
[JENKINS-54128] Define an SPI in LogStorage to satisfy WorkflowRun.getLogFile

jglick:
Factored out of https://github.com/jenkinsci/workflow-job-plugin/pull/118 at @svanoort’s request.


src/main/java/org/jenkinsci/plugins/workflow/log/LogStorage.java
```diff
@@ -108,6 +112,43 @@
...
+                 while (true) {
+                     long pos2 = logText.writeRawLogTo(pos, os);
+                     if (pos2 <= pos) {
+                         break;
+                     }
+                     pos = pos2;
+                 }
+             }
+             return f;
+         } catch (Exception x) {
```
svanoort: :ant: probably should log the exception

PR #83, opened 2018-11-15
[JENKINS-54566] finalize vs. flush

jglick:
[JENKINS-54566](https://issues.jenkins-ci.org/browse/JENKINS-54566)

Amends #81. https://github.com/jenkinsci/remoting/pull/308 suppresses the error, but this should be the actual fix. Note that flushing a soon-to-be-collected stream is intended as a best effort to capture any stray messages not already delivered, due perhaps to callables neglecting to explicitly flush.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
@@ -116,9 +116,9 @@ final class DelayBufferedOutputStream extends BufferedOutputStream {
 
     }
 
-    //
     /**
      * Flushes streams prior to garbage collection.
+     * ({@link BufferedOutputStream} does not do this automatically.)
      * TODO Java 9+ could use java.util.Cleaner
      */
     private static final class FlushRef extends PhantomReference<DelayBufferedOutputStream> {
@@ -134,7 +134,6 @@ final class DelayBufferedOutputStream extends BufferedOutputStream {
                     }
                     LOGGER.log(Level.FINE, "cleaning up phantom {0}", ref.out);
                     try {
-                        // Odd that this is not the default behavior for BufferedOutputStream.
                         ref.out.flush();
                     } catch (IOException x) {
                         LOGGER.log(Level.WARNING, null, x);
--- a/src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
@@ -28,9 +28,7 @@ import java.io.BufferedOutputStream;
 import java.io.FilterOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
-import java.lang.ref.PhantomReference;
 import java.lang.ref.Reference;
-import java.lang.ref.ReferenceQueue;
 import java.lang.ref.WeakReference;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
@@ -41,10 +39,8 @@ import org.jenkinsci.remoting.SerializableOnlyOverRemoting;
 /**
  * Buffered output stream which is guaranteed to deliver content after some time even if idle and the buffer does not fill up.
  * The automatic “flushing” does <em>not</em> flush the underlying stream, for example via {@code ProxyOutputStream.Flush}.
- * Also the stream will be flushed before garbage collection.
- * Otherwise it is similar to {@link BufferedOutputStream}.
  */
-final class DelayBufferedOutputStream extends FilterOutputStream {
+final class DelayBufferedOutputStream extends BufferedOutputStream {
 
     private static final Logger LOGGER = Logger.getLogger(DelayBufferedOutputStream.class.getName());
 
@@ -58,45 +54,6 @@ final class DelayBufferedOutputStream extends FilterOutputStream {
         static final Tuning DEFAULT = new Tuning();
     }
 
-    /**
-     * The interesting state of the buffered stream, kept as a separate object so that {@link FlushRef} can hold on to it.
-     */
-    private static final class Buffer {
-
-        final OutputStream out;
-        private final byte[] dat;
-        private int pos;
-
-        Buffer(OutputStream out, int size) {
-            this.out = out;
-            dat = new byte[size];
-        }
-
-        synchronized void drain() throws IOException {
-            if (pos == 0) {
-                return;
-            }
-            out.write(dat, 0, pos);
-            pos = 0;
-        }
-
-        void write(int b) throws IOException {
-            assert Thread.holdsLock(this);
-            if (pos == dat.length) {
-                drain();
-            }
-            dat[pos++] = (byte) b;
-        }
-
-        synchronized void write(byte[] b, int off, int len) throws IOException {
-            for (int i = 0; i < len; i++) {
-                write(b[off + i]);
-            }
-        }
-
-    }
-
-    private final Buffer buf;
     private final Tuning tuning;
     private long recurrencePeriod;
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
```diff
@@ -120,6 +116,45 @@ void flushAndReschedule() {
...
+        static {
+            Timer.get().scheduleWithFixedDelay(() -> {
+                while (true) {
+                    FlushRef ref = (FlushRef) rq.poll();
+                    if (ref == null) {
+                        break;
+                    }
+                    LOGGER.log(Level.FINE, "cleaning up phantom {0}", ref.out);
+                    try {
+                        ref.out.flush();
```
dwnusbaum: Why flush the underlying stream instead of the `DelayBufferedOutputStream` itself? Couldn't there be data left in the buffer that wouldn't get flushed?

jglick: It is not possible to get to the `DelayBufferedOutputStream` from a finalizer. But yes, this will mean `flushBuffer()` is not called. I think the only way to fix that would be to turn `DelayBufferedOutputStream` into a `FilterOutputStream` that delegates to a plain `BufferedOutputStream`.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
```diff
@@ -56,6 +58,45 @@ private Tuning() {}
...
+        synchronized void drain() throws IOException {
+            if (pos == 0) {
+                return;
+            }
+            out.write(dat, 0, pos);
+            pos = 0;
+        }
+
+        void write(int b) throws IOException {
+            assert Thread.holdsLock(this);
```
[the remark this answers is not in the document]
jglick: Could do it that way. Amounts to much the same thing.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
```diff
@@ -54,6 +58,45 @@ private Tuning() {}
         static final Tuning DEFAULT = new Tuning();
     }
 
+    /**
+     * The interesting state of the buffered stream, kept as a separate object so that {@link FlushRef} can hold on to it.
+     */
+    private static final class Buffer {
```
svanoort: Why not just use a [NIO buffer](https://docs.oracle.com/javase/7/docs/api/java/nio/Buffer.html)?

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
```diff
@@ -54,6 +58,45 @@ private Tuning() {}
...
+
+        final OutputStream out;
+        private final byte[] dat;
+        private int pos;
```
svanoort: I feel like we're tempting fate here by having both unsynchronized and synchronized access to these values without making them `volatile`

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
```diff
@@ -54,6 +58,45 @@ private Tuning() {}
...
+            assert Thread.holdsLock(this);
+            if (pos == dat.length) {
+                drain();
+            }
+            dat[pos++] = (byte) b;
+        }
+
+        synchronized void write(byte[] b, int off, int len) throws IOException {
+            for (int i = 0; i < len; i++) {
+                write(b[off + i]);
```
svanoort: Urgh, this is not very efficient, better to support writing an array chunk at a time (up to the buffer capacity, anyway)

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
```diff
@@ -62,85 +105,97 @@ private Tuning() {}
...
-            }
-        };
-
-        FlushControlledOutputStream(OutputStream out) {
-            super(out);
+    /**
+     * Flushes streams prior to garbage collection.
+     * In Java 9+ could use {@code java.util.Cleaner} instead.
+     */
+    private static final class FlushRef extends PhantomReference<DelayBufferedOutputStream> {
```
svanoort: Seriously hacky, but I *think* this will work IIUC what it's doing.

Makes me wonder why Remoting doesn't have a way to register a callback to execute before disposing the lease on an object.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
svanoort: (In that case we'd store the stream separately from a Buffer. )

svanoort:
Also the fact that we're not doing array-at-a-time writes feels like it'll turn up as a bottleneck in the future.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
jglick: I see no advantage in doing so.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
jglick: All access is synchronized.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
jglick: The OpenJDK implementation does something along those lines. We cannot copy it. I chose to write the simplest possible code that could work so it is efficiently reviewable.

jglick: More specifically, the OpenJDK implementation has an optimization for the case that `write(byte[], int, int)` is called with a `len` which is greater than the buffer size (or maybe it is the currently open space in the buffer). Since our buffer size is pretty big by default, this is unlikely to matter much. The normal mode of operation is that the buffer fills up and then is written in one chunk.

Now copying one in-memory array to another bytewise is admittedly not terribly efficient in the case of gigantic arrays (again the OpenJDK implementation has an optimization for this which we cannot legally copy), but this has no effect on I/O, and would have made the implementation more complex and harder to review.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
svanoort: At the *top* level -- it just seems silly to mix these around internally.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
svanoort: The problem is overheads from many small single-byte function calls vs. doing a single large-block write (and if it exceeds the buffer size we should probably write directly to the output).   It is not needed to do native-code work -- `System.arrayCopy` already does that for us, and reputedly they use hand-tuned assembly for that (benchmarks I've done in the past bear that out). 

See this logic from `BufferedInputStream`:

```java
public synchronized void write(byte b[], int off, int len) throws IOException {
        if (len >= buf.length) {
            /* If the request length exceeds the size of the output buffer,
               flush the output buffer and then write the data directly.
               In this way buffered streams will cascade harmlessly. */
            flushBuffer();
            out.write(b, off, len);
            return;
        }
        if (len > buf.length - count) {
            flushBuffer();
        }
        System.arraycopy(b, off, buf, count, len);
        count += len;
    }
```

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
svanoort: At the top level, yes.  Internally mixing the two forms feels a bit nonsensical.  

Anyway, this is pretty much the same comment Devin made, just phrased differently.

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
svanoort: I.E. this isn't a problem at the *disk* level because we're going to be behind one or more levels of internal buffering (here) or at the OS level.   But given these logging methods are going to be invoked a *lot* it makes sense to keep the overhead from this operation low.

jglick: Yes again I am well aware of the optimizations in `BufferedOutputStream`, which I deliberately did not attempt to copy.

src/main/java/org/jenkinsci/plugins/workflow/log/GCFlushedOutputStream.java
```diff
@@ -0,0 +1,98 @@
...
+                        ref.out.flush();
+                    } catch (IOException x) {
+                        LOGGER.log(Level.WARNING, null, x);
+                    }
+                }
+            }, 0, 10, TimeUnit.SECONDS);
+        }
+
+        static void register(GCFlushedOutputStream fos, OutputStream out) {
+            new FlushRef(fos, out, rq).enqueue();
```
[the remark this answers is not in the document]
jglick: Sorry, I recall little of this. It is possible tests in `pipeline-cloudwatch-logs` implicitly exercise this behavior. There are also some JEP-210-related tests in `workflow-durable-task-step` which deal with remote `TaskListenerDecorator`s or something like that.

PR #84, opened 2018-11-29
Update FlowGraphWalker.java

sternr:
We see many incidents where n.getParents() fail due to NullPointerException


dwnusbaum:
@sternr Thanks for the PR! I'm going to close this for now since it causes test failures and doesn't seem like the behavior we want. Feel free to create a Jira issue as I mentioned [here](https://github.com/jenkinsci/workflow-api-plugin/pull/84#pullrequestreview-180343342) for the underlying issue and ping me (`@dnusbaum` on Jira) on that ticket.

PR #88, opened 2019-04-02
Using new JenkinsRule.showAgentLogs

jglick:
Downstream of https://github.com/jenkinsci/plugin-pom/pull/188.


src/test/java/org/jenkinsci/plugins/workflow/log/LogStorageTestBase.java
```diff
@@ -173,6 +171,9 @@ protected static void close(TaskListener listener) throws Exception {
         assertEquals(stepPos, assertStepLog("1", stepPos, "", true));
         assertThat(logging.getMessages(), empty());
     }
+    protected Map<String, Level> slaveLoggers() {
```
batmat: `agentLoggers` please.

PR #89, opened 2019-04-11
Use correct variable for nodeAfter in MemoryFlowChunk constructor

dwnusbaum:
Noticed while working on adding some new tests in pipeline-graph-analysis. Thankfully, it doesn't look like there are any callers currently using this overload of the constructor ([here is a GitHub search for MemoryFlowChunk in the jenkinsci org](https://github.com/search?p=1&q=user%3Ajenkinsci+MemoryFlowChunk&type=Code)).

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/MemoryFlowChunkTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/MemoryFlowChunkTest.java
@@ -1,0 +1,58 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2019 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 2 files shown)


dwnusbaum:
> I find it disturbing that there is no test coverage for this constructor

From what I can tell, this is because the normal usage idiom is to maintain a single `MemoryFlowChunk` at the field level in graph analysis-related classes, and update its values over time rather than constructing new instances (presumably to reduce GC pressure?). See the note [here](https://github.com/jenkinsci/workflow-api-plugin/blob/9a38fc66f26667dba98c2d3652ae975d544030b7/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/StandardChunkVisitor.java#L22) for example. Other scanning-related methods and visitors use `FlowNode` directly and so don't run into the problem.

I will go ahead and add a basic unit test for the constructor.

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/MemoryFlowChunkTest.java
```diff
@@ -0,0 +1,58 @@
...
+    public void constructor() {
+        MockFlowNode start = new MockFlowNode("1");
+        MockFlowNode blockStart = new MockFlowNode("2", start);
+        MockFlowNode blockEnd = new MockFlowNode("3", blockStart);
+        MockFlowNode end = new MockFlowNode("4", blockEnd);
+        MemoryFlowChunk chunk = new MemoryFlowChunk(start, blockStart, blockEnd, end);
+        assertThat(chunk.getNodeBefore(), equalTo(start));
+        assertThat(chunk.getFirstNode(), equalTo(blockStart));
+        assertThat(chunk.getLastNode(), equalTo(blockEnd));
+        assertThat(chunk.getNodeAfter(), equalTo(end));
```
dwnusbaum: Test fails here if the change in `src/main` is reverted.

PR #91, opened 2019-04-18
[JENKINS-39203] Create new action to mark a FlowNode with a result such as unstable

dwnusbaum:
See [JENKINS-39203](https://issues.jenkins-ci.org/browse/JENKINS-39203). (I am intentionally not updating the status of that ticket or assigning it to myself until I am ready to merge/release all relevant changes to avoid giving users premature hope of a fix and so that I have clear answers for what will and won't work and when any fixes will be released.) Subsumes #63.

This PR adds a new action named `WarningAction`. This action can be stored on a `FlowNode` to mark that although it completed successfully, some kind of non-fatal problem occurred. For now, this action will be used by Blue Ocean to mark stages containing a `FlowNode` with a `WarningAction` as `Result.UNSTABLE`.

In the future, if we wanted to add the ability to track some new kind of non-fatal result, we could update `WarningAction` to store a custom `Result` (or similar) and use that in visualizations to modify how the action should be displayed.

Upstream PRs:

* https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/83
* https://github.com/jenkinsci/pipeline-graph-analysis-plugin/pull/25
* https://github.com/jenkinsci/blueocean-plugin/pull/1954
* https://github.com/jenkinsci/junit-plugin/pull/118
* https://github.com/jenkinsci/warnings-ng-plugin/pull/58

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
@@ -29,7 +29,7 @@ import hudson.remoting.ClassFilter;
 import hudson.remoting.ProxyException;
 import javax.annotation.CheckForNull;
 import org.codehaus.groovy.control.MultipleCompilationErrorsException;
-import org.jenkinsci.plugins.workflow.graph.AtomNode;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import javax.annotation.Nonnull;
 import jenkins.model.Jenkins;
 import org.apache.commons.io.output.NullOutputStream;
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
@@ -35,7 +35,7 @@ import jenkins.model.Jenkins;
 import org.apache.commons.io.output.NullOutputStream;
 
 /**
- * Attached to {@link AtomNode} that caused an error.
+ * Attached to {@link FlowNode} that caused an error.
  *
  * This has to be Action because it's added after a node is created.
  */
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/WarningAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/WarningAction.java
@@ -1,30 +1,41 @@
 package org.jenkinsci.plugins.workflow.actions;
 
 import hudson.model.Result;
+import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 
 /**
- * Action to be attached to a {@link FlowNode} to signify that some non-fatal warning occurred
- * during execution of a {@code Step}.
+ * Action to be attached to a {@link FlowNode} to signify that some non-fatal event occurred
+ * during execution of a {@code Step} but execution continued normally.
  *
- * Visualizations should treat FlowNodes with this action as if the FlowNode's result was
- * {@link Result#UNSTABLE}.
+ * {@link #withMessage} should be used whenever possible to give context to the warning.
+ * Visualizations should treat FlowNodes with this action as if the FlowNode's result was {@link #result}.
  */
-public final class WarningAction implements PersistentAction {
-    private final @Nonnull String message;
+public class WarningAction implements PersistentAction {
+    private @Nonnull Result result;
+    private @CheckForNull String message;
 
-    public WarningAction(@Nonnull String message) {
-        this.message = message;
+    public WarningAction(@Nonnull Result result) {
+        this.result = result;
     }
 
-    public @Nonnull String getMessage() {
+    public WarningAction withMessage(String message) {
+        this.message = message;
+        return this;
+    }
+
+    public @CheckForNull String getMessage() {
         return message;
+    }
+
+    public @Nonnull Result getResult() {
+        return result;
     }
 
     @Override
     public String getDisplayName() {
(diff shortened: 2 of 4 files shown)


dwnusbaum:
Closing and reopening in an attempt to get a CI build.

dwnusbaum:
I think I'm running into [JENKINS-54126](https://issues.jenkins-ci.org/browse/JENKINS-54126), so I'll just deploy snapshots manually.

src/main/java/org/jenkinsci/plugins/workflow/actions/WarningAction.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * This action can be attached to a {@link FlowNode} to signify that some non-fatal
+ * warning occurred during execution. Visualizations should treat FlowNodes with this
+ * action as if the FlowNode's result was {@link Result#UNSTABLE}.
+ */
+public final class WarningAction implements PersistentAction {
+    private final @Nonnull String message;
+
+    public WarningAction(@Nonnull String message) {
```
dwnusbaum: Maybe it would make more sense for this to be `@Nullable` so that something like `unstable()` with no message would work without having to create some kind of default message.

src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
```diff
@@ -29,13 +29,13 @@
...
 import org.codehaus.groovy.control.MultipleCompilationErrorsException;
-import org.jenkinsci.plugins.workflow.graph.AtomNode;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import javax.annotation.Nonnull;
 import jenkins.model.Jenkins;
 import org.apache.commons.io.output.NullOutputStream;
 
 /**
- * Attached to {@link AtomNode} that caused an error.
+ * Attached to {@link FlowNode} that caused an error.
```
jglick: Good catch!

src/main/java/org/jenkinsci/plugins/workflow/actions/WarningAction.java
```diff
@@ -0,0 +1,37 @@
+package org.jenkinsci.plugins.workflow.actions;
+
+import hudson.model.Result;
+import javax.annotation.Nonnull;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+
+/**
+ * This action can be attached to a {@link FlowNode} to signify that some non-fatal
+ * warning occurred during execution. Visualizations should treat FlowNodes with this
```
jglick: BTW I find it clearest to put the Javadoc summary sentence on its own line in sources, as it will be treated specially.

src/main/java/org/jenkinsci/plugins/workflow/actions/WarningAction.java
jglick: I do not see a use case for a null message, so would suggest keeping it `@Nonnull` unless and until one is found.

jglick: On second thought, perhaps it would be appropriate to have the constructor take no arguments, then use the idiom

```java
public WarningAction withMessage(@Nonnull String message) {
    this.message = message;
    return this;
}
```

as I think it is likely we will at some point want an associated URL and perhaps even a badge icon.

dwnusbaum: Any particular reason to prefer `withMessage`, `withUrl`, etc. over adding new constructors that take additional arguments? With the former pattern it is easy to forget to call `withMessage`. We could use the builder pattern and validate that the message is non-null when constructing the `WarningAction` from  the builder, but that seems like overkill.

jglick: The builder style makes it easier to introduce new features compatibly. If we decide that a message is mandatory, then keep it in the constructor.

src/main/java/org/jenkinsci/plugins/workflow/actions/WarningAction.java
dwnusbaum: Done in 7d94c66.

src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -274,20 +275,20 @@ public String getDisplayFunctionName() {
...
         BallColor c = null;
         if(error != null) {
             if(error.getError() instanceof FlowInterruptedException) {
+                // TODO: Use FlowInterruptedException.getResult() to determine ball color.
                 c = BallColor.ABORTED;
             } else {
                 c = BallColor.RED;
             }
+        } else if (getPersistentAction(WarningAction.class) != null) {
+            c = BallColor.YELLOW;
```
dwnusbaum: Here is an example of what this change looks like for the following pipeline:

```
stage('unstable-one') {
    unstable('foobar')
}
stage('success') {
    echo('whatever')
}
stage('failure') {
    unstable('three')
    error('uncaught')
}
```
Screen Shot 2019-04-23 at 12 24 36

For now only the actual `FlowNode` that has the `WarningAction` attached will be shown as unstable (unlike Blue Ocean in https://github.com/jenkinsci/pipeline-graph-analysis-plugin/pull/25 which searches all `FlowNodes` in a stage/branch for a `WarningAction`.). This seems like the most obvious behavior to me, but I could make it work like Blue Ocean if desired.

dwnusbaum:
All of the upstream PRs are using the current version of the API and I don't anticipate any further changes to it right now, so I'm going to go ahead and merge and release this plugin.

PR #92, opened 2019-04-25
[JENKINS-52189] FlowExecutionListener.onCreated

t-8ch:
This method is called after the FlowExecution has been created, but before it
is started. This allows to customize the FlowExecution earlier, especially a
GraphListener can be added that also receives the FlowStartNode.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
@@ -16,6 +16,18 @@ import javax.annotation.Nonnull;
 public abstract class FlowExecutionListener implements ExtensionPoint {
 
     /**
+     * Called when a {@link FlowExecution} has been created, but before it starts running.
+     *
+     * The {@link FlowExecution} will already have been added to the {@link FlowExecutionList} by this point.
+     * Methods relating to in-progress execution state such as {@link FlowExecution#getCurrentHeads}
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
```diff
@@ -15,6 +15,16 @@
  */
 public abstract class FlowExecutionListener implements ExtensionPoint {
 
+    /**
+     * Called when a {@link FlowExecution} has been created.
```
dwnusbaum:
```suggestion
     * Called when a {@link FlowExecution} has been created, but before it starts running.
```

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListener.java
```diff
@@ -15,6 +15,16 @@
...
+     *
+     * The {@link FlowExecution} will already have been added to the {@link FlowExecutionList} by this point.
```
dwnusbaum:
```suggestion
     * The {@link FlowExecution} will already have been added to the {@link FlowExecutionList} by this point. Methods relating to in-progress execution state such as {@link FlowExecution#getCurrentHeads} will not work as intended and should not be used.
```

t-8ch:
@dwnusbaum Thanks for the review. I used your recommendations (plus a few newlines)

PR #94, opened 2019-06-06
[JENKINS-57805] Harden bruteForceScanForEnclosingBlocks against nodes with no parents

dwnusbaum:
See [JENKINS-57805](https://issues.jenkins-ci.org/browse/JENKINS-57805).

This PR keeps `bruteForceScanForEnclosingBlocks` from throwing an `IndexOutOfBoundsException` when it encounters a `FlowNode` with no parents.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
@@ -147,6 +147,9 @@ public final class StandardGraphLookupView implements GraphLookupView, GraphList
             }
 
             // Now see if we have a winner among parents
+            if (current.getParents().isEmpty()) {
+                return null;
+            }
             FlowNode parent = current.getParents().get(0);
             if (parent instanceof BlockStartNode) {
                 nearestEnclosingBlock.put(current.getId(), parent.getId());
--- a/src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
@@ -26,8 +26,9 @@ package org.jenkinsci.plugins.workflow.graph;
 
 import hudson.model.BallColor;
 import hudson.model.Result;
-
 import java.lang.reflect.Method;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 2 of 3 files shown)


src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
```diff
@@ -54,21 +55,20 @@
...
 import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.LoggerRule;
-import org.jvnet.hudson.test.JenkinsRule;
 import org.jvnet.hudson.test.RestartableJenkinsRule;
 import org.jvnet.hudson.test.TestExtension;
 
 public class FlowNodeTest {
 
     @Rule public RestartableJenkinsRule rr = new RestartableJenkinsRule();
-    @Rule public JenkinsRule r = new JenkinsRule();
```
dwnusbaum: I was getting some curious error messages at the end of tests and then noticed this 😄

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
```diff
@@ -454,6 +457,34 @@ public void evaluate() throws Throwable {
...
+            Files.write(cpsExec.getStorageDir().toPath().resolve("3.xml"), "garbage".getBytes(StandardCharsets.UTF_8));
+        });
+        rr.then(j -> {
+            WorkflowJob job = j.jenkins.getItemByFullName("p", WorkflowJob.class);
+            WorkflowRun run = job.getBuildByNumber(1);
+            CpsFlowExecution cpsExec = (CpsFlowExecution)run.getExecution();
+            // We expect to see a warning in the logs, but we want to make sure that an IndexOutOfBoundsException
+            // is not thrown out of StandardGraphLookupView.bruteForceScanForEnclosingBlockSafety
+            assertThat(cpsExec.getCurrentHeads().get(0).getEnclosingBlocks(), hasSize(0));
+            assertThat(logging.getMessages(), hasItem("failed to load parents of 4"));
```
dwnusbaum: This comes from the warning printed to the logs by `FlowNode.loadParents` [here](https://github.com/jenkinsci/workflow-api-plugin/blob/6da45c1339cf2b9bdd9c7f79d08b6caa827af2d9/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java#L173).

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
```diff
@@ -147,6 +147,9 @@ BlockStartNode bruteForceScanForEnclosingBlock(@Nonnull final FlowNode node) {
             }
 
             // Now see if we have a winner among parents
+            if (current.getParents().isEmpty()) {
```
dwnusbaum: I think we could make the inverse of this the while loop condition and stop special-casing `FlowStartNode`, but that would mean that we would load the parents of every node that we look at, even if they are `BlockEndNode`s that could skip straight to their `BlockStartNode`, so I opted for this local change.

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
```diff
@@ -454,6 +457,34 @@ public void evaluate() throws Throwable {
...
+            WorkflowJob job = j.jenkins.createProject(WorkflowJob.class, "p");
+            job.setDefinition(new CpsFlowDefinition(
+                    "echo 'to-corrupt'\n" +
+                    "sleep 1\n" + // Trigger a save
+                    "semaphore 'marker'", true));
+            WorkflowRun run = job.scheduleBuild2(0).waitForStart();
+            SemaphoreStep.waitForStart("marker/1", run);
+            CpsFlowExecution cpsExec = (CpsFlowExecution)run.getExecution();
+            // Corrupt a flow node so that we get an error when trying to load it.
+            Files.write(cpsExec.getStorageDir().toPath().resolve("3.xml"), "garbage".getBytes(StandardCharsets.UTF_8));
```
jglick: nice

PR #96, opened 2019-06-25
[JENKINS-58084] Add StepListener

abayer:
[JENKINS-58084](https://issues.jenkins-ci.org/browse/JENKINS-58084)

(downstream in workflow-cps coming shortly)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/StepListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/StepListener.java
@@ -1,0 +1,41 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2019, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/flow/StepListener.java
```diff
@@ -0,0 +1,41 @@
...
+
+import javax.annotation.Nonnull;
+
+/**
+ * {@link StepListener}s are fired before invoking a step but after the {@link FlowNode} has been created and the
+ * {@link StepContext} has been populated. A {@link StepListener} can perform actions before the {@link Step} is
+ * executed, including throwing an exception to cause the step to fail.
+ */
+public interface StepListener extends ExtensionPoint {
+    void newStep(@Nonnull Step step, @Nonnull StepContext context) throws Exception;
```
jglick: Could probably benefit from a more descriptive method name, I just dashed this one off…

abayer: Yeah, this is very much a first draft. =)

src/main/java/org/jenkinsci/plugins/workflow/flow/StepListener.java
```diff
@@ -0,0 +1,41 @@
...
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+import org.jenkinsci.plugins.workflow.steps.Step;
+import org.jenkinsci.plugins.workflow.steps.StepContext;
+
+import javax.annotation.Nonnull;
+
+/**
+ * {@link StepListener}s are fired before invoking a step but after the {@link FlowNode} has been created and the
+ * {@link StepContext} has been populated. A {@link StepListener} can perform actions before the {@link Step} is
+ * executed, including throwing an exception to cause the step to fail.
```
dwnusbaum: I think the docs here need to be tweaked after the downstream changes.

```suggestion
 * executed, such as calling {@link StepContext#onFailure} to make the step fail before it starts.
```

src/main/java/org/jenkinsci/plugins/workflow/flow/StepListener.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+public interface StepListener extends ExtensionPoint {
+    void notifyOfNewStep(@Nonnull Step step, @Nonnull StepContext context) throws Exception;
```
dwnusbaum: Since any exceptions thrown here will just end up logging a warning over in workflow-cps, it seems better to remove `throws Exception` to make it clear that exceptions are expected to be handled by the listener (either by logging them or converting them into step failure via `context.onFailure` or something). I'd also copy some Javadoc from the interface to the method: 

```suggestion
    /**
     * Called before a {@link Step} is invoked, but after its {@link FlowNode} and {@link StepContext} have been created.
     * Listeners can make the step fail by calling {@link StepContext#onFailure}.
     */
    void notifyOfNewStep(@Nonnull Step step, @Nonnull StepContext context);
```

PR #76, opened 2018-09-05
[JENKINS-45693] Defined TaskListenerDecorator

jglick:
[JENKINS-45693](https://issues.jenkins-ci.org/browse/JENKINS-45693)

Subsumes #17. See https://github.com/jenkinsci/workflow-support-plugin/pull/72 & https://github.com/jenkinsci/workflow-job-plugin/pull/105.

PR #97, opened 2019-07-03
[JENKINS-45693] Remove beta mark from TaskListenerDecorator

jglick:
This API was introduced ten months ago in #76 and has proven itself irreplaceable in https://github.com/jenkinsci/timestamper-plugin/pull/25 and https://github.com/jenkinsci/kubernetes-plugin/pull/494, as well as one proprietary usage I am aware of by @olamy. I think it is time to be considered a supported API.


jglick:
Ping?

nick0511:
why i can`t get pipeline log use extends TaskListenerDecorator,override decorate like in  ConsoleLogFilter ,after running a  pipeline

PR #100, opened 2019-08-13
[JENKINS-58899] Avoid leaking ThreadLocals on jenkins.util.Timer threads

dwnusbaum:
See [JENKINS-58899](https://issues.jenkins-ci.org/browse/JENKINS-58899).

I think it would be possible to demonstrate the fix in a test, but I have not tried to do so.

I am a bit curious to see if we are also leaking the `ThreadLocal` just from calls to `FlushControlledOutputStream.flush` itself. The fix would be the same, and I think it would be harmless even if it didn't matter, so maybe worth adding a try/finally and calling `ThreadLocal.remove` in that method as well?

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
@@ -77,15 +77,7 @@ final class DelayBufferedOutputStream extends BufferedOutputStream {
     private void flushBuffer() throws IOException {
         ThreadLocal<Boolean> enableFlush = ((FlushControlledOutputStream) out).enableFlush;
         enableFlush.set(false);
-        try {
-            flush();
-        } finally {
-            // This method is always called from a thread in the jenkins.util.Timer thread pool. We want to avoid
-            // leaking ThreadLocals on these long-lived threads (see JENKINS-58899), and we do not care about
-            // maintaining the value from call to call, since we only set it to false here for the duration of
-            // flush, and leave it as true in all other cases and on all other threads.
-            enableFlush.remove();
-        }
+        flush(); // Note that the ThreadLocal is removed from the thread inside of FlushControlledOutputStream.flush.
     }
 
     void flushAndReschedule() {
@@ -138,8 +130,16 @@ final class DelayBufferedOutputStream extends BufferedOutputStream {
         }
 
         @Override public void flush() throws IOException {
+            try {
             if (enableFlush.get()) {
                 super.flush();
+            }
+            } finally {
+                // We want to avoid leaking ThreadLocals on long-lived threads that happen to flush this stream
+                // (see JENKINS-58899), and we do not care about maintaining the value from call to call, since we
+                // only set it to false in DelayBufferedOutputStream.flushBuffer() for the duration of a single call
+                // to flush, and leave it as true in all other cases.
+                enableFlush.remove();
             }
         }
 
--- a/src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
@@ -76,12 +76,15 @@ final class DelayBufferedOutputStream extends BufferedOutputStream {
     /** We can only call {@link BufferedOutputStream#flushBuffer} via {@link #flush}, but we do not wish to flush the underlying stream, only write out the buffer. */
     private void flushBuffer() throws IOException {
         ThreadLocal<Boolean> enableFlush = ((FlushControlledOutputStream) out).enableFlush;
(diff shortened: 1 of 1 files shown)


glavoie:
@dwnusbaum I tested the fix and it's indeed also leaking through `FlushControlledOutputStream.flush`. Adding a call to `remove()` in there does solve the leak.

I first tested adding a `remove()` call on `close()`, but it didn't work.

dwnusbaum:
@glavoie Thanks for testing! Did you happen to keep track of which threads were leaking leaking through `FlushControlledOutputStream.flush`? No worries if you did not, it would just be interesting to know for future reference.

glavoie:
Seems to be the same thread (`org.jenkinsci.plugins.workflow.steps.durable_task.DurableTaskStep`. I only looked at those.

dwnusbaum:
Ok, Stack Overflow and I came up with some code to see where things were leaking and to make sure the fixes worked. Without either fix, when running `FileLogStorageTest.remoting`, we see the following threads reference the listed number of `DelayBufferedOutputStream$FlushControlledOutputStream$1` `ThreadLocal`s:

```
Computer.threadPoolForRemoting [#10]: 1
Computer.threadPoolForRemoting [#4]: 1
jenkins.util.Timer [#7]: 1
jenkins.util.Timer [#6]: 1
jenkins.util.Timer [#8]: 1
jenkins.util.Timer [#5]: 1
jenkins.util.Timer [#4]: 2
jenkins.util.Timer [#1]: 1
jenkins.util.Timer [#3]: 1
jenkins.util.Timer [#10]: 2
jenkins.util.Timer [#2]: 3
Computer.threadPoolForRemoting [#17]: 1
jenkins.util.Timer [#9]: 2
Executing org.jenkinsci.plugins.workflow.log.FileLogStorageTest: 5
Computer.threadPoolForRemoting [#13]: 1
```

With the current fix, all references from `jenkins.util.Timer` go away, but we still leak on `Computer.threadPoolForRemoting` and `Executing org.jenkinsci.plugins.workflow.log.FileLogStorageTest` (only relevant for testing, though I guess analogous to `flush` being called from other random threads). If we call `ThreadLocal.remove` in `flush` itself, all references go away.

<details>
    <summary>I don't think it makes sense to actually add my code to FileLogStorageTest, but here it is for future reference:</summary>

```
@After public void checkLeaks() throws Exception {
    Set<Thread> threads = Thread.getAllStackTraces().keySet();
    for (Thread thread : threads) {
        Integer count = getThreadLocalCounts(thread).get("org.jenkinsci.plugins.workflow.log.DelayBufferedOutputStream$FlushControlledOutputStream$1");
        if (count != null) {
            System.out.println(thread.getName() + ": " + count);
        }
    }
}

private Map<String, Integer> getThreadLocalCounts(Thread thread) throws Exception {
    Field threadLocalsField = Thread.class.getDeclaredField("threadLocals");
    threadLocalsField.setAccessible(true);
    Object threadLocalTable = threadLocalsField.get(thread);
    if (threadLocalTable == null) {
        return Collections.emptyMap();
    }

    // Get a reference to the array holding the thread local variables inside the
    // ThreadLocalMap of the current thread
    Class threadLocalMapClass = Class.forName("java.lang.ThreadLocal$ThreadLocalMap");
    Field tableField = threadLocalMapClass.getDeclaredField("table");
    tableField.setAccessible(true);
    Object table = tableField.get(threadLocalTable);

    // The key to the ThreadLocalMap is a WeakReference object. The referent field of this object
    // is a reference to the actual ThreadLocal variable
    Field referentField = Reference.class.getDeclaredField("referent");
    referentField.setAccessible(true);

    Map<String, Integer> counts = new HashMap<>();
    for (int i=0; i < Array.getLength(table); i++) {
        Object entry = Array.get(table, i);
        if (entry != null) {
            ThreadLocal threadLocal = (ThreadLocal)referentField.get(entry);
            if (threadLocal != null) {
                counts.merge(threadLocal.getClass().getName(), 1, (oldValue, value) -> oldValue + value);
            }
        }
    }
    return counts;
}
```
</details>

src/main/java/org/jenkinsci/plugins/workflow/log/DelayBufferedOutputStream.java
```diff
@@ -135,9 +130,17 @@ void flushAndReschedule() {
         }
 
         @Override public void flush() throws IOException {
+            try {
             if (enableFlush.get()) {
```
car-roll: indentation?

dwnusbaum: Intentionally unchanged to keep `git blame` as useful as possible, but I am happy to indent it if preferred.

car-roll: that's what i figured. Nah, i'm fine with it.

PR #106, opened 2019-10-04
Added a new method to be executed before completion of Flow

joseblas:
added a method for beforeCompleted to be run before onCompleted.

cc @bitwiseman @dwnusbaum


abayer:
Looks fine to me.

dwnusbaum:
We ended up moving the place that `fireCompleted` was called instead of using a new method for now, see https://github.com/jenkinsci/workflow-job-plugin/pull/143.

PR #107, opened 2019-10-04
Fix bad assertion in TestVisitor#assertNoIllegalNullsInEvents

basil:
Noticed this while looking through some IntelliJ warnings. This assertion as-written is always true because `ce.ids` is a primitive `int[]`. What was really meant was to check if the integer is -1. This is because the integers in the data structure are initialized to -1 and only filled in if the corresponding value in `FlowNode` is non-null:

```
public static class CallEntry {
    CallType type;
    int[] ids = {-1, -1, -1, -1};

    public void setIds(FlowNode... nodes) {
        for (int i=0; i<nodes.length; i++) {
            if (nodes[i] == null) {
                ids[i] = -1;
            } else {
                ids[i] = Integer.parseInt(nodes[i].getId());
            }
        }
    }
    [...]
}
```


basil:
CC: @jenkinsci/hacktoberfest

dwnusbaum:
Thanks!

PR #110, opened 2019-11-07
Fix memory leak of GroovyClassLoaders

jonsten:
MissingPropertyExceptions thrown due to Pipeline scripts can, given the
right circumstances, contain references indirect references to the
Groovy class loader for the pipeline execution. A memory leak will
occur if the exception is saved in a ErrorAction, i.e. if the exception
flies all the way or passes a step before it is caught and treated.

In more detail, if the pipeline code tries to lookup a missing property
in the WorkflowScript class instance, then a MissingPropertyException
will be thrown and the type field in the exception will contain a link
to the WorkflowScript instance, which in turn contains a static link to
the Groovy class loader.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
@@ -30,7 +30,6 @@ import groovy.lang.MissingPropertyException;
 import hudson.remoting.ProxyException;
 import javax.annotation.CheckForNull;
 import org.codehaus.groovy.control.MultipleCompilationErrorsException;
-import org.codehaus.groovy.tools.GroovyClass;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import javax.annotation.Nonnull;
 import jenkins.model.Jenkins;
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
@@ -26,9 +26,11 @@ package org.jenkinsci.plugins.workflow.actions;
 
 import groovy.lang.GroovyClassLoader;
 import groovy.lang.MissingMethodException;
+import groovy.lang.MissingPropertyException;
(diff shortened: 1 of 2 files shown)


jonsten:
Worth mentioning that there are probably other exception types which will cause similar problems. This only fixes one of them, but it might be worth looking for other common exceptions which holds references to the class.

BTW, failing Jenkins seems to be infrastructure related.

dwnusbaum:
Triggering a new build.

jonsten:
> directly analogous to #102.
Indeed, that was my inspiration for how to fix bug.

src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
```diff
@@ -26,9 +26,11 @@
 
 import groovy.lang.GroovyClassLoader;
 import groovy.lang.MissingMethodException;
+import groovy.lang.MissingPropertyException;
 import hudson.remoting.ProxyException;
 import javax.annotation.CheckForNull;
 import org.codehaus.groovy.control.MultipleCompilationErrorsException;
+import org.codehaus.groovy.tools.GroovyClass;
```
dwnusbaum: Btw, is this actually used? If not, would be nice to remove it.

jonsten: No it is it not needed, I blame my IDE, but in the end I'm the one controlling it so I should probably blame myself ;) Fixed now anyways.

jonsten:
Triggering a new build.

jonsten:
What is the trick for retriggering tests, cause closing and reopening never works for me...

dwnusbaum:
> What is the trick for retriggering tests, cause closing and reopening never works for me...

I usually close the PR, wait 10 seconds, and reopen it. Not sure if write permissions on the GitHub repo are needed or something. Lately, ci.jenkins.io has been having problems, so maybe it was unresponsive when you tried to retrigger it?

jonsten:
Yea, I probably didn't wait long enough before reopening. Anyways, glad to see this fix merged now, thanks!

PR #118, opened 2020-02-15
[JEP-223] Make Pipeline durability config available to MANAGE

daniel-beck:
Downstream of https://github.com/jenkinsci/jenkins/pull/4501, not intended to be merged as is.

> Screenshot


src/main/java/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel.java
```diff
@@ -43,6 +44,10 @@ public void setDurabilityHint(FlowDurabilityHint hint){
             save();
         }
 
+        public Permission getRequiredPermission() {
```
jglick: Is this an `@Override`?

daniel-beck: Yes.

PR #119, opened 2020-02-20
 update workflow-api version 2.40

joseblas:
This update is needed for PCT issues

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.40</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.138.4</jenkins.version>
+        <jenkins.version>2.150.1</jenkins.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <workflow-support-plugin.version>2.21</workflow-support-plugin.version>
@@ -150,7 +150,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>ssh-slaves</artifactId>
-            <version>1.26</version>
+            <version>1.30.1</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.40</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.150.1</jenkins.version>
+        <jenkins.version>2.150.3</jenkins.version>
         <java.level>8</java.level>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.40</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.138.4</jenkins.version>
+        <jenkins.version>2.150.1</jenkins.version>
```
dwnusbaum: I'd go to the latest release of that LTS line, so 2.150.3:

```suggestion
        <jenkins.version>2.150.3</jenkins.version>
```

pom.xml
```diff
@@ -150,7 +150,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>ssh-slaves</artifactId>
-            <version>1.26</version>
+            <version>1.30.1</version>
```
jglick: Can you use the BOM instead?

src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
```diff
@@ -108,7 +113,9 @@ private static void wrapInContainer(@Nonnull JenkinsRule r, @CheckForNull Artifa
             DumbSlave agent;
             if (image != null) {
                 runningContainer = image.start(JavaContainer.class).start();
-                agent = new DumbSlave("test-agent", "/home/test/slave", new SSHLauncher(runningContainer.ipBound(22), runningContainer.port(22), "test", "test", "", ""));
+                StandardUsernameCredentials creds = new UsernamePasswordCredentialsImpl(CredentialsScope.SYSTEM, "test","desc","test","test");
```
jglick:
```suggestion
                StandardUsernameCredentials creds = new UsernamePasswordCredentialsImpl(CredentialsScope.SYSTEM, "test", "desc", "test", "test");
```

pom.xml
joseblas: do you mean to add a bom to the ssh-slaves plugin?

jglick: No, to use `jenkinsci/bom` here rather than hard-coding individual plugin versions. Not necessary and could be done as a follow-up of course.

joseblas: lgtm, I will do it in another branch. Thanks

dwnusbaum:
Ok, I tried to track down the reason for the change. My best guess is that it is because https://github.com/jenkinsci/ssh-slaves-plugin/pull/114 broke binary compatibility by removing deprecated methods on `SSHLauncher` which we were using in tests, and because that PR updated the Jenkins core dependency to a newer version than this plugin.

PR #121, opened 2020-02-26
Start using bom

joseblas:
start using bom dependencies

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,67 +67,67 @@
         <jenkins.version>2.150.3</jenkins.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
-        <workflow-support-plugin.version>2.21</workflow-support-plugin.version>
         <useBeta>true</useBeta>
     </properties>
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>io.jenkins.tools.bom</groupId>
+                <artifactId>bom-2.150.x</artifactId>
+                <version>5</version>
+                <scope>import</scope>
+                <type>pom</type>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
     <dependencies>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-step-api</artifactId>
-            <version>2.16</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>scm-api</artifactId>
-            <version>2.2.6</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-support</artifactId>
-            <version>${workflow-support-plugin.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.26</version>
             <scope>test</scope>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -67,67 +67,67 @@
...
-            <version>2.16</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>scm-api</artifactId>
-            <version>2.2.6</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-support</artifactId>
```
jglick: Probably was here only to manage a version.

```suggestion
```

pom.xml
```diff
@@ -67,67 +67,67 @@
...
         <dependency>  <!-- For Semaphore step -->
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-support</artifactId>
-            <version>${workflow-support-plugin.version}</version>
             <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-scm-step</artifactId>
```
jglick: probably this too
```suggestion
```

pom.xml
```diff
@@ -67,67 +67,67 @@
...
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-durable-task-step</artifactId>
-            <version>2.22</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
```
jglick:
```suggestion
```

joseblas:
@jglick should I try to flatten the dependencies? remove all the unnecessary ones?

jglick:
> remove all the unnecessary ones?

This. Sometimes a `dependency` was in a POM merely to satisfy `RequireUpperBoundDeps`; it was not being used as such. In these cases it should always have been in `dependencyManagement`, and can now be removed if the BOM declares it.

joseblas:
sorry @jglick, I got confused.

pom.xml
```diff
@@ -67,80 +67,56 @@
...
         </dependency>
-        <dependency>
-            <groupId>org.jenkins-ci.plugins.workflow</groupId>
-            <artifactId>workflow-scm-step</artifactId>
-            <version>2.4</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.jenkins-ci.plugins.workflow</groupId>
-            <artifactId>workflow-durable-task-step</artifactId>
```
jglick: This is definitely being used in tests and so ought to remain as an explicit dependency, even if it happens to also be pulled in from something else.

joseblas: Is a test dependency, I guess as far as tests are passing, is all good.

jglick: Looks like it is being pulled in via `workflow-basic-steps`. Just poor style generally to rely on transitive dependencies for things actually used directly. Anyway yes, if a future update of `workflow-basic-steps` drops that dep (as is quite possible in this case), updating to that would (I hope!) cause tests to start failing until you readded the explicit `workflow-durable-task-step` dep.

joseblas: I'd say `workflow-basic-steps` is directly used and its dependencies. But, let me know if you want this added back.

jglick: The `node` step is used in a few tests. Not important, just FYI.

PR #124, opened 2020-05-15
[JENKINS-62303] Jenkins.MANAGE users access to global config

amuniz:
[JENKINS-62303](https://issues.jenkins-ci.org/browse/JENKINS-62303)

Allow Jenkins.MANAGE users to configure "Pipeline Default Speed/Durability Level" under Jenkins Global configuration.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.41</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.176.4</jenkins.version>
+        <jenkins.version>2.222.3</jenkins.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <useBeta>true</useBeta>
@@ -132,6 +132,13 @@
           <groupId>org.jenkins-ci.plugins</groupId>
           <artifactId>apache-httpcomponents-client-4-api</artifactId>
           <scope>test</scope>
+        </dependency>
+        <!-- Extracted from core at 2.190.1 -->
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>trilead-api</artifactId>
+            <version>1.0.6</version>
+            <scope>test</scope>
         </dependency>
     </dependencies>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.41</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.222.3</jenkins.version>
+        <jenkins.version>2.176.4</jenkins.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <useBeta>true</useBeta>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.41</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.176.4</jenkins.version>
+        <jenkins.version>2.222.3</jenkins.version>
```
dwnusbaum, line 67: I would prefer to still support older LTS lines for a widely-used plugin like this. Is it possible to use reflection and drop the `@Override` annotation on `getRequiredGlobalConfigPagePermission` so that we can keep the core dependency as-is while still supporting the new behavior for anyone running the latest LTS?

amuniz: Sure! Done.

amuniz:
Not sure why the build is failing... can someone re-trigger?

dwnusbaum:
> Not sure why the build is failing... can someone re-trigger?

Looks like normal ci.jenkins.io infra issues:

> C:\Jenkins\workspace\Plugins\workflow-api-plugin: There is not enough space on the disk.

I will close and open the PR to trigger a new build (For future reference you can do the same, just wait 10 or so seconds after closing before reopening)

src/main/java/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel.java
```diff
@@ -72,9 +74,19 @@ public static FlowDurabilityHint getSuggestedDurabilityHint() {
...
         public Permission getRequiredGlobalConfigPagePermission() {
-            return Jenkins.MANAGE;
+            return getJenkinsManageOrAdmin();
+        }
+
+        // TODO: remove when Jenkins core baseline is 2.222+
+        Permission getJenkinsManageOrAdmin() {
+            Permission systemRead;
+            try { // System Read is available starting from Jenkins 2.222 (https://jenkins.io/changelog/#v2.222). See JEP-224 for more info
+                systemRead = (Permission) ReflectionUtils.getPublicProperty(Jenkins.get(), "SYSTEM_READ");
```
mikecirioli, line 85: Shoudn't this be "MANAGE" instead of "SYSTEM_READ"?

amuniz: :facepalm: you are totally right. Fixed.

amuniz: Good catch BTW! Thanks.

amuniz:
This should be ready now @dwnusbaum

src/test/java/org/jenkinsci/plugins/workflow/flow/DurabilityBasicsTest.java
```diff
@@ -43,4 +60,44 @@ public void defaultHandling() {
...
+                Collection<Descriptor> descriptors = Functions.getSortedDescriptorsForGlobalConfigUnclassified();
+                Optional<Descriptor> found = descriptors.stream().filter(descriptor -> descriptor instanceof GlobalDefaultFlowDurabilityLevel.DescriptorImpl).findFirst();
+                assertTrue("Global configuration should be accessible to MANAGE users", found.isPresent());
+            }
+        });
+    }
+
+    // TODO: remove when Jenkins core baseline is 2.222+
+    private Permission getJenkinsManage() throws NoSuchMethodException, IllegalAccessException, InvocationTargetException {
+        // Jenkins.MANAGE is available starting from Jenkins 2.222 (https://jenkins.io/changelog/#v2.222). See JEP-224 for more info
```
dwnusbaum, line 100:
```suggestion
        // Jenkins.MANAGE is available starting from Jenkins 2.222 (https://jenkins.io/changelog/#v2.222). See JEP-223 for more info
```

PR #128, opened 2020-10-22
use jenkins-core stash returned values

joseblas:


car-roll (approved):

jglick:
For https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/133 presumably?

pom.xml
```diff
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.41</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.176.4</jenkins.version>
+        <jenkins.version>2.264-SNAPSHOT</jenkins.version>
```
jglick, line 67: Why?

pom.xml
```diff
@@ -132,5 +132,10 @@
           <artifactId>apache-httpcomponents-client-4-api</artifactId>
           <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>trilead-api</artifactId>
```
jglick, line 137: ??

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -109,26 +114,36 @@ public static void stash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull
      * @see StashAwareArtifactManager#stash
      */
     @SuppressFBWarnings(value="RV_RETURN_VALUE_IGNORED_BAD_PRACTICE", justification="fine if mkdirs returns false")
-    public static void stash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull FilePath workspace, @Nonnull Launcher launcher, @Nonnull EnvVars env, @Nonnull TaskListener listener,
+    public static List<File> stash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull FilePath workspace, @Nonnull Launcher launcher, @Nonnull EnvVars env, @Nonnull TaskListener listener,
                              @CheckForNull String includes, @CheckForNull String excludes, boolean useDefaultExcludes, boolean allowEmpty) throws IOException, InterruptedException {
         Jenkins.checkGoodName(name);
+        List<File> files = new ArrayList<>();
+        DirScanner.Glob scanner = new DirScanner.Glob(Util.fixEmpty(includes) == null ? "**" : includes, excludes, useDefaultExcludes);
+        scanner.scan(new File(workspace.getRemote()), new FileVisitor() {
```
jglick, line 122: `new File(workspace.getRemote())` is a sure sign that something is wrong. This takes a file path on an agent (in general) and treats it as a local file. Which will result, in, say, `C:\home\jenkins\something` or `/var/jenkins_home/C:/Program Files/something` or similar junk.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -109,26 +114,36 @@ public static void stash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull
...
+            @Override
+            public void visit(File f, String relativePath) throws IOException {
+                files.add(f);
```
jglick, line 125: We have a `TaskListener` right here. Why not print whatever messages right now, in real time, rather than returning metadata and asking the caller to print messages? There is no need for a new API or a patch to `workflow-basic-steps`.

jglick (changes requested):

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
joseblas: I'd say is nicer to print the messages in the workflow-basic-steps plugin, as is its own responsibility.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
joseblas: this is going away, all that will be done in Jenkins Core, where it actually happens.

pom.xml
joseblas: needed for testing

pom.xml
joseblas: Changes in Jenkins core https://github.com/jenkinsci/jenkins/pull/5023#pullrequestreview-514587286

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
jglick: Easier to leave it to stashing/unstashing implementation, which already has a `TaskListener` exactly for printing such messages as may be relevant. In the case of an alternate implementation such as https://github.com/jenkinsci/artifact-manager-s3-plugin/blob/5e6724a0af838ac48e7c0533520b1e3f081ceef3/src/main/java/io/jenkins/plugins/artifact_manager_jclouds/JCloudsArtifactManager.java#L214-L324 it may be appropriate to do things at different times, or from controller vs. agent, or to show more or less detail, or different formats, etc. etc. From the PoV of `workflow-basic-steps`, these are all implementation details.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
jglick: Saw that, and the same problem applies there. Do not use `File`.

pom.xml
jglick: Testing…what?

jglick:
See #62.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -109,26 +114,36 @@ public static void stash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull
      * @see StashAwareArtifactManager#stash
      */
     @SuppressFBWarnings(value="RV_RETURN_VALUE_IGNORED_BAD_PRACTICE", justification="fine if mkdirs returns false")
-    public static void stash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull FilePath workspace, @Nonnull Launcher launcher, @Nonnull EnvVars env, @Nonnull TaskListener listener,
+    public static List<File> stash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull FilePath workspace, @Nonnull Launcher launcher, @Nonnull EnvVars env, @Nonnull TaskListener listener,
```
jglick, line 117: This is not binary compatible I think (changed return value).

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -147,18 +162,18 @@ public static void unstash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnu
      * @throws AbortException in case there is no such saved stash
      * @see StashAwareArtifactManager#unstash
      */
-    public static void unstash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull FilePath workspace, @Nonnull Launcher launcher, @Nonnull EnvVars env, @Nonnull TaskListener listener) throws IOException, InterruptedException {
+    public static List<File> unstash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull FilePath workspace, @Nonnull Launcher launcher, @Nonnull EnvVars env, @Nonnull TaskListener listener) throws IOException, InterruptedException {
```
jglick, line 165: ditto

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -147,18 +162,18 @@ public static void unstash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnu
...
         Jenkins.checkGoodName(name);
         StashAwareArtifactManager saam = stashAwareArtifactManager(build);
         if (saam != null) {
             saam.unstash(name, workspace, launcher, env, listener);
-            return;
+            return Collections.emptyList();
```
jglick, line 170: So this is clearly wrong: will not work with `artifact-manager-s3`.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
joseblas: then, it will be printed only if the implementation has it, not the plugin. Maybe the workflow-basic-step could provide a default implementation, that could be changed in the *Artifact Managers*

pom.xml
joseblas: for using JenkinsRule (or similar)

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
joseblas: Olivier mention that in other PR, It will be changed and _overloaded_

pom.xml
jglick: Right, but why this plugin, which does not appear to have anything to do with the changes you are making? Smells like a POM mistake.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
joseblas: artefact managers like s3 and azure will be done later.  Besides, what should be done with your own jcloud artefact manager?

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
jglick: Not sure I understand what you are saying.

I am simply proposing to move all `TaskListener` messaging into the code which actually does tarring or untarring.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
jglick: That is `artifact-manager-s3`, which includes a generic layer based on JClouds plus an implementation using the JClouds binding for S3. Read https://jenkins.io/jep/202 before touching anything in this area.

pom.xml
jglick: Use incremental versions if you want CI.

bitwiseman:
@joseblas 
If you get a chance to update this feel free to reopen.

PR #133, opened 2021-01-09
[JENKINS-64438] Implement removeAll on FlowNode actions to support addOrReplaceAction.

mrginglymus:
This fixes https://issues.jenkins.io/browse/JENKINS-64438, which means that `addOrReplaceAction` on `FlowNode`s no longer explodes with a `UnsupportedOperationException`.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
@@ -486,6 +486,7 @@ Action format:
         });
     }
 
+    @Issue("JENKINS-64438")
     @Test
     public void addOrReplaceActionWorks()  {
         rr.then(r -> {
@@ -613,4 +614,3 @@ Action format:
         }
     }
 }
-
--- a/src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
@@ -486,6 +486,15 @@ Action format:
         });
     }
 
+    @Test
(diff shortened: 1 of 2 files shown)


timja (approved):
cc @dwnusbaum @bitwiseman

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
```diff
@@ -486,6 +486,15 @@ public void nodeWithNoParentsInBruteForceScanForEnclosingBlock() {
         });
     }
 
+    @Test
```
dwnusbaum, line 489: Just to check, does this test fail without the fix? Also, can you please add `@Issue` so we know that this is a regression test?

```suggestion
    @Issue("JENKINS-64438")
    @Test
```

dwnusbaum (commented):
@mrginglymus Thanks for the PR! Can you please add `@Issue` to the new test?

src/test/java/org/jenkinsci/plugins/workflow/graph/FlowNodeTest.java
timja: I can it locally and I can confirm that without the fix the test fails

mrginglymus: Yes - despite the commit order I did run the test without the fix to check it failed :)

dwnusbaum (approved):
Thanks! Once the CI build passes I'll merge this.

timja:
> Thanks! Once the CI build passes I'll merge this.

you could enable auto-complete in the repo settings 😉

PR #135, opened 2021-01-16
Make workflow-api compatible with Guava 21.0 and newer

timja:


src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -180,9 +181,11 @@ public void onLoaded() {
                 Futures.addCallback(e.getCurrentExecutions(false), new FutureCallback<List<StepExecution>>() {
                     @Override
                     public void onSuccess(List<StepExecution> result) {
-                        LOGGER.log(FINE, "Will resume {0}", result);
-                        for (StepExecution se : result) {
-                            se.onResume();
+                        if (result != null) {
```
jglick, line 184: Why would it be null?

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -5,6 +5,7 @@
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
```
jglick, line 8: As in https://github.com/jenkinsci/workflow-support-plugin/pull/114#discussion_r585678752 can this be made to tolerate either old or new Guava in the short term?

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -26,6 +26,7 @@
 
 import com.google.common.base.Predicate;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import org.checkerframework.checker.nullness.qual.Nullable;
```
jglick, line 29: Why?

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
jglick: Or just use `@SuppressFBWarnings`?

timja: why suppress? if we don't want a null check shouldn't we just add an assertion that it isn't null

jglick: Sure,

```java
assert result != null;
```

is also fine.

amuniz:
@timja https://github.com/timja/workflow-api-plugin/pull/1

timja:
@car-roll @dwnusbaum @bitwiseman PTAL

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -259,4 +258,25 @@ public void onFailure(Throwable t) {
...
+     * for compatibility with newer (> 18.0) versions of guava.
+     */
+    private static ExecutorService newExecutorService() {
+        try {
+            try {
+                // guava older than 18
+                Method method = MoreExecutors.class.getMethod("sameThreadExecutor");
+                return (ExecutorService) method.invoke(null);
+            } catch (NoSuchMethodException e) {
+                // TODO invert this to prefer the newer guava method once guava is upgraded in Jenkins core
```
jglick, line 273: Or simply update `jenkins.version`.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -194,7 +198,7 @@ public void onFailure(Throwable t) {
                             LOGGER.log(WARNING, "Failed to load " + e, t);
                         }
                     }
-                });
+                }, newExecutorService());
```
jglick, line 201: Not sure I get it. The original code was not specifying an executor service, so why do we need to _add_ this code now?

timja: Deprecated in v22 https://github.com/google/guava/blob/v22.0/guava/src/com/google/common/util/concurrent/Futures.java#L1073

Removed in v26
https://github.com/google/guava/blob/v26.0/guava/src/com/google/common/util/concurrent/Futures.java (not there anymore)

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -254,4 +258,25 @@ public void onFailure(Throwable t) {
...
+     * for compatibility with newer (> 18.0) versions of guava.
+     */
+    private static ExecutorService newExecutorService() {
+        try {
+            try {
+                // guava older than 18
+                Method method = MoreExecutors.class.getMethod("sameThreadExecutor");
+                return (ExecutorService) method.invoke(null);
+            } catch (NoSuchMethodException e) {
+                // TODO invert this to prefer the newer guava method once guava is upgraded in Jenkins core
```
timja, line 273:
```suggestion
                // TODO invert this to prefer the newer guava method once guava is upgraded in Jenkins core, (or update jenkins.version)
```

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
jglick: https://github.com/google/guava/blob/v22.0/guava/src/com/google/common/util/concurrent/Futures.java#L1158-L1167 I guess you mean. :+1:

jglick (approved):

timja:
@car-roll @dwnusbaum @bitwiseman ping

car-roll (approved):

basil (commented):
Alternatively I think you could take the `MoreExecutors` and `DirectExecutor` classes from jenkinsci/workflow-support-plugin#117 and then pass `MoreExecutors.directExecutor()` without having to use reflection here.

basil (approved):

timja:
release please? ❤️

jglick:
https://github.com/jenkinsci/jep/blob/master/jep/229/README.adoc#motivation

PR #153, opened 2021-06-04
[JENKINS-65821] Introducing some synchronisation mechanisms to prevent some race condition

twasyl:
[JENKINS-65821](https://issues.jenkins.io/browse/JENKINS-65821)

On some big pipelines, the graph may take some time to be generated due to possible synchronous updates to `Map`s. Trying to introduce some synchronisation mechanism to prevent those race conditions.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
@@ -39,7 +39,7 @@ public final class StandardGraphLookupView implements GraphLookupView, GraphList
     @Override
     public synchronized void onNewHead(@Nonnull FlowNode newHead) {
         if (newHead instanceof BlockEndNode) {
-            blockStartToEnd.put(((BlockEndNode) newHead).getStartNode().getId(), newHead.getId());
+            blockStartToEnd.put(((BlockEndNode)newHead).getStartNode().getId(), newHead.getId());
             String overallEnclosing = nearestEnclosingBlock.get(((BlockEndNode) newHead).getStartNode().getId());
             if (overallEnclosing != null) {
                 nearestEnclosingBlock.put(newHead.getId(), overallEnclosing);
@@ -92,7 +92,7 @@ public final class StandardGraphLookupView implements GraphLookupView, GraphList
         scan.setup(start.getExecution().getCurrentHeads());
         for (FlowNode f : scan) {
             if (f instanceof BlockEndNode) {
-                BlockEndNode end = (BlockEndNode) f;
+                BlockEndNode end = (BlockEndNode)f;
                 BlockStartNode maybeStart = end.getStartNode();
                 // Cache start in case we need to scan again in the future
                 blockStartToEnd.put(maybeStart.getId(), end.getId());
@@ -115,6 +115,7 @@ public final class StandardGraphLookupView implements GraphLookupView, GraphList
     /** Do a brute-force scan for the enclosing blocks **/
     synchronized BlockStartNode bruteForceScanForEnclosingBlock(@Nonnull final FlowNode node) {
         FlowNode current = node;
+
         while (!(current instanceof FlowStartNode)) {  // Hunt back for enclosing blocks, a potentially expensive operation
             if (current instanceof BlockEndNode) {
                 // Hop over the block to the start
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
@@ -13,6 +13,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * Provides overall insight into the structure of a flow graph... but with limited visibility so we can change implementation.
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
```diff
@@ -31,13 +31,17 @@
...
+        }
+        synchronized (nearestEnclosingBlock) {
+            nearestEnclosingBlock.clear();
+        }
     }
 
     /** Update with a new node added to the flowgraph */
     @Override
-    public void onNewHead(@Nonnull FlowNode newHead) {
+    public synchronized void onNewHead(@Nonnull FlowNode newHead) {
```
twasyl, line 44: I preferred to `synchronized` the whole method as it uses both `blockStartToEnd` and `nearestEnclosingBlock` and the paths in the method are pretty simple. I thought it would be more readable this way as synchronising each maps at the proper use wouldn't have a huge added value.

jtnord: but as us this is still no longer safe?
thread 1 could call `bruteForceScanForEnd(..)` or `getEndNode()` and thread2 `onNewHead(...)`  resulting in concurrent search and modification which is IIUC the exact thing you are trying to defend against?

twasyl: Yes, you're right. I'm going to change that. Thanks

amuniz:
Wouldn't be better to use a rw lock so reads do not block each other?

twasyl:
> Wouldn't be better to use a rw lock so reads do not block each other?

I also thought about that, but was not really sure due to some discussion that were happening. I could probably follow this path.

dwnusbaum (commented):
Are you trying to improve performance, or is this intended to fix some concurrency issue you ran into? It's not clear to me how adding locks would increase performance. If you did see a concurrency issue, do you have stack traces or any more details about the issue?

(I don't see any specific problems with a patch like this, but depending on the details of the issue this might not be the best place to add synchronization. Also, since there are always fallback code paths here in case the caches are empty, and the computed values should always be the same for the same Pipeline graph, you might be able to just switch to `ConcurrentHashMap` to resolve any concurrency issues without needing to add explicit locks to this code, maybe with some minor updates to use methods like `compute` if you are seeing a lot of concurrent reads and want to make sure the cached values are used.)

jglick:
I have seen the thread dumps and I _suspect_ the bug was improper concurrent access to `HashMap` leading to corruption and infinite loops. Would be best to file in Jira and include (sanitized) dumps there.

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
```diff
@@ -24,50 +25,62 @@
...
     static final String INCOMPLETE = "";
 
-    /** Map the blockStartNode to its endNode, to accellerate a range of operations */
+    /** Map the blockStartNode to its endNode, to accelerate a range of operations */
     HashMap<String, String> blockStartToEnd = new HashMap<>();
 
     /** Map a node to its nearest enclosing block */
     HashMap<String, String> nearestEnclosingBlock = new HashMap<>();
 
+    final ReentrantLock lock = new ReentrantLock();
```
jglick, line 34:
```suggestion
    /** Lock for {@link #blockStartToEnd} and {@link #nearestEnclosingBlock}. */
    final ReentrantLock lock = new ReentrantLock();
```

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
```diff
@@ -24,50 +25,62 @@
...
+
     public void clearCache() {
-        blockStartToEnd.clear();
-        nearestEnclosingBlock.clear();
+        lock.lock();
+        try {
+            blockStartToEnd.clear();
+            nearestEnclosingBlock.clear();
+        } finally {
+            lock.unlock();
```
jglick, line 42: https://bugs.openjdk.java.net/browse/JDK-7084550 :cry: Could also just use a simple monitor, since you are not using any special features of `ReentrantLock`.

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
```diff
@@ -24,50 +25,62 @@
...
-            // Now we try to generate enclosing block info for caching, by looking at parents
-            // But we don't try to hard -- usually the cache is populated and we defer recursive walks of the graph
-            List<FlowNode> parents = newHead.getParents();
-            if (parents.size() > 0) {
-                FlowNode parent = parents.get(0);  // Multiple parents only for end of a parallel, and then both are BlockEndNodes
+        lock.lock();
+        try {
+            if (newHead instanceof BlockEndNode) {
+                blockStartToEnd.put(((BlockEndNode) newHead).getStartNode().getId(), newHead.getId());
+                String overallEnclosing = nearestEnclosingBlock.get(((BlockEndNode) newHead).getStartNode().getId());
```
jglick, line 53: The sort of thing that makes me dubious you could a lockless data structure for this purpose.

jglick (approved):
Looks right.

jglick:
> leading to corruption and infinite loops

Meaning that the PR description here is incorrect. The problem seems to have been a race condition and a full-blown spin hang, not just less-than-ideal performance.

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
jglick: For that matter, the simplest fix would probably just be to slap `synchronized` on all these methods. The only caller is https://github.com/jenkinsci/workflow-api-plugin/blob/01fbb0814a68950f927e9ee0739b2c6f835a2daa/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java#L256-L318 so there is no risk of using `this` as the monitor.

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
jglick: **Resolve conversation**?

jglick (approved):

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
jglick: @dwnusbaum suggests that you probably could, if you wanted to review the logic carefully.

dwnusbaum (approved):

jtnord:
what is the impact of this change - AFAICT it will force multiple read threads (the UI) to be single threaded.
Would a `ConcurrentHashMap` be better than this level of synchronization here?  or if you do not need re-entrant locks using a read/write lock?

jglick:
@jtnord yes perhaps. Unclear if concurrent access is common enough to be worth the extra risk & complexity though. These methods should all complete quickly enough that I doubt we want to bother optimizing more.

timja:
FYI we're seeing performance problems on ci.jenkins.io and this is a likely culprit

https://www.irccloud.com/pastebin/JGLey7jD/

Yesterday @daniel-beck noticed:

> more than a 100 threads are blocked at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.getCurrentHeads for a generate-data build

ci.jenkins.io is down atm because it's system log filled up with the stacktrace posted at the top

jglick:
> this is a likely culprit

You mean, a regression from this PR, or the bug that this PR purports to fix?

timja:
Regression from this PR, will see if ci.jenkins.io mis-behaves again and get a thread dump

jglick:
I see. We can try switching to a `ConcurrentHashMap`.

timja:
Few people are hitting this on Jira: https://issues.jenkins.io/browse/JENKINS-65885

olamy:
Definitely using a simple `ConcurrentHashMap` worths a try.

jglick:
Oh that is a _deadlock_, not just higher thread contention. In such a case this must be reverted immediately and an emergency fix cut. @bitwiseman @twasyl @car-roll @dwnusbaum whoever else may be listening

bitwiseman:
I'll file as soon as I'm back at my desk.

bitwiseman:
FYI, `StandardGraphLookupView` implements `GraphListener.Synchronous`:

https://github.com/jenkinsci/workflow-api-plugin/blob/ed2467fafbeadd12641395a4a0feb0d1004ad694/src/main/java/org/jenkinsci/plugins/workflow/flow/GraphListener.java#L46-L51

I'm not sure, but it seems like `synchronized` is not a good idea here.

PR #159, opened 2021-06-17
Remove most the Guava usage

olamy:
- start work on get rid of guava...
- remove most of the guava usage

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
@@ -26,6 +26,8 @@ package org.jenkinsci.plugins.workflow.actions;
 
 import hudson.PluginManager;
 import hudson.model.Describable;
+
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -34,6 +36,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import org.apache.commons.collections.CollectionUtils;
@@ -67,6 +70,10 @@ public abstract class ArgumentsAction implements PersistentAction {
         UNSERIALIZABLE
     }
 
+    private static final List<Class> WRAPPER_TYPES = Arrays.asList(Boolean.class, Byte.class, Character.class,
+                                                              Double.class, Float.class, Integer.class,
+                                                              Long.class, Short.class, Void.class);
+
     /** Largest String, Collection, or array size we'll retain -- provides a rough size limit on any single field.
      *  Set to 0 or -1 to remove length limits.
      */
@@ -85,7 +92,7 @@ public abstract class ArgumentsAction implements PersistentAction {
         if (maxElements <= 0 ) {
             return false;
         }
-        if (o == null || o.getClass().isPrimitive() || o.getClass().isEnum()) {
+        if (o == null || WRAPPER_TYPES.contains(o.getClass()) || o.getClass().isEnum()) {
             return false;
         }
         if (o instanceof CharSequence) {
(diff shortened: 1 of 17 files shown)


src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -87,7 +85,7 @@ public static boolean isOversized(@CheckForNull Object o, final int maxElements)
         if (maxElements <= 0 ) {
             return false;
         }
-        if (o == null || Primitives.isWrapperType(o.getClass()) || o.getClass().isEnum()) {
+        if (o == null || o.getClass().isPrimitive() || o.getClass().isEnum()) {
```
dwnusbaum, line 88: I think these have different meanings. The old Guava code is true for boxed types like `Integer` or `Double`, whereas the new code is true for primitive types like `int` or `double`.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
```diff
@@ -164,7 +162,7 @@ public String getUrlName() {
         if (internalArgs.size() == 0) {
             return Collections.emptyMap();
         }
-        HashMap<String, Object> filteredArguments = Maps.newHashMapWithExpectedSize(internalArgs.size());
+        HashMap<String, Object> filteredArguments = new HashMap(internalArgs.size());
```
dwnusbaum, line 165: Note that there is a subtle difference between these, see the discussion of `loadFactor` in the `HashMap` documentation.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/AbstractFlowScanner.java
```diff
@@ -244,15 +247,16 @@ public final void remove() {
      * @return First matching node, or null if no matches found
      */
     @CheckForNull
+    @Deprecated
     public FlowNode findFirstMatch(@CheckForNull Collection<FlowNode> heads,
-                                           @CheckForNull Collection<FlowNode> blackListNodes,
-                                           Predicate<FlowNode> matchCondition) {
+                                   @CheckForNull Collection<FlowNode> blackListNodes,
+                                   Predicate<FlowNode> matchCondition) {
```
dwnusbaum, line 253: FWIW, I think changing signatures like this will be a breaking change for some plugins.

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
olamy: Yes I know but looking at the context. It probably doesn't hurt...
```
        HashMap<String, Object> filteredArguments = new HashMap(internalArgs.size());
        for (Map.Entry<String, Object> entry : internalArgs.entrySet()) {
            if (entry.getValue() != null && !(entry.getValue() instanceof NotStoredReason)) {
                // TODO this is incorrect: value could be a Map/List with some nested entries that are NotStoredReason
                filteredArguments.put(entry.getKey(), entry.getValue());
            }
        }
``` 
If 25% of the internalArgs `Map` is filtered by the `if` condition this `optimisation` with size/loadfactor doesn't make any sense.
and by the way looking at the rest of the code this `Map` is only read. 
I might more using something such?:
```
        Map<String, Object>  filteredArguments = internalArgs.entrySet().stream()
            .filter(entry -> entry.getValue() != null && !(entry.getValue() instanceof NotStoredReason))
            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
```

src/main/java/org/jenkinsci/plugins/workflow/actions/ArgumentsAction.java
olamy: good catch

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/AbstractFlowScanner.java
olamy: Yup agree, it's a bit too agressive change... Even if I checked with usage-in-plugins and there is not so much usage of that.
If the code is using only lambda is not a problem it will work with this change :)
but yeah might be better to keep both (but that's a lot... especially a lot are already marked deprecated..)

dwnusbaum: Yeah, I think a lot of cases would be fine, but stuff like [this](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/15978cd172a8d34e2cd6bf8a6bfeaa5514d44dad/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy#L158-L176) along with callers like [this](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/4e07aed2384229428debc93da38512ddc60fdf60/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy#L358) would need to be updated.

dwnusbaum (commented):
I think there is one change here that will break `workflow-cps`'s tests in the PCT, and there are a few breaking changes that I think are probably safe in practice but that I would recommend either making them non-breaking or taking them a bit further to clean up a few more uses of Guava.

timja:
@olamy do you plan to revisit this? (hit in https://github.com/jenkinsci/junit-realtime-test-reporter-plugin/pull/33#discussion_r780708613)

jglick:
Conflicts with #177.

olamy:
old work. better to forget this.

PR #160, opened 2021-06-21
[JENKINS-65821] Fix race condition by using ConcurrentHashMap

bitwiseman:
JENKINS-65821 described a race condition resulting in high cpu usage.  #153 attempted to address these using locking, but resulted in deadlocks. 

This change goes the other direction and attempts to address the issue using `ConcurrentHashMap` instead of `HashMap`.  

@twasyl

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
@@ -11,13 +11,12 @@ import javax.annotation.Nonnull;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Provides overall insight into the structure of a flow graph... but with limited visibility so we can change implementation.
- * Designed to work entirely on the basis of the {@link FlowNode#id} rather than the {@link FlowNode}s themselves.
+ * Designed to work entirely on the basis of the {@link FlowNode#getId()} rather than the {@link FlowNode}s themselves.
  */
 @SuppressFBWarnings(value = "ES_COMPARING_STRINGS_WITH_EQ", justification = "Can can use instance identity when comparing to a final constant")
 @Restricted(NoExternalUse.class)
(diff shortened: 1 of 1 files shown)


jglick (approved):
I guess? I trust you have examined the code flow carefully and verified that there is no risk of one thread having written to just one of these maps when another thread comes along and tries to read from them. I suppose it cannot be worse than using `HashMap`. @dwnusbaum mentioned that the usages here were idempotent and thus safe to use locklessly.

olamy (approved):

twasyl:
I was wondering if one possible fix would be, in each _entry point_ of a `StandardGraphLookupView` instance, to create new maps from the instance ones and pass them along to all required methods. This way you only need to synchronize when creating those copies and then you ensure that even if the instance maps are updated in one thread, computations in other threads are based on the same data from their beginning to their end.

bitwiseman:
@jglick Yes, in my opinion it has to be better than HashMap. 

@twasyl 
That seems like a lot of extra work being done.

bitwiseman:
@jglick  From what I can see, this is approach is safe.  As you say, idempotent, the two caches do not depend on each other to update, and any read failure results in calling the "brute force" method the recalculate missing values.

src/main/java/org/jenkinsci/plugins/workflow/graph/StandardGraphLookupView.java
```diff
@@ -25,10 +25,10 @@
     static final String INCOMPLETE = "";
 
     /** Map the blockStartNode to its endNode, to accellerate a range of operations */
-    HashMap<String, String> blockStartToEnd = new HashMap<>();
+    ConcurrentHashMap<String, String> blockStartToEnd = new ConcurrentHashMap<>();
```
jglick, line 28: This should be fine since the value type is just `String` that already exists. Just a reminder that in general you need to be careful about having one thread add an entry to a map (or whatever) that another thread might pull off without the use of any synchronization barriers: if the object were recently constructed, its constructor might not have completed by the time the other thread uses it. One of those weird theoretical problems that probably does not happen often in reality but is hard to reason about.

bitwiseman: > if the object were recently constructed, its constructor might not have completed by the time the other thread uses it.

Wow, good to know.  Thanks!

PR #163, opened 2021-06-29
Redefining `SURVIVABLE_NONATOMIC` to use atomic writes but to skip `fsync(2)`

basil:
Not yet ready for review. Filing just to get an incremental build.

See [JENKINS-66001](https://issues.jenkins.io/browse/JENKINS-66001) and jenkinsci/jenkins#5599. Part 2 of a 5-part series to make Pipeline's `SURVIVABLE_NONATOMIC` mode behave [as advertised in the documentation](https://www.jenkins.io/doc/book/pipeline/scaling-pipeline/#what-am-i-giving-up-with-this-durability-setting-trade-off) (parts 3 through 5 will be in `workflow-support`, `workflow-cps`, and `workflow-job` respectively).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowDurabilityHint.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowDurabilityHint.java
@@ -23,6 +23,8 @@
  */
 package org.jenkinsci.plugins.workflow.flow;
 
+import hudson.util.AtomicFileWriter;
+
 import javax.annotation.Nonnull;
 
 /**
@@ -33,13 +35,15 @@ import javax.annotation.Nonnull;
  * @author Sam Van Oort
  */
 public enum FlowDurabilityHint {
-    PERFORMANCE_OPTIMIZED(false, false, Messages.FlowDurabilityHint_PERFORMANCE_OPTIMIZED_description(), Messages.FlowDurabilityHint_PERFORMANCE_OPTIMIZED_tooltip()),
+    PERFORMANCE_OPTIMIZED(false, false, false, Messages.FlowDurabilityHint_PERFORMANCE_OPTIMIZED_description(), Messages.FlowDurabilityHint_PERFORMANCE_OPTIMIZED_tooltip()),
 
-    SURVIVABLE_NONATOMIC(false,  true, Messages.FlowDurabilityHint_SURVIVABLE_NONATOMIC_description(), Messages.FlowDurabilityHint_SURVIVABLE_NONATOMIC_tooltip()),
+    SURVIVABLE_NONATOMIC(true, false, true, Messages.FlowDurabilityHint_SURVIVABLE_NONATOMIC_description(), Messages.FlowDurabilityHint_SURVIVABLE_NONATOMIC_tooltip()),
 
-    MAX_SURVIVABILITY (true,  true, Messages.FlowDurabilityHint_MAX_SURVIVABILITY_description(), Messages.FlowDurabilityHint_MAX_SURVIVABILITY_tooltip());
+    MAX_SURVIVABILITY(true, true, true, Messages.FlowDurabilityHint_MAX_SURVIVABILITY_description(), Messages.FlowDurabilityHint_MAX_SURVIVABILITY_tooltip());
 
     private final boolean atomicWrite;
+
+    private final boolean fsync;
 
     private final boolean persistWithEveryStep;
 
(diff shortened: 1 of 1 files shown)


bitwiseman (approved):

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowDurabilityHint.java
```diff
@@ -33,32 +37,61 @@
...
+    private final boolean force;
+
     private final boolean persistWithEveryStep;
 
     private final String description;
 
     private final String tooltip;
 
+    @Deprecated
     FlowDurabilityHint (boolean useAtomicWrite, boolean persistWithEveryStep, @Nonnull String description, String tooltip) {
```
jglick, line 57: No need to deprecate a package-accessible member—just delete it.

jglick:
There is a problem with core Javadoc. I am looking into it.

jtnord (commented):
the title is incorrect - writes are definatly not atomic with this change (writes never are atomic)

The move of the file after writing *may* be atomic, however it also may not be.  This is both OperatingSystem *and* file system dependant, and the atomicity of the move (overwrite/replace) is best effort.

thus this not correct to say it will tollerate JVM crashes.  It *may* for some users of some operating systems on some filesystems

src/main/resources/org/jenkinsci/plugins/workflow/flow/Messages.properties
```diff
@@ -1,6 +1,6 @@
 FlowDurabilityHint.PERFORMANCE_OPTIMIZED.description=Performance-optimized: much faster (requires clean shutdown to save running pipelines)
 FlowDurabilityHint.PERFORMANCE_OPTIMIZED.tooltip=Avoids writing data with every step, avoids atomic writes of data.  Pipelines can resume if Jenkins shuts down cleanly, but running pipelines lose step information and cannot resume if Jenkins unexpectedly fails.
-FlowDurabilityHint.SURVIVABLE_NONATOMIC.description=Less durability, a bit faster (specialty use only)
-FlowDurabilityHint.SURVIVABLE_NONATOMIC.tooltip=Writes data with every step but avoids atomic writes. On some filesytems this is faster than maximum durability mode, but running pipeline data may be lost if disk writes are interrupted or fail.
+FlowDurabilityHint.SURVIVABLE_NONATOMIC.description=Less durability, a bit faster (requires stable OS and storage but tolerates dirty JVM shutdown)
```
jtnord, line 3:
```suggestion
FlowDurabilityHint.SURVIVABLE_NONATOMIC.description=Less durability, a bit faster (requires stable OS and storage but may tolerate a dirty JVM shutdown if certain OS/Filesystem combinations are met)
```

src/main/resources/org/jenkinsci/plugins/workflow/flow/Messages.properties
```diff
@@ -1,6 +1,6 @@
...
+FlowDurabilityHint.SURVIVABLE_NONATOMIC.tooltip=Writes data with every step but avoids flushing the page cache to the storage device. On some filesytems this is faster than maximum durability mode, but running pipeline data may be lost if disk writes are interrupted or fail.
```
jtnord, line 4:
```suggestion
FlowDurabilityHint.SURVIVABLE_NONATOMIC.tooltip=Writes data with every step but avoids flushing the page cache to the storage device. On some filesytems this is faster than maximum durability mode, but running pipeline data may be lost if disk writes are interrupted or fail, or if the JVM terminates abruptly in some OS and filesystem combinations.
```

jtnord (commented):
update comments to reflect what is actually happening and that this is both OS and filesystem dependant

jtnord:
@basil are you still interested in this PR, do you want to look at the suggestions?

jglick:
> It _may_ [tolerate JVM crashes] for some users of some operating systems on some filesystems

FYI I have recently been checking behavior of Jenkins when restored from EBS snapshots (using default durability settings) and do occasionally see weird cases of stray `atomicxxx.tmp` files inside running build directories but have not tried to track down the behavior at this level. I have always been leery of asking users to pick a durability setting without necessarily understanding the consequences; if you could assume that the filesystem orders writes (maybe not a good assumption) then you could rework how metadata files are stored to use a more journaled mode, so that the last file written includes links to the desired versions of previous files, but this could be significant effort that might be better spent on e.g. SQL storage.

PR #164, opened 2021-07-14
Make use of optimizations in DirScanner when "includes" is empty.

h4n23s:
The [``DirScanner.Glob.scan()``](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/util/DirScanner.java#L131) method already checks if ``includes`` and ``excludes`` are empty and then uses an optimized ``DirScanner`` to improve performance. This PR removes a bit of code which was blocking this behaviour.

```java
public void scan(File dir, FileVisitor visitor) throws IOException {
    if(fixEmpty(includes)==null && excludes==null) {
         // optimization
         new Full().scan(dir,visitor);
         return;
    }
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
@@ -123,7 +123,7 @@ public class StashManager {
             listener.getLogger().println("Warning: overwriting stash ‘" + name + "’");
         }
         try (OutputStream os = new FileOutputStream(storage)) {
-            int count = workspace.archive(ArchiverFactory.TARGZ, os, new DirScanner.Glob(Util.fixEmpty(includes) == null ? "**" : includes, excludes, useDefaultExcludes));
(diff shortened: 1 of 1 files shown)


h4n23s:
I just noticed that with this change the behaviour of ``stash`` will change significantly: In my tests the old version stashed way less files (even with ``useDefaultExcludes: false``) than what was stashed with my change. For reference, here's the pipeline I used:

```groovy
pipeline {
    agent any

    stages {
        stage('Stash') {
            steps {
                cleanWs()
                sh 'git clone https://github.com/jenkinsci/jenkins.git'
                stash name: 'cloned', useDefaultExcludes: false
            }
        }
        stage('Unstash') {
            steps {
                unstash name: 'cloned'
            }
        }
    }
}
```

Even though there might probably be some improvements to speed, I'm closing this PR as it's changes are breaking.

h4n23s:
This could maybe be related to [JENKINS-56098](https://issues.jenkins.io/browse/JENKINS-56098)?

PR #178, opened 2021-11-04
[JENKINS-67164] Call StepExecution.onResume directly from WorkflowRun.onLoad rather than via FlowExecutionList.ItemListenerImpl to ensure step resumption

dwnusbaum:
See [JENKINS-67164](https://issues.jenkins.io/browse/JENKINS-67164).

We have seen various reports of cases where a Pipeline resumes after a Jenkins restart and is occupying an executor slot, but the Pipeline in question appears to have already completed based on log output and it hangs forever after resuming. For example, see [JENKINS-50199](https://issues.jenkins.io/browse/JENKINS-50199) (fixed via hardening a different resumption code path) and [JENKINS-43587](https://issues.jenkins.io/browse/JENKINS-43587) (which maybe should not have been marked as resolved).

I was able to reproduce this issue when restoring Jenkins from a backup taken while Pipeline builds were in progress. The Pipelines in question were in the process of running a `SynchronousNonBlockingStepExecution` when the backup was taken, and that step should have failed when the Pipeline resumed and `FlowExecutionListener$ItemListenerImpl` called `StepExecution.onResume`, but this call never happened because the builds in question were not present in `org.jenkinsci.plugins.workflow.flow.FlowExecutionList.xml` on disk for some reason, so the execution just hung forever waiting for that step to complete. Additionally, the build logs indicated that the builds had already proceeded past this step, so it is not clear to me why `program.dat` was outdated. The Pipelines were using the standard `PERFORMANCE_OPTIMIZED` durability mode, and their `build.xml` files showed them as still in progress.

This PR attempts to harden the `StepExecution` resumption code by replacing `FlowExecutionListener.ItemListenerImpl` with a `FlowExecutionListener` that implements `onResumed`. `WorkflowRun.onLoad` directly calls `FlowExecutionListener.fireOnResumed` when it resumes in-progress Pipelines (happens automatically when `WorkflowRun.onLoad` is called), so it is guaranteed to be called for every Pipeline that resumes regardless of the stated of `FlowExecutionList`.

I will keep investigating to try to understand how this might happen in cases besides backups, since I think that this issue has been seen even when just restarting a controller normally.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
@@ -243,6 +243,16 @@ public class FlowExecutionList implements Iterable<FlowExecution> {
             Futures.addCallback(e.getCurrentExecutions(false), new FutureCallback<List<StepExecution>>() {
                 @Override
                 public void onSuccess(List<StepExecution> result) {
+                    if (e.isComplete()) {
+                        // WorkflowRun.onLoad will not fire onResumed if the serialized execution was already
+                        // complete when loaded, but right now (at least for workflow-cps), the execution resumes
+                        // asynchronously before WorkflowRun.onLoad completes, so it is possible that the execution
+                        // finishes before onResumed gets called.
+                        // That said, there is nothing to prevent the execution from completing right after we check
+                        // isComplete. If we want to fully prevent that, we would need to delay actual execution
+                        // resumption until WorkflowRun.onLoad completes or add some form of synchronization.
+                        return;
+                    }
                     FlowExecutionList list = FlowExecutionList.get();
                     FlowExecutionOwner owner = e.getOwner();
                     if (!list.runningTasks.contains(owner)) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
@@ -10,6 +10,7 @@ import hudson.Extension;
 import hudson.ExtensionList;
 import hudson.XmlFile;
 import hudson.init.Terminator;
+import hudson.model.listeners.ItemListener;
 import hudson.remoting.SingleLaneExecutorService;
 import hudson.util.CopyOnWriteList;
 import jenkins.model.Jenkins;
@@ -161,6 +162,21 @@ public class FlowExecutionList implements Iterable<FlowExecution> {
             l = new FlowExecutionList();
         }
         return l;
+    }
+
+    /**
+     * When Jenkins starts up and everything is loaded, be sure to proactively resurrect
+     * all the ongoing {@link FlowExecution}s so that they start running again.
+     */
+    @Extension
+    public static class ItemListenerImpl extends ItemListener {
+        @Override
+        public void onLoaded() {
+            for (final FlowExecution e : FlowExecutionList.get()) {
+                // The call to FlowExecutionOwner.get in the implementation of iterator() is sufficent to load the Pipeline.
+                LOGGER.log(Level.FINE, "Eagerly loaded {0}", e);
+            }
+        }
     }
 
     /**
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
@@ -10,7 +10,6 @@ import hudson.Extension;
 import hudson.ExtensionList;
 import hudson.XmlFile;
 import hudson.init.Terminator;
-import hudson.model.listeners.ItemListener;
 import hudson.remoting.SingleLaneExecutorService;
 import hudson.util.CopyOnWriteList;
 import jenkins.model.Jenkins;
@@ -30,6 +29,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.DoNotUse;
 
@@ -164,41 +164,6 @@ public class FlowExecutionList implements Iterable<FlowExecution> {
     }
 
     /**
-     * When Jenkins starts up and everything is loaded, be sure to proactively resurrect
-     * all the ongoing {@link FlowExecution}s so that they start running again.
-     */
-    @Extension
-    public static class ItemListenerImpl extends ItemListener {
-        @Inject
-        FlowExecutionList list;
-
-        @Override
-        public void onLoaded() {
-            for (final FlowExecution e : list) {
-                LOGGER.log(Level.FINE, "Eager loading {0}", e);
-                Futures.addCallback(e.getCurrentExecutions(false), new FutureCallback<List<StepExecution>>() {
-                    @Override
-                    public void onSuccess(List<StepExecution> result) {
-                        LOGGER.log(Level.FINE, "Will resume {0}", result);
-                        for (StepExecution se : result) {
-                            se.onResume();
-                        }
-                    }
-
-                    @Override
-                    public void onFailure(Throwable t) {
-                        if (t instanceof CancellationException) {
-                            LOGGER.log(Level.FINE, "Cancelled load of " + e, t);
-                        } else {
-                            LOGGER.log(Level.WARNING, "Failed to load " + e, t);
-                        }
-                    }
-                }, MoreExecutors.directExecutor());
-            }
-        }
-    }
-
-    /**
      * Enumerates {@link StepExecution}s running inside {@link FlowExecution}.
      */
     @Extension
@@ -251,5 +216,37 @@ public class FlowExecutionList implements Iterable<FlowExecution> {
         SingleLaneExecutorService executor = get().executor;
         executor.shutdown();
         executor.awaitTermination(1, TimeUnit.MINUTES);
+    }
+
+    /**
+     * Whenever a Pipeline resumes, resume all incomplete steps in its {@link FlowExecution}.
+     *
+     * Called by {@code WorkflowRun.onLoad}, so guaranteed to run if a Pipeline resumes regardless of its presence in
+     * {@link FlowExecutionList}.
+     */
+    @Extension
+    public static class ResumeStepExecutionListener extends FlowExecutionListener {
+        @Override
+        public void onResumed(@Nonnull FlowExecution e) {
+            Futures.addCallback(e.getCurrentExecutions(false), new FutureCallback<List<StepExecution>>() {
+                @Override
+                public void onSuccess(List<StepExecution> result) {
+                    LOGGER.log(Level.FINE, "Will resume {0}", result);
+                    for (StepExecution se : result) {
+                        se.onResume();
+                    }
+                }
+
+                @Override
+                public void onFailure(Throwable t) {
+                    if (t instanceof CancellationException) {
+                        LOGGER.log(Level.FINE, "Cancelled load of " + e, t);
+                    } else {
+                        LOGGER.log(Level.WARNING, "Failed to load " + e, t);
+                    }
+                }
+
+            }, MoreExecutors.directExecutor()); // TODO: Unclear if we need to run this asynchronously or if StepExecution.onResume has any particular thread requirements.
+        }
     }
 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
@@ -5,7 +5,6 @@ import com.google.common.collect.AbstractIterator;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
-import com.google.inject.Inject;
 import hudson.Extension;
 import hudson.ExtensionList;
 import hudson.XmlFile;
@@ -184,14 +183,11 @@ public class FlowExecutionList implements Iterable<FlowExecution> {
      */
     @Extension
     public static class StepExecutionIteratorImpl extends StepExecutionIterator {
-        @Inject
-        FlowExecutionList list;
-
         @Override
         public ListenableFuture<?> apply(final Function<StepExecution, Void> f) {
             List<ListenableFuture<?>> all = new ArrayList<>();
 
-            for (FlowExecution e : list) {
+            for (FlowExecution e : FlowExecutionList.get()) {
                 ListenableFuture<List<StepExecution>> execs = e.getCurrentExecutions(false);
                 all.add(execs);
                 Futures.addCallback(execs,new FutureCallback<List<StepExecution>>() {
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
@@ -5,6 +5,8 @@ import com.google.common.collect.AbstractIterator;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
+import edu.umd.cs.findbugs.annotations.CheckForNull;
+import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.Extension;
 import hudson.ExtensionList;
 import hudson.XmlFile;
@@ -28,8 +30,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-import javax.annotation.CheckForNull;
-import javax.annotation.Nonnull;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.DoNotUse;
 
@@ -239,7 +239,7 @@ public class FlowExecutionList implements Iterable<FlowExecution> {
     @Extension
     public static class ResumeStepExecutionListener extends FlowExecutionListener {
         @Override
-        public void onResumed(@Nonnull FlowExecution e) {
+        public void onResumed(@NonNull FlowExecution e) {
             Futures.addCallback(e.getCurrentExecutions(false), new FutureCallback<List<StepExecution>>() {
                 @Override
                 public void onSuccess(List<StepExecution> result) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
@@ -247,6 +247,13 @@ public class FlowExecutionList implements Iterable<FlowExecution> {
             Futures.addCallback(e.getCurrentExecutions(false), new FutureCallback<List<StepExecution>>() {
                 @Override
                 public void onSuccess(List<StepExecution> result) {
+                    FlowExecutionList list = FlowExecutionList.get();
+                    FlowExecutionOwner owner = e.getOwner();
+                    if (!list.runningTasks.contains(owner)) {
+                        LOGGER.log(Level.WARNING, "Resuming {0}, which is missing from FlowExecutionList ({1}), so registering it now.",
+                                new Object[] {owner, list.runningTasks.getView()});
+                        list.register(owner);
+                    }
                     LOGGER.log(Level.FINE, "Will resume {0}", result);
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -163,41 +163,6 @@ public static FlowExecutionList get() {
         return l;
     }
 
-    /**
-     * When Jenkins starts up and everything is loaded, be sure to proactively resurrect
-     * all the ongoing {@link FlowExecution}s so that they start running again.
-     */
-    @Extension
-    public static class ItemListenerImpl extends ItemListener {
```
dwnusbaum, line 171: I tested things a bit, and as far as I can tell this is not necessary. In-progress builds seem to get loaded and resume even without this class. I will test investigate further to see what exactly is calling `WorkflowRun.onLoad` to understand whether it is reliable or if we need to bring this back in some form.

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
```diff
@@ -79,4 +91,60 @@
...
+            r.waitForMessage("Starting non-resumable step", b);
+            // TODO: Unclear how this might happen in practice.
+            FlowExecutionList.get().unregister(b.asFlowExecutionOwner());
+        });
+        sessions.then(r -> {
+            WorkflowJob p = r.jenkins.getItemByFullName("p", WorkflowJob.class);
+            WorkflowRun b = p.getBuildByNumber(1);
+            r.waitForCompletion(b);
+            r.assertBuildStatus(Result.FAILURE, b);
+            r.assertLogContains("Unable to resume NonResumableStep", b);
```
dwnusbaum, line 109: If you replace these lines with `Thread.sleep(5000)` and then check the logs, you will see that something still loads the build (which causes the Pipeline to resume) even after the removal of `FlowExecutionList$ItemListenerImpl`.

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
```diff
@@ -79,4 +91,60 @@
...
 
+    @Issue("TODO")
+    @Test public void resumeStepExecutions() throws Throwable {
+        sessions.then(r -> {
+            WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+            p.setDefinition(new CpsFlowDefinition("noResume()", true));
+            WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+            r.waitForMessage("Starting non-resumable step", b);
+            // TODO: Unclear how this might happen in practice.
+            FlowExecutionList.get().unregister(b.asFlowExecutionOwner());
```
dwnusbaum, line 102: As mentioned, I am not sure how this might happen outside of something like a backup where files are being copied indiscriminately. The code around build completion in `WorkflowRun` looks reasonable to me from a quick glance, and the call to `unregister` is the final thing that happens, so the only thing that comes to mind is that there is some issue causing the build to not be saved as expected, for example perhaps some issue with a `BulkChange` being aborted or something like that.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -252,4 +217,36 @@ public void onFailure(Throwable t) {
...
+     * {@link FlowExecutionList}.
+     */
+    @Extension
+    public static class ResumeStepExecutionListener extends FlowExecutionListener {
+        @Override
+        public void onResumed(@Nonnull FlowExecution e) {
+            Futures.addCallback(e.getCurrentExecutions(false), new FutureCallback<List<StepExecution>>() {
+                @Override
+                public void onSuccess(List<StepExecution> result) {
+                    LOGGER.log(Level.FINE, "Will resume {0}", result);
```
dwnusbaum, line 234: TODO: maybe add a sanity check to see if the execution exists in `FlowExecutionList` and if not, add it.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
jglick: You need to consult `FlowExecutionList` after a resume because otherwise there is no guarantee that all the `WorkflowRun`s which were running before will actually be visited. Due to lazy loading, Jenkins might never even pay attention to older builds. Now _often_ the running builds would be loaded for one reason or another as you load the dashboard especially weather column (last five builds) and so on, but this is not necessarily true in all cases.

jglick (changes requested):

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
dwnusbaum: > You need to consult FlowExecutionList after a resume because otherwise there is no guarantee that all the WorkflowRuns which were running before will actually be visited.

Makes sense, but that's why I want to test things a bit more to see exactly what is causing builds to be loaded even when this is not present. Maybe all we need to do here is iterate over `FlowExecutionList` and nothing else.

jglick: > exactly what is causing builds to be loaded even when this is not present

Potentially a variety of things. Put in a `Thread.dumpStackTrace();` and you will see some examples.

> Maybe all we need to do here is iterate over `FlowExecutionList` and nothing else.

Fine to go through https://github.com/jenkinsci/workflow-job-plugin/blob/5be9183b6fe8c89f002a1e64d9cc8ae96ef3a1ea/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L565-L567 so long as your iteration of `FlowExecutionList` forces `Run`s to be loaded.

dwnusbaum: > Potentially a variety of things. Put in a Thread.dumpStackTrace(); and you will see some examples.

Yeah that's what I did, and it turns out that even running `mvn clean test -Dtest=blah`, `FlowExecutionList$ItemListenerImpl` was still around somehow and being called. I added the class back but deleted the method contents to try to invalidate whatever cached artifact was causing issues and then the Pipeline was not loaded after the restart.

From some testing it looks like iterating through `FlowExecutionList` is sufficient because `FlowExecutionList.iterator()` internally calls `FlowExecutionOwner.get()` on each execution, which for `WorkflowRun$Owner` calls `WorkflowRun$Owner.run()`, which loads the build.

jglick: Sounds right. Just recommend making sure some test fails when the iteration is commented out, so you can be sure that you are not accidentally relying on an unrelated build load call stack.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
dwnusbaum: Added in b3ad2d4.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
dwnusbaum: Added back in bdcda5e with a new test that times out if the build isn't loaded.

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
dwnusbaum: Something was wrong with my test environment, see https://github.com/jenkinsci/workflow-api-plugin/pull/178#discussion_r743709072.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -169,31 +170,11 @@ public static FlowExecutionList get() {
      */
     @Extension
     public static class ItemListenerImpl extends ItemListener {
-        @Inject
```
jglick, line 172: check for unused imports

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -252,4 +233,43 @@ public void onFailure(Throwable t) {
...
+    @Extension
+    public static class ResumeStepExecutionListener extends FlowExecutionListener {
+        @Override
+        public void onResumed(@Nonnull FlowExecution e) {
+            Futures.addCallback(e.getCurrentExecutions(false), new FutureCallback<List<StepExecution>>() {
+                @Override
+                public void onSuccess(List<StepExecution> result) {
+                    FlowExecutionList list = FlowExecutionList.get();
+                    FlowExecutionOwner owner = e.getOwner();
+                    if (!list.runningTasks.contains(owner)) {
```
jglick, line 252: Maybe check `isComplete` first?

jglick (approved):
Looking right.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
dwnusbaum: Yeah I was thinking about that as well. `CpsFlowExecution.isComplete` is checked in `CpsFlowExecution.onLoad` before `loadProgramAsync` gets called, but `WorkflowRun.onLoad` only checks `CpsFlowExecution.isComplete()` if `WorkflowRun.completed` is `null`, which I think should only be true for old Pipelines from before `WorkflowRun.completed` was added (which seems wrong, probably better to always check in case there is a mismatch between the serialized state of the execution and the run, so I will look at that in a bit).

I also wasn't sure if it would be possible for there to be a timing issue, for example if the pipeline was suspended somewhere where it doesn't need any additional input to proceed and so the CPS VM thread finishes the build while we are on another thread still in `WorkflowRun.onLoad`.

jglick: > if it would be possible for there to be a timing issue

Almost certainly.

dwnusbaum: > Almost certainly.

If this is the case though, then we cannot fix the issue here, because there is no synchronization that can guarantee that the build does not complete right after we check `isComplete` but before we do anything else. Maybe the CPS VM thread should not attempt to execute anything until `WorkflowRun.onLoad` finishes or something like that.

jglick: What makes you think the observed problem is the result of a race condition?

dwnusbaum: I'm not following your question. I just mean that it is not clear to me whether checking `isComplete` here actually gives us any useful guarantees, partly because `WorkflowRun.onLoad` is not guaranteed to check it and partly because the execution actually completing will happen on the CPS VM thread, which is unrelated to the thread calling `WorkflowRun.onLoad`. I don't necessarily think that the underlying reported problem is a race condition.

dwnusbaum: When I said

> then we cannot fix the issue here,

I just meant that we cannot fix any timing issues related to checking `isComplete` or not. Maybe that was not clear.

dwnusbaum: Ah ok, so reading through `WorkflowRun.onLoad` again, it _is_ guaranteed to call `isComplete` inside of `WorkflowRun.getExecution` [here](https://github.com/jenkinsci/workflow-job-plugin/blob/1f016e0f8b4c812b63b40d6d15d70ffc20a94ea5/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L701-L704), and if the execution is complete, it marks the `WorkflowRun` as complete (this was the fix for the symptoms of [JENKINS-50199](https://issues.jenkins.io/browse/JENKINS-50199)), so [this condition](https://github.com/jenkinsci/workflow-job-plugin/blob/1f016e0f8b4c812b63b40d6d15d70ffc20a94ea5/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L565) would be false and so `FlowExecutionListener.fireResumed` would be skipped along with `ResumeStepExecutionListener.onResumed` if the execution was marked as complete in its serialized form.

That leaves us with a potential case where `WorkflowRun.getExecution` loads and resumes the execution, which has not yet completed, but then the execution completes before `ResumeStepExecutionListener.onResumed` has been called by `WorkflowRun.onLoad` after `WorkflowRun.getExecution` and so we resume the steps unnecessarily. We could check `isComplete` here in case this happens, but it seems somewhat unlikely, and the issue is even if we do check it, the execution could complete right after the check before `ResumeStepExecutionListener.onResumed` attempts to resume the step executions, so we wouldn't really be guaranteed to prevent the problem. We could try to modify execution resumption to somehow only start after `WorkflowRun.onLoad` returns, but that seems like a somewhat risky change unless we have some evidence that it is a problem in practice.

I think I'll go ahead and add the `isComplete` check since it won't hurt and could potentially help in some case, but I'll also add a comment mentioning that execution is asynchronous and so there could still be problems.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
dwnusbaum: I removed another unnecessary usage of `@Inject` along with the import in ec57f3d.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
dwnusbaum: Check added with a comment in e9f2421.

dwnusbaum:
Once the build completes I will do some downstream testing, probably just with a draft PR against https://github.com/jenkinsci/bom.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -30,6 +29,7 @@
 import java.util.logging.Logger;
 
 import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
```
jglick, line 32: FYI this package is deprecated.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -252,4 +229,53 @@ public void onFailure(Throwable t) {
...
+                        // That said, there is nothing to prevent the execution from completing right after we check
+                        // isComplete. If we want to fully prevent that, we would need to delay actual execution
+                        // resumption until WorkflowRun.onLoad completes or add some form of synchronization.
+                        return;
+                    }
+                    FlowExecutionList list = FlowExecutionList.get();
+                    FlowExecutionOwner owner = e.getOwner();
+                    if (!list.runningTasks.contains(owner)) {
+                        LOGGER.log(Level.WARNING, "Resuming {0}, which is missing from FlowExecutionList ({1}), so registering it now.",
+                                new Object[] {owner, list.runningTasks.getView()});
```
jglick, line 260: Note that this is one of the unusual cases where the `Supplier` overload to `Logger.warning` would be inappropriate since the call to `getView` must happen now for the message to be meaningful.

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
```diff
@@ -79,4 +99,113 @@
         });
     }
 
+    @Test public void forceLoadRunningExecutionsAfterRestart() throws Throwable {
```
jglick, line 102: Any associated `@Issue` at this point?

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
```diff
@@ -79,4 +99,113 @@
...
+                at org.jenkinsci.plugins.workflow.job.WorkflowRun$Owner.run(WorkflowRun.java:937)
+                at org.jenkinsci.plugins.workflow.job.WorkflowRun$Owner.get(WorkflowRun.java:948)
+                at org.jenkinsci.plugins.workflow.flow.FlowExecutionList$1.computeNext(FlowExecutionList.java:65)
+                at org.jenkinsci.plugins.workflow.flow.FlowExecutionList$1.computeNext(FlowExecutionList.java:57)
+                at com.google.common.collect.AbstractIterator.tryToComputeNext(AbstractIterator.java:143)
+                at com.google.common.collect.AbstractIterator.hasNext(AbstractIterator.java:138)
+                at org.jenkinsci.plugins.workflow.flow.FlowExecutionList$ItemListenerImpl.onLoaded(FlowExecutionList.java:175)
+                at jenkins.model.Jenkins.<init>(Jenkins.java:1019)
+            */
+            waitFor(logging::getMessages, hasItem(containsString("Will resume [org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep")));
```
jglick, line 135: or https://javadoc.jenkins.io/component/jenkins-test-harness/org/jvnet/hudson/test/LoggerRule.html#recorded-org.hamcrest.Matcher- if that is helpful

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
```diff
@@ -79,4 +99,113 @@
...
+            public String getFunctionName() {
+                return "noResume";
+            }
+        }
+    }
+
+    /**
+     * Wait up to 5 seconds for the given supplier to return a matching value.
+     */
+    private static <T> void waitFor(Supplier<T> valueSupplier, Matcher<T> matcher) throws InterruptedException {
```
jglick, line 203: or add a test dep on Awaitility

jglick (approved):
I believe this should be safe.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
dwnusbaum: Oops, yeah I will replace it.

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
dwnusbaum: Not yet, I will double-check for existing issues tomorrow and file something linked to JENKINS-43587.

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
dwnusbaum: For something small and one-off like this I prefer to avoid adding a new dependency.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
dwnusbaum: Fixed in e56d14e.

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
dwnusbaum: I filed [JENKINS-67164](https://issues.jenkins.io/browse/JENKINS-67164). I only added `@Issue("JENKINS-67164")` to the other test though, since this one just makes sure that `FlowExecutionList` resumes Pipelines after Jenkins restarts and isn't directly related to the issue.

car-roll (approved):

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -252,4 +229,53 @@ public void onFailure(Throwable t) {
...
+                @Override
+                public void onFailure(Throwable t) {
+                    if (t instanceof CancellationException) {
+                        LOGGER.log(Level.FINE, "Cancelled load of " + e, t);
+                    } else {
+                        LOGGER.log(Level.WARNING, "Failed to load " + e, t);
+                    }
+                }
+
+            }, MoreExecutors.directExecutor()); // TODO: Unclear if we need to run this asynchronously or if StepExecution.onResume has any particular thread requirements.
```
dwnusbaum, line 278: It looks like running this synchronously caused [JENKINS-67351](https://issues.jenkins.io/browse/JENKINS-67351). I will look into switching to using a different thread.

PR #185, opened 2021-12-08
Code clean-up

offa:
Some code clean-up and removal of JSR 305.

----------------------------

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
@@ -109,7 +109,7 @@ public abstract class FlowExecution implements FlowActionStorage, GraphLookupVie
     /**
      * Should be called by the flow owner after it is deserialized.
      */
-    public /*abstract*/ void onLoad(FlowExecutionOwner owner) throws IOException {
+    public /*abstract*/ void onLoad(FlowExecutionOwner owner) {
         if (Util.isOverridden(FlowExecution.class, getClass(), "onLoad")) {
             onLoad();
         }
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
@@ -367,11 +367,8 @@ public class StashManager {
                 Map<String,String> files = new HashMap<>();
                 for (String path : srcroot.list("**/*", null, false)) {
                     files.put(path, path);
-                    InputStream in = srcroot.child(path).open();
-                    try {
+                    try(InputStream in = srcroot.child(path).open()) {
                         dstDir.child(path).copyFrom(in);
-                    } finally {
-                        IOUtils.closeQuietly(in);
                     }
                 }
                 if (!files.isEmpty()) {
--- a/src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
@@ -58,6 +58,7 @@ import hudson.util.StreamTaskListener;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
+import java.nio.charset.Charset;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.logging.Level;
@@ -401,7 +402,7 @@ public class ArtifactManagerTest {
             assertEquals(contents.length(), f.length());
             assertThat(f.lastModified(), not(is(0)));
             try (InputStream is = f.open()) {
-                assertEquals(contents, IOUtils.toString(is));
+                assertEquals(contents, IOUtils.toString(is, Charset.defaultCharset()));
             }
             URL url = f.toExternalURL();
             if (url != null) {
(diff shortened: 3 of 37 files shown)


offa:
For someone more experienced in the subject: Are adding `final` and using lambdas instead of anonymous classes safe here?

jglick:
> Are adding `final` and using lambdas instead of anonymous classes safe here?

In general, yes. You need to be careful with any classes that might be serialized.

#190 should cover most of the diff here. Sorry about the PR review/merge backlog.

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -367,11 +367,8 @@ public boolean shouldClearAll(@NonNull Run<?,?> build) {
                 Map<String,String> files = new HashMap<>();
                 for (String path : srcroot.list("**/*", null, false)) {
                     files.put(path, path);
-                    InputStream in = srcroot.child(path).open();
-                    try {
+                    try(InputStream in = srcroot.child(path).open()) {
                         dstDir.child(path).copyFrom(in);
-                    } finally {
-                        IOUtils.closeQuietly(in);
```
jglick, line 374: Fix imports?

src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
```diff
@@ -401,7 +402,7 @@ private void assertFile(VirtualFile f, String contents) throws Exception {
             assertEquals(contents.length(), f.length());
             assertThat(f.lastModified(), not(is(0)));
             try (InputStream is = f.open()) {
-                assertEquals(contents, IOUtils.toString(is));
+                assertEquals(contents, IOUtils.toString(is, Charset.defaultCharset()));
```
jglick, line 405: Use UTF-8 here. (Should be irrelevant, since the Unicode tests we are doing involve file _names_ not contents, but as a matter of style we try to use UTF-8 unless there is a specific reason like external process output why we would expect something to be in another encoding.)

src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
```diff
@@ -416,7 +417,7 @@ private void assertFile(VirtualFile f, String contents) throws Exception {
                 this.u = u;
             }
             @Override public String call() throws IOException {
-                return IOUtils.toString(u);
+                return IOUtils.toString(u, Charset.defaultCharset());
```
jglick, line 420: ditto

src/test/java/org/jenkinsci/plugins/workflow/flow/DurabilityBasicsTest.java
```diff
@@ -84,7 +85,7 @@ public void managePermissionShouldAccessGlobalConfig() throws Throwable {
 
             try (ACLContext c = ACL.as(User.getById(USER, true))) {
                 Collection<Descriptor> descriptors = Functions.getSortedDescriptorsForGlobalConfigUnclassified();
-                assertTrue("Global configuration should not be accessible to READ users", descriptors.size() == 0);
+                assertEquals("Global configuration should not be accessible to READ users", 0, descriptors.size());
```
jglick, line 88: or probably something like (plus imports)
```suggestion
                assertThat("Global configuration should not be accessible to READ users", descriptors, empty());
```

jglick (approved):
Looks good, thanks! Left a few very minor comments—feel free to ignore. Sorry about the long delay and the merge conflicts.

offa:
No problem. Thanks for the suggestions. :+1:

src/test/java/org/jenkinsci/plugins/workflow/flow/DurabilityBasicsTest.java
offa: Good catch!

src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
jglick: I did not check, I just guessed that `IOUtils` became unused.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
```diff
@@ -109,7 +109,7 @@ public FlowDurabilityHint getDurabilityHint() {
     /**
      * Should be called by the flow owner after it is deserialized.
      */
-    public /*abstract*/ void onLoad(FlowExecutionOwner owner) throws IOException {
```
dwnusbaum, line 112: Changes to `throws` clauses like this break source compatibility with `CpsFlowExecution` and need to be reverted:

```
-------------------------------------------------------------
COMPILATION ERROR : 
-------------------------------------------------------------
workflow-cps-plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java:[721,17] onLoad(org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner) in org.jenkinsci.plugins.workflow.cps.CpsFlowExecution cannot override onLoad(org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner) in org.jenkinsci.plugins.workflow.flow.FlowExecution
  overridden method does not throw java.io.IOException
1 error
-------------------------------------------------------------
```

jglick: Indeed this is wrong and should be reverted.

PR #187, opened 2021-12-14
[JENKINS-40161] Handle exceptions from `StepExecution.onResume`

jglick:
Preparation for https://issues.jenkins.io/browse/JENKINS-40161. Cannot safely add `throws Exception` to the method signature, or start letting implementations throw exceptions, until most users have updated to get this code, so might as well do it now rather than later; at least will already handle `RuntimeException` and `Error`.

Not really related to #178, which just moved this block of code around without changing what it did when called.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
@@ -262,7 +262,11 @@ public class FlowExecutionList implements Iterable<FlowExecution> {
                     }
                     LOGGER.log(Level.FINE, "Will resume {0}", result);
                     for (StepExecution se : result) {
-                        se.onResume();
+                        try {
+                            se.onResume();
+                        } catch (Throwable x) {
+                            se.getContext().onFailure(x);
+                        }
                     }
                 }
 


car-roll (approved):

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -262,7 +262,11 @@ public void onSuccess(List<StepExecution> result) {
                     }
                     LOGGER.log(Level.FINE, "Will resume {0}", result);
                     for (StepExecution se : result) {
-                        se.onResume();
+                        try {
+                            se.onResume();
+                        } catch (Throwable x) {
+                            se.getContext().onFailure(x);
```
dwnusbaum, line 268: Should we return or exit the loop here?

jglick: I do not think so. This step failed to resume, but others might resume fine, and there might be `retry` or other error handlers in place.

Also here I am just matching the established idiom of an `onResume` implementation doing a bunch of stuff all wrapped in a `try`-`catch` block.

dwnusbaum: Oh, yeah for some reason I was thinking we were failing the Pipeline here, not the step.

dwnusbaum (approved):

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
dwnusbaum, line 268: Looking at this again, it might be better to use `se.stop` instead, although I'm not sure that it matters for any existing steps besides `node`. Compare to semi-related changes in https://github.com/jenkinsci/workflow-cps-plugin/pull/780 and https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/344.

PR #193, opened 2022-01-09
Provide Maven wrapper which is more convenient for local development

krichter722:

Providing the wrapper allows a developer to build the project without the need to manage different maven installations. I comes with the downside of a large git checkout. At least Jenkins won't notice it because it handle fetches well.

Unfortunately the Wrapper won't be used by the CI which would be an interesting change, especially if `buildPlugin` in `Jenkinsfile` would detect `mvnw` and use it for the build rather than the provided Maven installation.


jglick:
We do not generally do this in @jenkinsci repos. Would be a lot of work to maintain (not covered by Dependabot AFAIK), and `Jenkinsfile` / `jenkins-infra/pipeline-library` is not set up to expect it.

PR #195, opened 2022-01-10
Update parent, baseline, BOM

jglick:
Supersedes #194.


basil (approved):

jglick:
Unfortunately @basil

> Merging can be performed automatically with 1 approving review.

you do not count so far as GH is concerned. Maybe you can get yourself added to various plugins as comaintainer?

car-roll (approved):

PR #201, opened 2022-02-16
`onLoad(FlowExecutionOwner)` is meant to throw `IOException`

jglick:
Reverting an incorrect change (I suppose caused by unvalidated use of some refactoring tool by @offa) from #185.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
@@ -109,7 +109,7 @@ public abstract class FlowExecution implements FlowActionStorage, GraphLookupVie
     /**
      * Should be called by the flow owner after it is deserialized.
      */
-    public /*abstract*/ void onLoad(FlowExecutionOwner owner) {
+    public /*abstract*/ void onLoad(FlowExecutionOwner owner) throws IOException {
         if (Util.isOverridden(FlowExecution.class, getClass(), "onLoad")) {
             onLoad();
         }


dwnusbaum:
I think `FlowExecutionOwner.getListener` also needs to have `throws IOException` added back for `WorkflowRun.Owner.getListener`.

jglick:
> I think `FlowExecutionOwner.getListener` also needs to have `throws IOException` added back

https://github.com/jenkinsci/workflow-api-plugin/blob/7f5f1759dc1635f4e704d26e7bcba37a20e59c4b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionOwner.java#L128 is still OK. https://github.com/jenkinsci/workflow-api-plugin/blob/7f5f1759dc1635f4e704d26e7bcba37a20e59c4b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionOwner.java#L179 is just a private impl, should not be a problem.

jglick:
@offa please disable whichever IDE suggestion you may have to automatically strip unthrown exceptions as this is not safe to use without human inspection in general.

dwnusbaum (approved):

basil:
> @offa please disable whichever IDE suggestion you may have to automatically strip unthrown exceptions as this is not safe to use without human inspection in general.

Yeah, I tend to ignore this IDE suggestion in general. The one case where I have found it useful is for removing e.g. `throws Exception` from test methods that don't need it, but even in that case the value of the suggestion seems dubious.

PR #202, opened 2022-02-21
Update `<select>` config appearance 

janfaracik:
This PR updates the `<select>` config appearance to use the new classes introduced in [Jenkins #5923](https://github.com/jenkinsci/jenkins/pull/5923).

**Before**
image

**After**
image

Just a heads up that the pre-existing `tooltip` attribute seemingly doesn't do anything on `<option>`, tried it on Safari and Chrome.

Cheers.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/global.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/global.jelly
@@ -2,16 +2,17 @@
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
     <f:section title="Pipeline Speed/Durability Settings">
         <f:entry title="${%Pipeline Default Speed/Durability Level}" field="durabilityHint">
-            <select class="setting-input" name="durabilityHint">
-                <f:option value="null" selected="${instance[field] == null}">None: use pipeline default (<j:out value="${descriptor.suggestedDurabilityHint.name()}" />)</f:option>
-                <j:forEach var="it" items="${descriptor.durabilityHintValues}">
-                    <j:set var="optionBody" encode="false">${it.description}</j:set>
-                    <option value="${it.name()}"
-                            selected="${(it==instance[field]) ? 'true':null}"
-                            tooltip="${it.tooltip}"><j:out value="${optionBody}"/></option>
-                </j:forEach>
-            </select>
+            <div class="jenkins-select">
+                <select class="jenkins-select__input" name="durabilityHint">
+                    <f:option value="null" selected="${instance[field] == null}">None: use pipeline default (<j:out value="${descriptor.suggestedDurabilityHint.name()}" />)</f:option>
+                    <j:forEach var="it" items="${descriptor.durabilityHintValues}">
+                        <j:set var="optionBody" encode="false">${it.description}</j:set>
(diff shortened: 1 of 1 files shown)


olamy:
this will need to upgrade the pom value `<jenkins.version>2.249.1</jenkins.version>` to a non LTS version.

janfaracik:
> this will need to upgrade the pom value `<jenkins.version>2.249.1</jenkins.version>` to a non LTS version.

Apologies, bit of a novice in this area. Is this still necessary if this PR still works/performs the same way on current LTS Jenkins?

olamy:
> > this will need to upgrade the pom value `<jenkins.version>2.249.1</jenkins.version>` to a non LTS version.
> 
> Apologies, bit of a novice in this area. Is this still necessary if this PR still works/performs the same way on current LTS Jenkins?

no worries. 
if this need too much unmaintainable code for a limited period of time not sure it worth it.
there are usually some rules https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/
I can see some plugins are upgrading to some non LTS but depends what others contributors think about this.

janfaracik:
> > > this will need to upgrade the pom value `<jenkins.version>2.249.1</jenkins.version>` to a non LTS version.
> > 
> > 
> > Apologies, bit of a novice in this area. Is this still necessary if this PR still works/performs the same way on current LTS Jenkins?
> 
> no worries. if this need too much unmaintainable code for a limited period of time not sure it worth it. there are usually some rules https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/ I can see some plugins are upgrading to some non LTS but depends what others contributors think about this.

Any thoughts on this @timja?

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/global.jelly
```diff
@@ -2,16 +2,17 @@
...
-                <f:option value="null" selected="${instance[field] == null}">None: use pipeline default (<j:out value="${descriptor.suggestedDurabilityHint.name()}" />)</f:option>
-                <j:forEach var="it" items="${descriptor.durabilityHintValues}">
-                    <j:set var="optionBody" encode="false">${it.description}</j:set>
-                    <option value="${it.name()}"
-                            selected="${(it==instance[field]) ? 'true':null}"
-                            tooltip="${it.tooltip}"><j:out value="${optionBody}"/></option>
-                </j:forEach>
-            </select>
+            <div class="jenkins-select">
+                <select class="jenkins-select__input" name="durabilityHint">
```
timja, line 6: This change appears to be backwards and forwards compatible
```suggestion
                <!-- TODO Remove setting-input when baseline above 2.335 -->
                <select class="jenkins-select__input setting-input" name="durabilityHint">
```

timja (commented):
@janfaracik Jenkins plugins use the version defined in the `jenkins.version` as the minimum version of Jenkins it's compatible with.

Historically maintainers did not like bumping it too high as people not up to date wouldn't get updates.
That's changed a bit in the last year or two, people seem to have decided that the people updating will get updates and critical fixes can always be backported to the old line.

Basically either you need to make sure it works on old versions of core and new versions.
Either by applying both stylings or version / feature detection

Or bump the minimum version up.

This plugin could certainly do with a minimum version bump but given how trivial it is to keep compatibility probably doesn't need to be done in this PR

timja:
Thinking aloud is there not a tag built into Jenkins for this?

Why is this writing plain HTML ?

timja:
see https://github.com/jenkinsci/workflow-api-plugin/pull/205

janfaracik:
> see #205

#205 is infinitely better than mine - good job! I'll close this one.

PR #205, opened 2022-03-01
Replace handbuilt html with jelly for flow durability

timja:
The downside I can see with this is that tooltips are no longer shown.

But tooltips on a drop down are really weird. So not sure if it's the worst idea to get rid of this.

alternative to https://github.com/jenkinsci/workflow-api-plugin/pull/202

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/global.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/global.jelly
@@ -2,15 +2,7 @@
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
     <f:section title="Pipeline Speed/Durability Settings">
         <f:entry title="${%Pipeline Default Speed/Durability Level}" field="durabilityHint">
-            <select class="setting-input" name="durabilityHint">
-                <f:option value="null" selected="${instance[field] == null}">None: use pipeline default (<j:out value="${descriptor.suggestedDurabilityHint.name()}" />)</f:option>
-                <j:forEach var="it" items="${descriptor.durabilityHintValues}">
-                    <j:set var="optionBody" encode="false">${it.description}</j:set>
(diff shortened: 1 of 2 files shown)


janfaracik (approved):

timja:
> > But tooltips on a drop down are really weird. So not sure if it's the worst idea to get rid of this.
> 
> Maybe it's place a help button below the dropdown you can hover over to get some more intel about the mode selected? Could be beneficial for less versed pipeline users.

Can’t see how that would look good, there’s already some help for the input itself which should answer most questions

jglick:
You can use `FormValidation.ok` (or `.okWithHtml`) to offer some descriptions of the current selection.

jglick:
Merge conflicts @timja.

Do you have time to try https://github.com/jenkinsci/workflow-api-plugin/pull/205#issuecomment-1113625208? If not, I guess this is OK as is.

src/main/resources/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel/global.jelly
```diff
@@ -2,15 +2,7 @@
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
     <f:section title="Pipeline Speed/Durability Settings">
         <f:entry title="${%Pipeline Default Speed/Durability Level}" field="durabilityHint">
-            <select class="setting-input" name="durabilityHint">
-                <f:option value="null" selected="${instance[field] == null}">None: use pipeline default (<j:out value="${descriptor.suggestedDurabilityHint.name()}" />)</f:option>
-                <j:forEach var="it" items="${descriptor.durabilityHintValues}">
```
jglick, line 7: These methods could now be made private, I think, or simply inlined: https://github.com/jenkinsci/workflow-api-plugin/blob/786bbb37b24c7b0393f72a602547a8a85640d3e8/src/main/java/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel.java#L82-L88

jglick (commented):
I also just noticed that this whole thing should have been extending `GlobalConfiguration`. Probably too late to change that compatibly, though.

jglick (approved):

PR #198, opened 2022-01-14
Revert fixes related to JENKINS-67351 and JENKINS-67164

dwnusbaum:
Reverts #188, #187, and #178 due to problems in https://github.com/jenkinsci/workflow-api-plugin/pull/188 (see https://github.com/jenkinsci/workflow-api-plugin/pull/188#issuecomment-1013382364) which have no obvious fix.

The `FlowExecutionList.isResumptionComplete` API from #188 is preserved to avoid binary compatibility issues with `workflow-durable-task-step` prior to https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/197.

These fixes were generally expected only to apply to uncommon scenarios related to disaster recovery, so I think it is ok to revert them for now.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #220, opened 2022-05-09
[JENKINS-40161] Handle exceptions from `StepExecution.onResume`

jglick:
Reapplying #187, which happened to have gotten caught up in some barely related flow control changes and reverted in #198.


dwnusbaum (approved):
oops, sorry!

PR #221, opened 2022-05-11
[JENKINS-67164] Call `StepExecution.onResume` in response to `WorkflowRun.onLoad` not `FlowExecutionList.ItemListenerImpl`

jglick:
[JENKINS-67164](https://issues.jenkins.io/browse/JENKINS-67164): Reapplies #178 and its amendment #188, reverting the reversion #198.

Fixes the problem identified in `SSHAgentStepWorkflowTest.sshAgentAvailableAfterRestart` whereby a `sh` step began running while an `sshagent` step was still in the middle of an `onResume`, by introducing a `FlowExecution.afterStepExecutionsResumed` hook which `CpsFlowExecution` will implement to unpause the program near the end of `WorkflowRun.onLoad`, rather than unpausing immediately upon deserializing `program.dat`.

Seems to offer a nicer way to implement the `ExecutorPickle` replacement portion of https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/180, by letting `ExecutorStepExecution.onResume` wait for an agent to reconnect before resuming progran execution. Compared to `ExecutorPickle` the advantage is that if there is in fact a timeout, rather than marking the entire program as a failure to load, the `node` step can throw a proper Groovy-level exception which can be handled gracefully.

A matching PR to `workflow-cps-plugin` is needed to avoid the class of regression seen with `sshagent`. ~I would like to run some integration tests in advance.~ https://github.com/jenkinsci/bom/pull/1122#issuecomment-1124901390

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
@@ -23,13 +23,20 @@ import org.jenkinsci.plugins.workflow.steps.StepExecutionIterator;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+import org.jenkinsci.plugins.workflow.graphanalysis.LinearBlockHoppingScanner;
 
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.Beta;
@@ -259,29 +266,19 @@ public class FlowExecutionList implements Iterable<FlowExecution> {
             Futures.addCallback(e.getCurrentExecutions(false), new FutureCallback<List<StepExecution>>() {
                 @Override
                 public void onSuccess(@NonNull List<StepExecution> result) {
-                    try {
-                        if (e.isComplete()) {
-                            // WorkflowRun.onLoad will not fireResumed if the execution was already complete when loaded,
-                            // and CpsFlowExecution should not then complete until afterStepExecutionsResumed, so this is defensive.
-                            return;
-                        }
-                        FlowExecutionList list = FlowExecutionList.get();
-                        FlowExecutionOwner owner = e.getOwner();
-                        if (!list.runningTasks.contains(owner)) {
-                            LOGGER.log(Level.WARNING, "Resuming {0}, which is missing from FlowExecutionList ({1}), so registering it now.", new Object[] {owner, list.runningTasks.getView()});
-                            list.register(owner);
-                        }
-                        LOGGER.log(Level.FINE, "Will resume {0}", result);
-                        for (StepExecution se : result) {
-                            try {
-                                se.onResume();
-                            } catch (Throwable x) {
-                                se.getContext().onFailure(x);
-                            }
-                        }
-                    } finally {
-                        e.afterStepExecutionsResumed();
+                    if (e.isComplete()) {
+                        // WorkflowRun.onLoad will not fireResumed if the execution was already complete when loaded,
+                        // and CpsFlowExecution should not then complete until afterStepExecutionsResumed, so this is defensive.
+                        return;
                     }
+                    FlowExecutionList list = FlowExecutionList.get();
+                    FlowExecutionOwner owner = e.getOwner();
+                    if (!list.runningTasks.contains(owner)) {
+                        LOGGER.log(Level.WARNING, "Resuming {0}, which is missing from FlowExecutionList ({1}), so registering it now.", new Object[] {owner, list.runningTasks.getView()});
+                        list.register(owner);
+                    }
+                    LOGGER.log(Level.FINE, "Will resume {0}", result);
+                    new ParallelResumer(result, e::afterStepExecutionsResumed).run();
                 }
 
                 @Override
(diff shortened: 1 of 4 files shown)


dwnusbaum (approved):
Seems fine to me. For what it's worth, I would have preferred to see the old commits cherry-picked or a revert-the-revert commit with your changes applied as a new commit to make it easier to tell what the new changes are, although I guess the followup commits in #198 would have created a bunch of conflicts.

jglick:
> a revert-the-revert commit with your changes applied as a new commit

Agreed, though it would have been tricky, since I had to play with various uncommitted edits before I found something that worked.

car-roll (approved):

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -294,7 +291,90 @@ public void onFailure(@NonNull Throwable t) {
...
-            }, Timer.get()); // We always hold RunMap and WorkflowRun locks here, so we resume steps on a different thread to avoid potential deadlocks. See JENKINS-67351.
+            }, MoreExecutors.directExecutor());
         }
     }
+
+    /** Calls {@link StepExecution#onResume} for each step in a running build.
+     * Does so in parallel, but always completing enclosing blocks before the enclosed step.
+     * A simplified version of https://stackoverflow.com/a/67449067/12916, since this should be a tree not a general DAG.
+     */
+    private static final class ParallelResumer {
```
dwnusbaum, line 302: Is this for correctness or an optimization?

jglick: It is an optimization. Previously all steps were resumed serially in a topological order. (Or I hope `getCurrentExecutions` enforced a topological order; the sorting by `CpsThread` is a bit opaque to me.) The problem in https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/180 arises when you have a bunch of parallel branches all running `node` and all of them are going to fail due to missing agents. Previously, Jenkins would wait 5m for the first agent, report that it was gone, then wait 5m for the next agent, etc. Now all the `node` blocks are resumed in parallel, followed by the `sh` steps inside them, etc. Have a test demonstrating this in `workflow-durable-task-step` but only now got an incremental build from https://github.com/jenkinsci/workflow-cps-plugin/pull/534 that I needed.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -256,4 +252,129 @@ public void onFailure(@NonNull Throwable t) {
...
+                    ready.put(n, entry.getValue());
+                }
+            }
+            LOGGER.fine(() -> "Ready to resume: " + ready);
+            nodes.keySet().removeAll(ready.keySet());
+            for (Map.Entry<FlowNode, StepExecution> entry : ready.entrySet()) {
+                FlowNode n = entry.getKey();
+                StepExecution exec = entry.getValue();
+                processing.add(n);
+                Timer.get().submit(() -> {
```
jglick, line 362: Probably should not use `Timer` for this, as it is throttled to 10 threads. Could use `threadPoolForRemoting`, or expose `SynchronousNonBlockingStepExecution.executorService` (see https://github.com/jenkinsci/workflow-step-api-plugin/pull/110).

jglick:
I believe this is ready to release **if** https://github.com/jenkinsci/workflow-cps-plugin/pull/534 can be released at the same time (to avoid regressions in some corner cases). https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/226 can then also be released (but there is no rush in doing so).

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
dwnusbaum: Ok, I guess it's kind of unusual for `ExecutorStepExecution.onResume` to block if no other steps do so, but looking through https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/180 it seems like you already tried various alternatives, so this seems fine.

dwnusbaum (approved):

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
jglick: Basically this is the replacement for the `TryRepeatedly` pickle. I played with a lot of alternatives indeed. I am not claiming this approach is ideal but it seems to be straightforward enough and do the job.

Earlier, before hitting on the idea of having `onResume` block, I had hoped to arrange it so that the program would resume right away, and then a `sh` step running on the dead agent would eventually figure out there was no hope and abort. I ran into problems in functional tests, though, things like

```groovy
node('agent') {
  input 'Proceed?' // restart here
  sh 'true'
}
```

The program would run past `input` to the `sh` step, which would then either fail with a `MissingContextVariableException` (confusing) or block the CPS VM thread inside `DSL` (which will throw an error if it blocks >5s). Admittedly this is an artificial case (you should not hold an executor like that) but a lot of tests were failing that I had to work around in artificial ways and it seemed uncomfortable. It would have been possible to fix all these cases but only in pretty intrusive ways: by defining a new `StepContext` object in lieu of `FilePath` (perhaps `DynamicFilePathContext.Representation`) which could be deserialized without pickles, but every step currently expecting `FilePath` would need to be adjusted to expect this instead, and then the step would need to have complicated logic to wait for the new contextual object to be ready (translatable to a live `FilePath`) before doing anything useful. 

The compromise I settled on is closer to the original pickle behavior in that CPS code and steps do not run until an attempt is made to restore the original context: until all open `node` blocks have gotten their agent to reconnect, or timed out trying. The difference is only that in case this fails (with a timeout), the error is thrown out of `StepExecution.onResume` and thus `StepContext.onFailure` so it becomes a properly modeled `Throwable` with a CPS VM stack trace that we can catch and handle—unlike the original situation where the entire build would be effectively hard-killed with no possible cleanup.

PR #224, opened 2022-05-16
Use proper CSS class for build status icon

NotMyFault:
As opposed to other locations in Jenkins, the build status isn't wrapped in the proper class, which leads to a false rendering:

The change proposed takes care of that and ensures the format is no longer out of style:

cc @car-roll


jglick (approved):
Works as advertised.

PR #273, opened 2023-01-18
Require Java 11 and update parent POM, BOM, and minimum supported Jenkins version

dwnusbaum:
Replaces #268 and #272.

(I do not particularly care about releasing these changes; I only want to see if `FileLogStorageTest` passes on Windows when running Java 11.)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,4 @@
 buildPlugin(useContainerAgent: true, configurations: [
-  [ platform: "linux", jdk: "8" ],
-  [ platform: "windows", jdk: "8" ],
-  [ platform: "linux", jdk: "11" ]
+  [ platform: "linux", jdk: "17" ],
+  [ platform: "windows", jdk: "11" ]
 ])
--- a/src/test/java/org/jenkinsci/plugins/workflow/log/LogStorageTestBase.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/log/LogStorageTestBase.java
@@ -74,10 +74,6 @@ import org.jvnet.hudson.test.LoggerRule;
  */
 public abstract class LogStorageTestBase {
 
-    static {
-        System.setProperty("line.separator", "\n");
-    }
-
     @ClassRule public static JenkinsRule r = new JenkinsRule();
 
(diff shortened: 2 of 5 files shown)


dwnusbaum:
Indeed it looks like the behavior of setting the system property `line.separator` at runtime changes from Java 8 (new value is used) to Java 11 (new value is ignored). I will try to adapt the tests.

Jenkinsfile
```diff
@@ -1,5 +1,4 @@
 buildPlugin(useContainerAgent: true, configurations: [
-  [ platform: "linux", jdk: "8" ],
-  [ platform: "windows", jdk: "8" ],
-  [ platform: "linux", jdk: "11" ]
+  [ platform: "linux", jdk: "17" ],
+  [ platform: "windows", jdk: "11" ]
```
jglick, line 3: or perhaps more intuitively
```suggestion
  [platform: "linux", jdk: 17],
  [platform: "windows", jdk: 11]
```

jglick (approved):

src/test/java/org/jenkinsci/plugins/workflow/log/LogStorageTestBase.java
```diff
@@ -330,6 +343,13 @@ private AnnotatedLargeText<?> text(String id) {
         return createStorage().stepLog(new MockNode(id), true);
     }
 
+    /**
+     * Concatenate the given lines after interspersing system-dependent line separators between them and adding a final line separator.
+     */
+    private String lines(CharSequence... lines) {
```
dwnusbaum, line 349: Maybe [unlines](https://hackage.haskell.org/package/base-4.17.0.0/docs/Prelude.html#v:unlines) or something else would be a better name.

jglick: `lines` seems intuitive as is.

jglick (approved):

jglick (approved):

dwnusbaum:
@jglick Do you want a release of #274 with the old baseline before I merge this? Or does it not matter?

jglick:
:eyes: three-week-old BOM, probably fine either way.

basil (approved):

PR #284, opened 2023-05-10
Improve consistency of `ErrorAction.findOrigin` across Jenkins restarts

dwnusbaum:
`ErrorAction.findOrigin` currently returns different a `FlowNode` before and after a Jenkins restart if the original error is inside of a branch in a `parallel` step and multiple branches fail (and likely in various other cases). This PR makes it so that the same `FlowNode` is returned regardless of whether Jenkins is restarted (in most cases) by ignoring suppressed exceptions when checking equality.

The problem is that before restarting Jenkins, we are able to simply check reference equality here when finding the origin of an error:

https://github.com/jenkinsci/workflow-api-plugin/blob/0cc7c6e0da9e89b4ed99edf8e830990014fca727/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java#L147

Reference equality is robust against modifications made to the original `Throwable` when it gets thrown to enclosing blocks. For example, calls to `setStackTrace` or `addSuppressedException` do not break reference equality.

However, after restarting Jenkins, we will deserialize new instances of all of the `ErrorAction`s in `FlowNode` XML files, breaking reference equality of the underlying errors, so we have to check types and stack traces:

https://github.com/jenkinsci/workflow-api-plugin/blob/0cc7c6e0da9e89b4ed99edf8e830990014fca727/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java#L149-L153

The stack trace of the serialized errors are frozen at the time they are written to disk. This means that the calls to `setStackTrace` or `addSuppressedException` made after the error is serialized break the ability of `findOrigin` to find the origin as desired.

### Testing done

See new automated tests.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
@@ -32,7 +32,8 @@ import edu.umd.cs.findbugs.annotations.CheckForNull;
 import org.codehaus.groovy.control.MultipleCompilationErrorsException;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import edu.umd.cs.findbugs.annotations.NonNull;
-import hudson.Functions;
+import java.util.Arrays;
+import java.util.Objects;
 import jenkins.model.Jenkins;
 import org.apache.commons.io.output.NullOutputStream;
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/actions/ErrorActionTest.java
```diff
@@ -56,7 +59,7 @@ public class ErrorActionTest {
     public static BuildWatcher buildWatcher = new BuildWatcher();
 
     @Rule
-    public JenkinsRule r = new JenkinsRule();
+    public JenkinsSessionRule rr = new JenkinsSessionRule();
```
dwnusbaum, line 62: I recommend activating "Hide whitespace" when reviewing this file.

src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
```diff
@@ -148,8 +149,28 @@ private static boolean equals(Throwable t1, Throwable t2) {
...
         } else {
-            return Functions.printThrowable(t1).equals(Functions.printThrowable(t2));
+            // TODO: Steps which throw exceptions synchronously on the CPS VM thread break this equality check after
+            // a Jenkins restart. In that case, the original FlowNode gets an ErrorAction with the Throwable, which is
+            // written to disk, and then afterwards, when the error is thrown into the CPS-transformed script to resume
+            // execution, the location in the script and Continuable.SEPARATOR_STACK_ELEMENT are added to the stack
+            // trace of the same Throwable object using setStackTrace, so all subsequent serialized FlowNodes include
+            // those stack trace elements. This breaks stack trace equality, resulting in findOrigin returning the
+            // BlockEndNode of the parent of the step that threw the synchronous error.
+            // The `t1 == t2` reference equality check above masks the issue if you do not restart Jenkins.
```
dwnusbaum, line 162: In particular, this affects the `error` step. I would like to fix this, but I couldn't really think of a straightforward way to do so, other than walking over `getStackTrace` with indices and ignoring missing `Continuable.SEPARATOR_STACK_ELEMENT` and `WorkflowScript.run` stack trace elements.

jglick:
Why `developer`? This seems like a `bug`.

src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
jglick: The only thing I can think of is to inject an additional suppressed exception somewhere with a UUID, and then if two throwables both have that UUID somewhere in their cause/suppressed tree, they must be from a common origin.

jglick (approved):

jglick:
> end users would not care about this directly

Well, it could affect for example the reporting by CloudBees CI of GitHub Checks for a failed `sh` step, though I guess in the somehow contrived case that the `sh` step fails, the build continues for a while, the controller restarts, then the build fails.

src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
dwnusbaum: Hmm, yeah I guess when we create `ErrorAction` (in various places in `workflow-cps`), we could check if there is already a marker exception somewhere in its cause/suppressed tree, and if not, add one. Seems messy though. I will think about it.

jglick: Tip: you can override `fillInStackTrace` to do nothing. Then the suppressed exception would add only one line to `Functions.printThrowable`. (If that is even used; not e.g. for `FlowInterruptedException` or `AbortExeption`.)

PR #287, opened 2023-06-22
Prevent ForkScanner from visiting nodes more than once in some cases for in-progress builds with nested parallelism

dwnusbaum:
In-progress pipelines with nested `parallel` steps sometimes end up visiting the outer `parallel` step start node and all of its parents more than once. In my particular case this breaks a visualization tool that uses `ForkScanner`.

### Testing done

Automated regression test added.

I tested manually against the following plugins to make sure their tests are not affected by the change. I assume though that all uses of `ForkScanner` in these plugins were subject to the same potential problem, and so this PR may affect them even if their tests do not cover in-progress builds with nested parallelism:

* [x] https://github.com/jenkinsci/pipeline-graph-analysis-plugin
* [x] https://github.com/jenkinsci/pipeline-graph-view-plugin
* [x] https://github.com/jenkinsci/blueocean-plugin (specifically `blueocean-pipeline-api-impl` and `blueocean-rest-impl`)

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
@@ -48,6 +48,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * Scanner that will scan down all forks when we hit parallel blocks before continuing, but generally runs in linear order
@@ -443,8 +444,15 @@ public class ForkScanner extends AbstractFlowScanner {
             throw new IllegalStateException("No least common ancestor found from " + heads);
         }
 
-        // If we hit issues with the ordering of blocks by depth, apply a sorting to the parallels by depth
-        return convertForksToBlockStarts(parallelForks);
+        // The result must be in reverse topological order, i.e. inner branches must be visited before outer branches.
+        return convertForksToBlockStarts(parallelForks).stream().sorted((pbs1, pbs2) -> {
+            if (pbs1.forkStart.getEnclosingBlocks().contains(pbs2.forkStart)) {
+                return -1;
+            } else if (pbs2.forkStart.getEnclosingBlocks().contains(pbs1.forkStart)) {
+                return 1;
+            }
+            return 0;
+        }).collect(Collectors.toCollection(ArrayDeque::new));
     }
 
     @Override
--- a/src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
@@ -59,6 +59,9 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
 /**
  * Tests for internals of ForkScanner
  */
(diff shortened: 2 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -992,4 +995,43 @@ public void testParallelCorrectEndNodeForVisitor() throws Exception {
...
+        WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+        SemaphoreStep.waitForStart("outerA/1", b);
+        SemaphoreStep.waitForStart("innerA/1", b);
+        ForkScanner scanner = new ForkScanner();
+        scanner.setup(b.getExecution().getCurrentHeads());
+        List<String> visited = new ArrayList<>();
+        for (FlowNode node : scanner) {
+            visited.add(node.getId());
+        }
+        assertThat("Each FlowNode should be visited only once " + visited, visited.size(), equalTo(new HashSet<>(visited).size()));
```
dwnusbaum, line 1032: Right now, the parallel step start node, as well as the stage body start node, the stage step start node, and the the flow start node (!!) get visited more than once!

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -992,4 +992,39 @@ public void testParallelCorrectEndNodeForVisitor() throws Exception {
...
+                "        }\n" +
+                "      )\n" +
+                "    }\n" +
+                "  )\n" +
+                "}\n", true));
+        WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+        SemaphoreStep.waitForStart("outerA/1", b);
+        SemaphoreStep.waitForStart("innerA/1", b);
+        ForkScanner scanner = new ForkScanner();
+        sanityTestIterationAndVisiter(b.getExecution().getCurrentHeads());
```
dwnusbaum, line 1024: Previously, the parallel step start node (5), as well as the stage body start node (4), the stage step start node (3), and the the flow start node (2) were visited twice.

With the right structure you could end up visiting almost the whole Pipeline twice.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -443,8 +444,15 @@ ArrayDeque<ParallelBlockStart> leastCommonAncestor(@NonNull final Set<FlowNode>
             throw new IllegalStateException("No least common ancestor found from " + heads);
         }
 
-        // If we hit issues with the ordering of blocks by depth, apply a sorting to the parallels by depth
```
dwnusbaum, line 446: I interpreted this as a kind of TODO comment, and debugging showed that in existing test cases with nested parallelism, the result was ordered from innermost to outermost, but the ordering was opposite for my new test, and making the order stable fixes my test, so...

To be clear, I do not fully understand the specifics of this method or really this class, so there could still be other problems, but the current behavior is pretty bad and the new behavior should at least be better in some cases.

car-roll (approved):

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -443,8 +444,15 @@ ArrayDeque<ParallelBlockStart> leastCommonAncestor(@NonNull final Set<FlowNode>
...
-        return convertForksToBlockStarts(parallelForks);
+        // The result must be in reverse topological order, i.e. inner branches must be visited before outer branches.
+        return convertForksToBlockStarts(parallelForks).stream().sorted((pbs1, pbs2) -> {
+            if (pbs1.forkStart.getEnclosingBlocks().contains(pbs2.forkStart)) {
+                return -1;
+            } else if (pbs2.forkStart.getEnclosingBlocks().contains(pbs1.forkStart)) {
+                return 1;
+            }
+            return 0;
```
jglick, line 454: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Comparator.html this is not “consistent with equals”. That is generally harmless when sorting a list (and I suppose also a stream) but I do not think you can reliably implement a topological sort this way. For example in `jshell`

```java
import java.util.stream.*;
Stream.of("axx a ax".split(" ")).sorted((x, y) -> x.startsWith(y) ? 1 : y.startsWith(x) ? -1 : 0).collect(Collectors.joining(" "))
```

works (`a ax axx`) but on the input `axx b a ax` you get back the same list, whereas you might expect `a b ax axx` or similar.

Whether there is any observable bug in _this_ function as a result is another question.

dwnusbaum: Hmm yes, I can look into this.

> Whether there is any observable bug in this function as a result is another question.

Perhaps, if you have more than 3 levels of nested parallel steps, or if you have multiple parallel step children under a single parallel step, although I wonder if the latter case would work regardless of this change because of other issues.

dwnusbaum: From some quick testing, I have only been able to reproduce cases where the outermost parallel step is the second to last element before the sort, when it needs to always be the final element. Other than that, the pre-sort order is fine. Perhaps a more targeted fix could be made further up in the method, and it seems plausible that with some really unusual Pipelines there is some reachable case that I was not able to trigger, but I am not sure.

jglick: Would recommend at least leaving in a code comment warning that this logic would not work for exotic cases which may or may not be reachable.

dwnusbaum: See https://github.com/jenkinsci/workflow-api-plugin/pull/288.

PR #289, opened 2023-07-03
Make ErrorAction.equals a public API

dwnusbaum:
`ErrorAction.findOrigin` is useful, but can be prohibitively slow in complex Pipelines with a lot of errors, as it scans the entire flow graph. In particular, I have a case where I only need to know whether a specific `ErrorAction` represents an original or re-thrown error, and `findOrigin` is currently the only way to get access to the `ErrorId` equality logic.

This PR exposes `ErrorAction.equals` so that other plugins can check error originality using more limited scans in cases where finding the true origin of an error is unnecessary and too slow.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
@@ -43,6 +43,8 @@ import org.jenkinsci.plugins.workflow.flow.FlowExecution;
 import org.jenkinsci.plugins.workflow.graph.AtomNode;
 import org.jenkinsci.plugins.workflow.graph.BlockEndNode;
 import org.jenkinsci.plugins.workflow.graphanalysis.ForkScanner;
+import org.kohsuke.accmod.Restricted;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
```diff
@@ -165,7 +167,8 @@ public String getUrlName() {
      * {@link Throwable#equals} might not be reliable if the program has resumed
      * and stuff is deserialized.
      */
-    private static boolean equals(Throwable t1, Throwable t2) {
+    @Restricted(Beta.class)
+    public static boolean equals(Throwable t1, Throwable t2) {
```
dwnusbaum, line 171: For my particular case, I could instead create a new `public static boolean isRethrown(FlowNode node)` method that checks whether the `ErrorAction` on the associated `FlowNode` was also thrown by one of its immediate parents, but exposing `equals` directly does not seem particularly problematic to me and may be useful in other cases.

joseblas (approved):

jglick (approved):
Makes sense.

PR #290, opened 2023-07-11
Catch errors in `TaskListenerDecorator.decorate` even when wrapped with `merge`

jglick:
See https://github.com/jenkinsci/gradle-plugin/pull/318 for context. Do not risk having a decorator be completely skipped just because an unrelated decorator was erroneous.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/log/TaskListenerDecorator.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/log/TaskListenerDecorator.java
@@ -186,7 +186,7 @@ public abstract class TaskListenerDecorator implements /* TODO Remotable */ Seri
         @NonNull
         @Override public OutputStream decorate(@NonNull OutputStream logger) throws IOException, InterruptedException {
             // TODO BodyInvoker.MergedFilter probably has these backwards
-            return original.decorate(subsequent.decorate(logger));
+            return decorateAll(logger, List.of(subsequent, original));
         }
 
         @Override public String toString() {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/log/TaskListenerDecorator.java
```diff
@@ -219,6 +219,17 @@ private static class ConsoleLogFilterAdapter extends TaskListenerDecorator {
 
     }
 
+    private static OutputStream decorateAll(OutputStream base, List<TaskListenerDecorator> decorators) {
+        for (TaskListenerDecorator decorator : decorators) {
+            try {
+                base = decorator.decorate(base);
+            } catch (Throwable x) {
+                LOGGER.log(Level.WARNING, null, x);
```
jtnord, line 227: Whilst this defends against the issue I am concerned this error will never be picked up - most people are not looking at jenkins logs.  Without a build to log at least prompt the build user to ask the admin to look at the logs.

Having said that - I have no idea how we would get the charset to print something here - without the charset we risk corrupting the entire output :(

jtnord (approved):

rsandell (approved):

src/main/java/org/jenkinsci/plugins/workflow/log/TaskListenerDecorator.java
jglick: ⇒ d191982eb38dd0dc85a75c2378e17b3a278e1ad8

> the charset to print something here

⇒ 696e4f8900771738d075bbc38524ff0de7899a77

src/main/java/org/jenkinsci/plugins/workflow/log/TaskListenerDecorator.java
```diff
@@ -219,6 +221,18 @@ private static class ConsoleLogFilterAdapter extends TaskListenerDecorator {
...
+                Functions.printStackTrace(x, new PrintStream(base, true, StandardCharsets.UTF_8));
```
jtnord, line 230: is the pipeline console log always `UTF-8` and never platform dependant?

jtnord (approved):

jglick (commented):
Got GitHub into a weird corner case where I have zero pending changes yet still have a pending review I need to cancel and cannot do so without leaving a comment.

PR #294, opened 2023-07-13
Define `OutputStreamTaskListener` & close `BufferedBuildListener`

jglick:
Logging improvements needed for https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/323.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/log/OutputStreamTaskListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/log/OutputStreamTaskListener.java
@@ -1,0 +1,100 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2023 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.jenkinsci.plugins.workflow.log;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.model.TaskListener;
+import java.io.FilterOutputStream;
+import java.io.OutputStream;
+import java.io.PrintStream;
+import java.lang.reflect.Field;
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/workflow/log/OutputStreamTaskListener.java
```diff
@@ -0,0 +1,113 @@
...
+        if (listener instanceof OutputStreamTaskListener) {
+            return ((OutputStreamTaskListener) listener).getOutputStream();
+        }
+        PrintStream ps = listener.getLogger();
+        if (ps.getClass() != PrintStream.class) {
+            Logger.getLogger(OutputStreamTaskListener.class.getName()).warning(() -> "Unexpected PrintStream subclass " + ps.getClass().getName() + " which might override write(…); error handling is degraded unless OutputStreamTaskListener is used: " + listener.getClass().getName());
+            return ps;
+        }
+        if (Runtime.version().compareToIgnoreOptional(Runtime.Version.parse("17")) >= 0) {
+            Logger.getLogger(OutputStreamTaskListener.class.getName()).warning(() -> "On Java 17+ error handling is degraded unless OutputStreamTaskListener is used: " + listener.getClass().getName());
```
dwnusbaum, line 67: Should we add a stack trace to these messages? Otherwise I don't really see how a Jenkins developer would be able to track down the problematic code.

dwnusbaum (approved):

src/main/java/org/jenkinsci/plugins/workflow/log/OutputStreamTaskListener.java
jglick: The problem is just with the named class. A stack trace would not really add anything important.

dwnusbaum: But the interesting thing is where that class is being used, right? E.g. If I see `StreamTaskListener` this message is unhelpful. Or do you only think this will only happen for custom listener implementations where the class name makes the use site obvious?

jglick: It should only happen for custom `TaskListener` implementations (this is for Pipeline builds where `StreamTaskListener` should not be used), and the use site is beside the point—either the class implements the new interface or it does not.

dwnusbaum: >It should only happen for custom TaskListener implementations (this is for Pipeline builds where StreamTaskListener should not be used)

Ok, probably fine then.

> either the class implements the new interface or it does not.

Sure, but it was not obvious to me what classes we would expect here. If `StreamTaskListener` or `LogTaskListener` is possible then the message is not helpful and my concern was that it might cause confusing log spam. For example if we hit https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/1f55bd46852f56767bd2d664fa055deaa98c88cf/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java#L433 (seems like an obscure case) then is this warning reachable and would it be logged repeatedly? It isn't obvious to me.

jglick: In that case probably, but something is already pretty broken then. The warning could be limited to classes defined in plugins, and/or for that error handler we could make a special implementation. I can file a follow-up to refine a little bit.

dwnusbaum: Ok, it does not seem particularly important given your explanation, it just wasn't totally clear to me what kinds of scenarios we should anticipate.

src/main/java/org/jenkinsci/plugins/workflow/log/OutputStreamTaskListener.java
```diff
@@ -0,0 +1,113 @@
...
+import java.lang.reflect.InaccessibleObjectException;
+import java.nio.charset.StandardCharsets;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import org.apache.commons.io.output.ClosedOutputStream;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.Beta;
+
+/**
+ * {@link TaskListener} which can directly return an {@link OutputStream} not wrapped in a {@link PrintStream}.
```
jglick, line 42: https://issues.jenkins.io/browse/JENKINS-71164 FTR

PR #296, opened 2023-07-18
[JENKINS-56446] Do not permanently close the log stream in FileLogStorage if an interrupted thread attempts to write to it

dwnusbaum:
See [JENKINS-56446](https://issues.jenkins.io/browse/JENKINS-56446).

While investigating other changes I finally tracked down an issue I have seen before where the main Pipeline log gets closed for some unknown reason, blocking all further writes. I will look for a JENKINS ticket shortly, and I know there is a CloudBees-internal ticket about this issue.

The problem is that `FileLogStorage.checkId` uses a `FileChannel` method to get the current position of the log whenever the log output switches from one step to another. `FileChannel` methods close the channel, and in this case the stream it was derived from, if the thread is interrupted when performing any IO operations. This behavior is undesirable for Pipeline, because it means that any thread writing log output for a step has the ability to inadvertently close the log stream permanently if that thread is interrupted, even if the interruption is expected/controlled (e.g. the `timeout` step).

For some additional context, I also saw a similar issue with another implementation of `LogStorage` that attempted to use `FileChannel`, and in that case we avoided it completely in favor of plain `FileOutputStream` to fix the issue.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
@@ -77,7 +77,9 @@ public final class FileLogStorage implements LogStorage {
     private final File index;
     @SuppressFBWarnings(value = "IS2_INCONSISTENT_SYNC", justification = "actually it is always accessed within the monitor")
     private FileOutputStream os;
+    @SuppressFBWarnings(value = "IS2_INCONSISTENT_SYNC", justification = "actually it is always accessed within the monitor")
     private long osStartPosition;
+    @SuppressFBWarnings(value = "IS2_INCONSISTENT_SYNC", justification = "actually it is always accessed within the monitor")
     private CountingOutputStream cos;
     @SuppressFBWarnings(value = "IS2_INCONSISTENT_SYNC", justification = "we only care about synchronizing writes")
     private OutputStream bos;
--- a/src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
@@ -48,6 +48,7 @@ import java.util.Objects;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.apache.commons.io.input.NullReader;
+import org.apache.commons.io.output.CountingOutputStream;
 import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.kohsuke.accmod.Restricted;
@@ -58,6 +59,9 @@ import org.kohsuke.stapler.framework.io.ByteBuffer;
  * Simple implementation of log storage in a single file that maintains a side file with an index indicating where node transitions occur.
  * Each line in the index file is a byte offset, optionally followed by a space and then a node ID.
  */
+/* Note: Avoid FileChannel methods in this class, as they close the channel and its parent stream if the thread is
+   interrupted, which is problematic given that we do not control the threads which write to the log file.
+*/
 @Restricted(Beta.class)
 public final class FileLogStorage implements LogStorage {
 
@@ -73,6 +77,8 @@ public final class FileLogStorage implements LogStorage {
     private final File index;
     @SuppressFBWarnings(value = "IS2_INCONSISTENT_SYNC", justification = "actually it is always accessed within the monitor")
     private FileOutputStream os;
+    private long osStartPosition;
+    private CountingOutputStream cos;
     @SuppressFBWarnings(value = "IS2_INCONSISTENT_SYNC", justification = "we only care about synchronizing writes")
     private OutputStream bos;
     private Writer indexOs;
@@ -86,7 +92,9 @@ public final class FileLogStorage implements LogStorage {
     private synchronized void open() throws IOException {
         if (os == null) {
             os = new FileOutputStream(log, true);
-            bos = new GCFlushedOutputStream(new DelayBufferedOutputStream(os));
+            osStartPosition = log.length();
+            cos = new CountingOutputStream(os);
+            bos = new GCFlushedOutputStream(new DelayBufferedOutputStream(cos));
             if (index.isFile()) {
                 try (BufferedReader r = Files.newBufferedReader(index.toPath(), StandardCharsets.UTF_8)) {
                     // TODO would be faster to scan the file backwards for the penultimate \n, then convert the byte sequence from there to EOF to UTF-8 and set lastId accordingly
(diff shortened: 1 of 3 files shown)


src/test/java/org/jenkinsci/plugins/workflow/log/FileLogStorageTest.java
```diff
@@ -55,4 +55,18 @@ public class FileLogStorageTest extends LogStorageTestBase {
...
+        TaskListener overall = ls.overallListener();
+        overall.getLogger().println("overall 1");
+        Thread.currentThread().interrupt();
+        TaskListener stepLog = ls.nodeListener(new MockNode("1"));
+        stepLog.getLogger().println("step 1");
+        assertTrue(Thread.interrupted());
+        close(stepLog);
+        overall.getLogger().println("overall 2");
+        close(overall);
+        assertOverallLog(0, lines("overall 1", "<span class=\"pipeline-node-1\">step 1", "</span>overall 2"), true);
```
dwnusbaum, line 69: Without the fix, only `overall 1` gets written to the log, and you get the "failed to flush" messages in the JENKINS ticket from `BuildWatcher` in the log as it copies the output which calls `maybeFlush`.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
```diff
@@ -126,17 +126,28 @@ private void checkId(String id) throws IOException {
...
-                indexOs.write(pos + "\n");
-            } else {
-                indexOs.write(pos + " " + id + "\n");
+            // FileChannel methods such as position() close the channel (and the stream it was derived from) if the
+            // current thread is interrupted. We must not allow interruption on some arbitrary thread attempting to
+            // write to the log to close the stream irrevocably, so we clear the interrupt status before using the
+            // channel and then restore the interrupt status afterwards.
+            boolean interrupted = Thread.interrupted();
+            try {
+                long pos = os.getChannel().position();
```
dwnusbaum, line 135: Not perfect, because we do not mask interruptions that happen during the call to `position()`, but maybe this is good enough for most practical cases? We could check for closure of the channel and recreate `os` in that case and then recheck the position, but I am not sure if there is a way to make this code robust against repeated interruption.

jglick, line 135: It would suffice to have the `finally` block run as soon as `Channel.position` has been called, right? I.e., the operations on `indexOs` need not be included?

src/test/java/org/jenkinsci/plugins/workflow/log/LogStorageTestBase.java
```diff
@@ -350,12 +350,12 @@ protected String lines(CharSequence... lines) {
...
     }
 
-    private static class MockNode extends FlowNode {
+    protected static class MockNode extends FlowNode {
         MockNode(String id) {super(new MockFlowExecution(), id);}
         @Override protected String getTypeDisplayName() {return null;}
     }
 
-    private static class MockFlowExecution extends FlowExecution {
+    protected static class MockFlowExecution extends FlowExecution {
```
jglick, line 358: (could be reverted)

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
jglick: > track the position manually

Using https://javadoc.io/static/commons-io/commons-io/2.13.0/org/apache/commons/io/output/CountingOutputStream.html I suppose? You would however need to check the file length when opening `os` in append mode to get an initial offset.

jglick (approved):

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
dwnusbaum: Yeah, it is a bit unfortunate because I think `FileOutputStream` has everything needed to read the position from the file descriptor in the same way that `FileChannel.position()` does, but there is no public API to support it.

dwnusbaum: Also I don't think we need an `AtomicLong`, just a `long`, since all of this code is in `synchronized` blocks.

jglick: :thinking: not quite, since those methods do not flush `bos`, so there would be a small race condition. Really want something like

```java
cos = new CountingOutputStream(os);
bos = new GCFlushedOutputStream(new DelayBufferedOutputStream(cos));
```

to track when bytes were actually flushed.

dwnusbaum: See [e611d9c](https://github.com/jenkinsci/workflow-api-plugin/pull/296/commits/e611d9c3ffa8627cab1a50a774b60ec3266c56f8).

jglick (approved):
Looks straightforward.

Many thanks for tracking this down! Out of curiosity, do you know what about the previously posted reproduction scenarios would lead to a thread interruption?

src/test/java/org/jenkinsci/plugins/workflow/log/LogStorageTestBase.java
dwnusbaum: Reverted in https://github.com/jenkinsci/workflow-api-plugin/pull/296/commits/e92927515ac10b73ffa373b8bfdcf80ddedc5d3e.

jglick:
Disabling automerge since I noticed https://github.com/jenkinsci/workflow-api-plugin/blob/e92927515ac10b73ffa373b8bfdcf80ddedc5d3e/src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java#L95 would not be well covered by existing tests: none of the ones in this plugin using `JenkinsSessionRule` seem to check anything interesting about log text. I probably need to either add a cross-restart test to `LogStorageTestBase`, or introduce a new test case specifically about `FileLogStorage` verifying that the append mode actually appends and that the index keeps up accordingly.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
```diff
@@ -126,7 +134,7 @@ private void checkId(String id) throws IOException {
         assert Thread.holdsLock(this);
         if (!Objects.equals(id, lastId)) {
             bos.flush();
-            long pos = os.getChannel().position();
+            long pos = osStartPosition + cos.getByteCount();
```
jglick, line 137: SpotBugs complaints

dwnusbaum:
> do you know what about the previously posted reproduction scenarios would lead to a thread interruption?

No, I do not. @jgreffe and I saw this same issue with an alternate implementation of `LogStorage` which used `FileChannel` extensively. We are both on Mac, but for whatever reason he would consistently see the same behavior as in the reported issue, just running mostly normal Pipelines with a lot of log output, but for me things were working fine. I was not sure how to track down the source of the interrupts though, perhaps it could be done with a Java agent or some other kind of JVM instrumentation? I looked briefly through searches like [this](https://github.com/search?type=code&q=+owner%3Ajenkinsci+.interrupt%28%29+path%3A%2Fsrc%2Fmain) at that time but nothing particularly stood out.

In this particular case I am prototyping some unrelated changes in `workflow-job` which caused [WorkflowRunRestartTest.interruptedWhileStartingPerformanceOptimized](https://github.com/jenkinsci/workflow-job-plugin/blob/d2290d3341af6011b1f7833ab57d4db28b824e3a/src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunRestartTest.java#L250-L274) to fail and allowed me to reproduce this issue consistently.

src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java
dwnusbaum: Any idea offhand what the cost of reentrant synchronization is on a method call? We could delete `assert Thread.holdsLock(this);` in `checkId` and just make the method `synchronized` to help SpotBugs understand what is happening, but maybe that has a nontrivial cost?

dwnusbaum: https://github.com/jenkinsci/workflow-api-plugin/pull/296/commits/b763772ecd87592dbfe8b1bf0a5bb66cde79393f suppresses the warnings though.

jglick (approved):

jglick:
> Is this something you are actively working on?

Yes. Actually backing up a bit to try to let `pipeline-cloudwatch-logs` tests be run using SSO (needed anyway to validate https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/323). One way or another, I hope to file a separate PR introducing better test coverage for logs across controller restarts in this repo, then get that reviewed & merged, then merge that into this PR so we can enable automerge again with confidence.

jglick:
Never mind. After exploring a bit, I found that in fact changing https://github.com/jenkinsci/workflow-api-plugin/blob/1679fa2f0f761c285a7309b8781c72e8fd9601c8/src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java#L88 to `false` causes https://github.com/jenkinsci/workflow-api-plugin/blob/1679fa2f0f761c285a7309b8781c72e8fd9601c8/src/test/java/org/jenkinsci/plugins/workflow/log/LogStorageTestBase.java#L139-L143 to fail. So does changing https://github.com/jenkinsci/workflow-api-plugin/blob/b763772ecd87592dbfe8b1bf0a5bb66cde79393f/src/main/java/org/jenkinsci/plugins/workflow/log/FileLogStorage.java#L97 to `0`.

PR #297, opened 2023-07-20
Add `LogStorage` utility methods that expose the functionality of `BufferedBuildListener`, `DelayBufferedOutputStream`, and `GCFlushedOutputStream`

dwnusbaum:
`LogStorage` implementations that are similar to `FileLogStorage` need to be able to use the same logic as `BufferedBuildListener`, `DelayBufferedOutputStream`, and `GCFlushedOutputStream` for both performance and correctness reasons, so we should expose these classes' functionality.

I think we can just add two methods to wrap a given `OutputStream` as needed without having to fully expose the classes. I am not sure what name would be best, feel free to recommend whatever you like.

### Testing done

```[tasklist]
### Submitter checklist
```


jglick (approved):
Allows fixes made for #294 to be used by custom storage systems.

jglick:
> I am not sure what name would be best

The method names chosen seem reasonable to me.

PR #299, opened 2023-08-12
Test on Java 21

NotMyFault:


basil:
A suggestion made without reasoning and inconsistent with the archetype. Should the suggestion be ignored without reasoning as well?

jglick:
Ah, never saw https://github.com/jenkinsci/archetypes/pull/651.

basil:
The reasoning behind dropping Java 11 from the archetype: since implementing a migration of this nature across the whole ecosystem involves so much code change and since Java 11 will likely be EOL in about a year, it is globally more efficient to do one pass through the ecosystem adding 21 and dropping 11 rather than two passes through the ecosystem separated in time by about a year. Since Java versions have gotten more strict over time rather than less strict, it is unlikely for a bug to be present when running on 11 but not 17 or 21.

jglick (approved):

PR #301, opened 2023-08-16
Define termination milestones for suspending & saving

jglick:
Downstream use in `workflow-cps` coming soon. Also see https://github.com/jenkinsci/workflow-job-plugin/pull/375.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
@@ -11,6 +11,7 @@ import hudson.Extension;
 import hudson.ExtensionList;
 import hudson.XmlFile;
 import hudson.init.InitMilestone;
+import hudson.init.TermMilestone;
 import hudson.init.Terminator;
 import hudson.model.Computer;
 import hudson.model.listeners.ItemListener;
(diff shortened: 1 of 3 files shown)


dwnusbaum (approved):

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -238,7 +253,9 @@ public void onFailure(@NonNull Throwable t) {
     }
 
     @Restricted(DoNotUse.class)
-    @Terminator public static void saveAll() throws InterruptedException {
+    @SuppressWarnings("deprecation")
+    @Terminator(requires = EXECUTIONS_SUSPENDED, attains = LIST_SAVED)
```
dwnusbaum, line 257: One thing I just thought about: if you have `workflow-api` installed but not `workflow-cps`, is the reactor smart enough to see that nothing attains `EXECUTIONS_SUSPENDED` and still allow this to run? Or does it block (bad), or get skipped (fine since you won't have any executions anyway)?

jglick: > is the reactor smart enough to see that nothing attains `EXECUTIONS_SUSPENDED` and still allow this to run?

I wondered the same, but actually you can see from the fact that tests in this plugin pass that it is OK: it just considers `EXECUTIONS_SUSPENDED` to have attained immediately.

dwnusbaum: > actually you can see from the fact that tests in this plugin pass that it is OK

Yeah I suspected that was the case, but wanted to make sure.

PR #302, opened 2023-08-24
Prevent ForkScanner from skipping nodes when scanning incomplete parallel steps that only have a single branch

dwnusbaum:
Looking into more issues along the lines of https://github.com/jenkinsci/workflow-api-plugin/pull/287. Note that I did verify that this particular issue occurs both before and after that PR.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
@@ -235,7 +235,6 @@ public class ForkScanner extends AbstractFlowScanner {
     // TODO see if this can be replaced with a FlowChunk acting as a container class for a list of FlowNodes
     static class FlowSegment implements FlowPiece {
         ArrayList<FlowNode> visited = new ArrayList<>();
-        FlowPiece after;
         boolean isLeaf = true;
 
         @Override
@@ -244,44 +243,20 @@ public class ForkScanner extends AbstractFlowScanner {
         }
 
         /**
-         * We have discovered a forking node intersecting our FlowSegment in the middle or meeting at the end
-         * Now we need to split the flow, or pull out the fork point and make both branches follow it
+         * We have discovered a parallel step start node at the end of our segment, so we create a fork for that step and point it to this segment.
          * @param nodeMapping Mapping of BlockStartNodes to flowpieces (forks or segments)
-         * @param joinPoint Node where the branches intersect/meet (fork point)
-         * @param joiningBranch Flow piece that is joining this
-         * @throws IllegalStateException When you try to split a segment on a node that it doesn't contain, or invalid graph structure
+         * @param joinPoint The parallel step start node
+         * @throws IllegalStateException When you try to split a segment in an invalid way
          * @return Recreated fork
          */
-        Fork split(@NonNull HashMap<FlowNode, FlowPiece> nodeMapping, @NonNull BlockStartNode joinPoint, @NonNull FlowPiece joiningBranch) {
-            int index = visited.lastIndexOf(joinPoint);  // Fork will be closer to end, so this is better than indexOf
-            Fork newFork = new Fork(joinPoint);
-
-            if (index < 0) {
-                throw new IllegalStateException("Tried to split a segment where the node doesn't exist in this segment");
-            } else if (index == this.visited.size()-1) { // We forked just off the most recent node
-                newFork.following.add(this);
-                newFork.following.add(joiningBranch);
-                this.visited.remove(index);
-            } else if (index == 0) {
-                throw new IllegalStateException("We have a cyclic graph or heads that are not separate branches!");
-            } else { // Splitting at some midpoint within the segment, everything before becomes part of the following
-                // Execute the split: create a new fork at the fork point, and shuffle the part of the flow after it
-                //   to a new segment and add that to the fork.
-
-                FlowSegment newSegment = new FlowSegment();
-                newSegment.after = this.after;
-                newSegment.visited.addAll(this.visited.subList(0, index));
-                newFork.following.add(newSegment);
-                newFork.following.add(joiningBranch);
-                this.after = newFork;
-                this.isLeaf = false;
-
-                // Remove the part before the fork point
-                this.visited.subList(0, index+1).clear();
-                for (FlowNode n : newSegment.visited) {
-                    nodeMapping.put(n, newSegment);
-                }
+        Fork split(@NonNull HashMap<FlowNode, FlowPiece> nodeMapping, @NonNull BlockStartNode joinPoint) {
+            int index = visited.size() - 1;
+            if (visited.isEmpty() || visited.get(index) != joinPoint) {
+                throw new AssertionError("Misuse of ForkScanner.split");
             }
+            Fork newFork = new Fork(joinPoint);
+            newFork.following.add(this);
+            this.visited.remove(index);
             nodeMapping.put(joinPoint, newFork);
             return newFork;
         }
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -1027,4 +1028,31 @@ public void inProgressParallelInParallel() throws Exception {
...
+                "    parallel('innerB': {\n" +
+                "       semaphore('innerB')\n" +
+                "    })\n" +
+                "  }\n" +
+                ")", true));
+        WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+        SemaphoreStep.waitForStart("innerA/1", b);
+        SemaphoreStep.waitForStart("innerB/1", b);
+        ForkScanner scanner = new ForkScanner();
+        sanityTestIterationAndVisiter(b.getExecution().getCurrentHeads());
```
dwnusbaum, line 1052: This fails. The current behavior is that ForkScanner visits the two `semapshore` steps,  the `innerA` and `innerB` branches, and then the _outer_ `parallel` step, and the `FlowStartNode`.

dwnusbaum: Oh and I have tested various combinations and this seems to be the simplest reproducer - 2 outer branches each with 1 inner branch. You can add more outer branches and the problem still occurs, but if you only have 1 outer branch things work fine. If you add more than 1 inner branch, the problem goes away for the outer branches that have more than 1 inner branch.

dwnusbaum: Ah actually we can slightly simplify things, 2 outer branches with the first branch having 1 inner branch and the second branch not having nested parallelism also reproduces the issue.

jglick: FTR https://github.com/jenkinsci/workflow-api-plugin/pull/302/checks?check_run_id=16190024918

```
ForkScanner should visit the same number of nodes as DepthFirstScanner expected:<10> but was:<6>
```

jglick (commented):
Fine to start by merging as `tests` with `@Ignore`.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -244,44 +243,20 @@ public boolean isLeaf() {
...
-                newFork.following.add(newSegment);
-                newFork.following.add(joiningBranch);
-                this.after = newFork;
-                this.isLeaf = false;
-
-                // Remove the part before the fork point
-                this.visited.subList(0, index+1).clear();
-                for (FlowNode n : newSegment.visited) {
-                    nodeMapping.put(n, newSegment);
-                }
```
dwnusbaum, line 283: This logic previously had branches to account for being called in various ways. I changed the way it is called making all but one of the branches unreachable.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -235,7 +235,6 @@ interface FlowPiece {  // Mostly a marker
     // TODO see if this can be replaced with a FlowChunk acting as a container class for a list of FlowNodes
     static class FlowSegment implements FlowPiece {
         ArrayList<FlowNode> visited = new ArrayList<>();
-        FlowPiece after;
```
dwnusbaum, line 238: Was not ever used apart from test assertions and my simplifications left it always `null`, making it completely pointless.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -407,27 +382,30 @@ ArrayDeque<ParallelBlockStart> leastCommonAncestor(@NonNull final Set<FlowNode>
                 FlowPiece existingPiece = branches.get(nextBlockStart);
                 if (existingPiece == null && myPiece instanceof FlowSegment) { // No merge, just add to segment
                     ((FlowSegment) myPiece).add(nextBlockStart);
-                    branches.put(nextBlockStart, myPiece);
+                    if (isParallelStart(nextBlockStart)) {
+                        // We make a fork for every parallel step preemptively in case it has less than 2 branches.
```
dwnusbaum, line 386: This is the fix, essentially instead of waiting until we have a piece that needs to be merge with an existing segment, we create the `Fork` objects right away as soon as we see a `parallel` step.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -407,27 +382,30 @@ ArrayDeque<ParallelBlockStart> leastCommonAncestor(@NonNull final Set<FlowNode>
...
-                        Fork f = ((FlowSegment) existingPiece).split(branches, (BlockStartNode)nextBlockStart, myPiece);
-                        // If we split the existing segment at its end, we created a fork replacing its latest node
-                        // Thus we must replace the piece with the fork ahead of it
-                        if (f.following.contains(existingPiece) ) {
-                            int headIndex = livePieces.indexOf(existingPiece);
-                            livePieces.set(headIndex, f);
-                        }
-                        parallelForks.add(f);
+                    } else {
+                        throw new IllegalStateException("Unexpected join on " + existingPiece);
```
dwnusbaum, line 408: Should be unreachable now, since all merges should hit the above condition 🤞

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -372,15 +377,6 @@ public void testFlowSegmentSplit() throws Exception {
...
 
-        ForkScanner.Fork forked = mainBranch.split(nodeMap, (BlockStartNode)exec.getNode("4"), sideBranch);
-        ForkScanner.FlowSegment splitSegment = (ForkScanner.FlowSegment)nodeMap.get(BRANCH1_END); // New branch
-        Assert.assertNull(splitSegment.after);
-        FlowTestUtils.assertNodeOrder("Branch 1 split after fork", splitSegment.visited, 9, 8, 6);
-
-        // Just the single node before the fork
-        Assert.assertEquals(forked, mainBranch.after);
-        FlowTestUtils.assertNodeOrder("Head of flow, pre-fork", mainBranch.visited, 3);
-
```
dwnusbaum, line 383: This test was using `split` directly in a way that is no longer supported and not done by production code, so I tried to adapt the tests.

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
dwnusbaum: [819d8a8](https://github.com/jenkinsci/workflow-api-plugin/pull/302/commits/819d8a89242aa123239f321c5566046727b69d98) should fix the issue.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
dwnusbaum: Turns out this is still reachable in at least some cases, see `com.cloudbees.workflow.rest.endpoints.ParallelStepTest.test_stages_order` in `pipeline-stage-view`

```
java.lang.IllegalStateException: Unexpected join on org.jenkinsci.plugins.workflow.graphanalysis.ForkScanner$FlowSegment@428b6b34
	at org.jenkinsci.plugins.workflow.graphanalysis.ForkScanner.leastCommonAncestor(ForkScanner.java:414)
	at org.jenkinsci.plugins.workflow.graphanalysis.ForkScanner.setHeads(ForkScanner.java:453)
	at org.jenkinsci.plugins.workflow.graphanalysis.AbstractFlowScanner.setup(AbstractFlowScanner.java:140)
	at org.jenkinsci.plugins.workflow.graphanalysis.AbstractFlowScanner.setup(AbstractFlowScanner.java:151)
	at org.jenkinsci.plugins.workflow.graphanalysis.ForkScanner.visitSimpleChunks(ForkScanner.java:630)
	at com.cloudbees.workflow.rest.external.RunExt.createNew(RunExt.java:323)
	at com.cloudbees.workflow.rest.external.RunExt.create(RunExt.java:311)
	at com.cloudbees.workflow.rest.external.JobExt.create(JobExt.java:143)
	at com.cloudbees.workflow.rest.endpoints.JobAPI.doRuns(JobAPI.java:69)
```

dwnusbaum: I haven't been able to reproduce the failure yet. The test in question starts a build that uses `sleep` in some parallel branches and just checks it as soon as it has a `Run`, so the flow execution is an an unknown and possibly changing state when `ForkScanner` runs.

dwnusbaum: Finally reproduced it, seems to be quite rare (fails about 1 in 40 times for me), so I'm glad it happened to fail in the PCT run!

dwnusbaum: The problematic case seems to be where the parallel step start node is itself a head. That means it gets left out of `blockStartIterator` for all heads, and we end up trying to join two heads on whatever segment has the first parent of the parallel step. It seems to be possible to get things working by restoring some of the old cases in `FlowSegment.split`, but that seems undesirable, because we end up with a `Fork` whose start node is not actually a parallel step, which is bizarre.

dwnusbaum: I have a fix for this, but as I add new test assertions I keep finding new things that don't work and need to compare against the previous behavior to understand if they ever worked, so it is slow going.

dwnusbaum:
This fix has subtle problems in various cases that are difficult to fix comprehensively. For now I am closing this, see https://github.com/jenkinsci/workflow-api-plugin/pull/306.

PR #304, opened 2023-08-25
Do not unregister completed executions when iterating over `FlowExecutionList` to avoid unnecessary log warnings

Vlatombe:
cf. https://github.com/jenkinsci/kubernetes-plugin/pull/1342#issuecomment-1692993104

### Testing done

AFAIU to reproduce the issue would require to set up a cloud agent. On its termination, it calls indirectly `computeNext()` which unregisters the `FlowExecutionOwner` if it is the last node in the pipeline. Then the regular `unregister` call when the `WorkflowRun` completes now issues a warning because it has already been unregistered.

```[tasklist]
### Submitter checklist
- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
@@ -80,7 +80,7 @@ public class FlowExecutionList implements Iterable<FlowExecution> {
      */
     @Override
     public Iterator<FlowExecution> iterator() {
-        return new AbstractIterator<FlowExecution>() {
+        return new AbstractIterator<>() {
             final Iterator<FlowExecutionOwner> base = runningTasks.iterator();
 
             @Override
@@ -88,12 +88,7 @@ public class FlowExecutionList implements Iterable<FlowExecution> {
                 while (base.hasNext()) {
                     FlowExecutionOwner o = base.next();
                     try {
-                        FlowExecution e = o.get();
-                        if (e.isComplete()) {
-                            unregister(o);
-                        } else {
-                            return e;
-                        }
+                        return o.get();
                     } catch (Throwable e) {
                         LOGGER.log(Level.FINE, "Failed to load " + o + ". Unregistering", e);
                         unregister(o);
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
@@ -39,7 +39,6 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.graphanalysis.LinearBlockHoppingScanner;
-import org.jvnet.hudson.reactor.Milestone;
 
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.Beta;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -80,20 +80,15 @@ public FlowExecutionList() {
...
                 while (base.hasNext()) {
                     FlowExecutionOwner o = base.next();
                     try {
-                        FlowExecution e = o.get();
-                        if (e.isComplete()) {
-                            unregister(o);
-                        } else {
-                            return e;
-                        }
+                        return o.get();
```
jglick, line 91: You need to keep the check for `o.isComplete()` but just `continue` (or do nothing) in that case, to find the next one (if any) which is _not_ complete.

jglick (changes requested):
Idea seems right, but we want to keep the `unregister` call from https://github.com/jenkinsci/workflow-api-plugin/blob/4521fef5822dd23c5713fede0a7faf82be73a4b4/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java#L198-L213 (in some form) as this cleans up stale entries.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -205,8 +208,11 @@ public static class ItemListenerImpl extends ItemListener {
         public void onLoaded() {
             FlowExecutionList list = FlowExecutionList.get();
             for (final FlowExecution e : list) {
-                // The call to FlowExecutionOwner.get in the implementation of iterator() is sufficent to load the Pipeline.
+                // The call to FlowExecutionOwner.get in the implementation of iterator() is sufficient to load the Pipeline.
                 LOGGER.log(Level.FINE, "Eagerly loaded {0}", e);
+                if (e.isComplete()) {
```
jglick, line 213: This will never happen because you are skipping complete executions. There would need to be an alternate iteration method or something.

jglick (changes requested):

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -206,16 +205,24 @@ public boolean isResumptionComplete() {
...
+    }
+
+    private void resume() {
+        for (final FlowExecutionOwner o : runningTasks) {
+            try {
+                FlowExecution e = o.get();
                 LOGGER.log(Level.FINE, "Eagerly loaded {0}", e);
                 if (e.isComplete()) {
-                    list.unregister(e.getOwner());
+                    unregister(o);
```
Vlatombe, line 218: Yikes that's going to fail as well

Vlatombe:
@dwnusbaum can you help with review/merge of this fix? Currently holds up another series of PRs (e.g. https://github.com/jenkinsci/kubernetes-plugin/pull/1342)

dwnusbaum (approved):

dwnusbaum:
@Vlatombe Did you check this against tests in `workflow-cps` and `workflow-job` since the new code seems to be not well covered by existing tests in this plugin? I can check it in a minute if you haven't already.

dwnusbaum:
> To cover the new lines would require a test pipeline involving cloud agents

It should be possible to set up a relevant test with `@LocalData` and/or a custom Pipeline step, but either way I mainly just wanted to make sure it didn't affect existing tests in those plugins, and from a quick local check things seem ok.

PR #305, opened 2023-08-25
Make ForkScanner visit parallel branches for incomplete parallel steps in a more consistent order (ideally reverse declaration order)

dwnusbaum:
Builds on top of #302 to make it easier to test both of them together, but a separate PR because because I think this will need further investigation.

Here is the effective diff: https://github.com/dwnusbaum/workflow-api-plugin/compare/more-forkscanner-problems...dwnusbaum:workflow-api-plugin:more-forkscanner-problems-2

There are two things I am trying to fix in this PR:
* We want `ForkScanner` to visit parallel branches in the same order regardless of whether the build is complete or not
* We want that order to be the reverse of the declared order of the branches in the Pipeline script

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
@@ -235,7 +235,6 @@ public class ForkScanner extends AbstractFlowScanner {
     // TODO see if this can be replaced with a FlowChunk acting as a container class for a list of FlowNodes
     static class FlowSegment implements FlowPiece {
         ArrayList<FlowNode> visited = new ArrayList<>();
-        FlowPiece after;
         boolean isLeaf = true;
 
         @Override
@@ -244,44 +243,20 @@ public class ForkScanner extends AbstractFlowScanner {
         }
 
         /**
-         * We have discovered a forking node intersecting our FlowSegment in the middle or meeting at the end
-         * Now we need to split the flow, or pull out the fork point and make both branches follow it
+         * We have discovered a parallel step start node at the end of our segment, so we create a fork for that step and point it to this segment.
          * @param nodeMapping Mapping of BlockStartNodes to flowpieces (forks or segments)
-         * @param joinPoint Node where the branches intersect/meet (fork point)
-         * @param joiningBranch Flow piece that is joining this
-         * @throws IllegalStateException When you try to split a segment on a node that it doesn't contain, or invalid graph structure
+         * @param joinPoint The parallel step start node
+         * @throws IllegalStateException When you try to split a segment in an invalid way
          * @return Recreated fork
          */
-        Fork split(@NonNull HashMap<FlowNode, FlowPiece> nodeMapping, @NonNull BlockStartNode joinPoint, @NonNull FlowPiece joiningBranch) {
-            int index = visited.lastIndexOf(joinPoint);  // Fork will be closer to end, so this is better than indexOf
-            Fork newFork = new Fork(joinPoint);
-
-            if (index < 0) {
-                throw new IllegalStateException("Tried to split a segment where the node doesn't exist in this segment");
-            } else if (index == this.visited.size()-1) { // We forked just off the most recent node
-                newFork.following.add(this);
-                newFork.following.add(joiningBranch);
-                this.visited.remove(index);
-            } else if (index == 0) {
-                throw new IllegalStateException("We have a cyclic graph or heads that are not separate branches!");
-            } else { // Splitting at some midpoint within the segment, everything before becomes part of the following
-                // Execute the split: create a new fork at the fork point, and shuffle the part of the flow after it
-                //   to a new segment and add that to the fork.
-
-                FlowSegment newSegment = new FlowSegment();
-                newSegment.after = this.after;
-                newSegment.visited.addAll(this.visited.subList(0, index));
-                newFork.following.add(newSegment);
-                newFork.following.add(joiningBranch);
-                this.after = newFork;
-                this.isLeaf = false;
-
-                // Remove the part before the fork point
-                this.visited.subList(0, index+1).clear();
-                for (FlowNode n : newSegment.visited) {
-                    nodeMapping.put(n, newSegment);
-                }
+        Fork split(@NonNull HashMap<FlowNode, FlowPiece> nodeMapping, @NonNull BlockStartNode joinPoint) {
+            int index = visited.size() - 1;
+            if (visited.isEmpty() || visited.get(index) != joinPoint) {
+                throw new AssertionError("Misuse of ForkScanner.split");
             }
+            Fork newFork = new Fork(joinPoint);
+            newFork.following.add(this);
+            this.visited.remove(index);
             nodeMapping.put(joinPoint, newFork);
             return newFork;
         }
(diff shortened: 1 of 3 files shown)


src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -1020,11 +1022,73 @@ public void inProgressParallelInParallel() throws Exception {
...
+                "    )\n" +                       // 12
+                "  },\n" +                        // 13
+                "  'outerB': {\n" +               // 6
+                "  }\n" +                         // 9
+                ")", true));                      // 14
+        WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+        SemaphoreStep.waitForStart("innerA/1", b);
+        sanityTestIterationAndVisiter(b.getExecution().getCurrentHeads());
+        // TODO: We would like for this order to match the order of the completed build, but that requires reverting
+        // https://github.com/jenkinsci/workflow-api-plugin/pull/287, which would reintroduce a much more severe problem.
```
dwnusbaum, line 1053: See https://github.com/jenkinsci/workflow-api-plugin/pull/287. This is pretty awkward. Perhaps there is a better fix for the problem that PR is trying to address so that we can visit branches in the same order as the completed build in this kind of case.

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -1020,11 +1022,73 @@ public void inProgressParallelInParallel() throws Exception {
...
+                "  '2': { semaphore('2') },\n" +
+                "  '3': { echo '3' },\n" +
+                "  '4': { semaphore('4') },\n" +
+                "  '5': { echo '5' },\n" +
+                ")", true));
+        WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+        SemaphoreStep.waitForStart("2/1", b);
+        SemaphoreStep.waitForStart("4/1", b);
+        sanityTestIterationAndVisiter(b.getExecution().getCurrentHeads());
+        assertBranchOrder(b.getExecution(), "5", "4", "3", "2", "1");
```
dwnusbaum, line 1076: Without this PR, this order is pretty unusual, I think it's 4 2 1 3 5, or maybe it was 1 3 5 4 2, either way it was not a trivial case of the list being backwards or something like that.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -376,13 +352,17 @@ ArrayDeque<ParallelBlockStart> leastCommonAncestor(@NonNull final Set<FlowNode>
...
+            iterators.add(new FilteratorImpl<>((Iterator) new ExpandedEnclosingBlockIterator(f), notAHead));  // We can do this because Parallels always meet at a BlockStartNode
             FlowSegment b = new FlowSegment();
             b.add(f);
             livePieces.add(b);
             branches.put(f, b);
         }
 
+        // We want to visit heads in order of most recent to least recent.
+        Collections.reverse(iterators);
+        Collections.reverse(livePieces);
```
dwnusbaum, line 364: This is the first issue with the ordering, we were looking at heads in the wrong order.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -396,6 +376,7 @@ ArrayDeque<ParallelBlockStart> leastCommonAncestor(@NonNull final Set<FlowNode>
 
                 // Welp we hit the end of a branch
                 if (!blockStartIterator.hasNext()) {
+                    // This case is why we use ExpandedEnclosingBlockIterator instead of FlowNode.iterateEnclosingBlocks - otherwise branch ordering is not preserved.
```
dwnusbaum, line 379: This is the second issue. Otherwise, heads for completed branches get skipped here to start with and are then examined separately later, so the completed branches and incomplete branches were kind of in two separate groups that got concatenated.

src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -785,4 +769,40 @@ public void visitSimpleChunks(@NonNull SimpleChunkVisitor visitor, @NonNull Chun
         }
     }
 
+    /**
+     * Like {@link FlowNode#iterateEnclosingBlocks}, but if the original node is a {@link BlockEndNode}, its
+     * corresponding {@link BlockStartNode} is the first element in the iterator.
```
dwnusbaum, line 774: This is kind of interesting to me, because I have been doing a lot of flow graph traversal lately and I have found that this behavior is what I typically want vs plain `iterateEnclosingBlocks`, so it was surprising to see the default behavior causing problems here. Also sometimes, if the node is itself is a `BlockStartNode`, I want to include it in the iteration as well. IDK if that matters here though - I think you would have to have a parallel branch start node in `FlowExecution.getCurrentHeads`, so it would only matter very temporarily.

dwnusbaum:
This change breaks [`MatrixTest.matrixStagesHaveStatusWhenSkipped`](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/077ad1928a4012780319d9cfc1a40fb0cb510290/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/MatrixTest.java#L703) in `pipeline-model-definition`, but I don't think it really indicates a problem with this change directly. Rather, code like [Util.findStageFlowNodes](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/077ad1928a4012780319d9cfc1a40fb0cb510290/pipeline-model-definition/src/main/groovy/org/jenkinsci/plugins/pipeline/modeldefinition/Utils.groovy#L230-L240) is not guaranteed to work correctly when you have more than one stage with the same name (as in most uses of `matrix`). The ordering change in this PR makes that test break, but you can already get broken behavior in Declarative before this PR with a Pipeline like this:

```
pipeline {
    agent none
    stages {
        stage('Outer') {
            parallel {
                stage('First') {
                    steps {
                        sleep 10
                    }
                }
                stage('Second') {
                    stages {
                        stage('Foo') {
                            steps {
                                error('oops')
                            }
                        }
                        stage('First') {
                            steps {
                                echo 'test'
                            }
                        }
                    }
                }
            }
        }
    }
}
```

Run that and look at the flow nodes on disk and you will see that `Outer/First` gets the `TagAction` for being skipped that is supposed to be assigned to `Outer/Second/First`. Amusingly though, Blue Ocean (and Pipeline Graph View with its derived graph visitor) are semi-unaffected by this problem on the display side because they have a [special case](https://github.com/jenkinsci/pipeline-graph-view-plugin/blob/0844db6ca5544e6181e87381ce8ed6aadea88b6a/src/main/java/io/jenkins/plugins/pipelinegraphview/utils/PipelineNodeGraphVisitor.java#L223) that seems to get hit in this case, although it does not quite work perfectly and so the stage that should be marked as skipped is shown with a blue ellipsis instead of with a line bypassing it as is normally the case. Either way, to really fix this, Declarative needs to guide the search using `FlowNode` IDs or something like that.

dwnusbaum:
This fix has subtle problems in various cases that are difficult to fix comprehensively. For now I am closing this, see https://github.com/jenkinsci/workflow-api-plugin/pull/306.

PR #306, opened 2023-08-29
Add warning to ForkScanner Javadoc describing potential issues when using it with multiple heads

dwnusbaum:
After looking into #302 and #305 further I am not convinced that my proposed fixes are complete or totally safe. For my use case it is far easier to use `DepthFirstScanner` and simply order sibling nodes in my results based on its iteration order rather than try to fix these complex issues with `ForkScanner`, but I think it makes sense to add a warning for others based on what I found while investigating these issues.

If you are writing code that needs to work with arbitrary Pipelines which may be incomplete, I would avoid `ForkScanner`.

Blue Ocean and Pipeline Graph View mainly use the `ForkScanner.visitSimpleChunks` method, which I only looked into briefly, and so I think their main issue would be that chunks could be skipped and the chunk boundaries may be wrong in unusual cases. Declarative uses `ForkScanner`, but its usage is problematic for complex reasons unrelated to the type of scanner, see https://github.com/jenkinsci/workflow-api-plugin/pull/305#issuecomment-1696291234.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
@@ -51,8 +51,22 @@ import java.util.Set;
 import java.util.stream.Collectors;
 
 /**
- * Scanner that will scan down all forks when we hit parallel blocks before continuing, but generally runs in linear order
- * <p>Think of it as the opposite of {@link DepthFirstScanner}.
+ * Scanner that will scan down all forks when we hit parallel blocks before continuing (as opposed to {@link DepthFirstScanner}), but generally runs in linear order.
+ * <p><strong>Warning</strong> This scanner has various issues when iterating over incomplete builds, or more generally
+ * when multiple heads are passed to {@link #setup}, especially if the build contains nested parallelism.
+ * Consider using {@link DepthFirstScanner} instead in those cases to avoid these issues, employing its iteration order
+ * as needed to sort results (although its global iteration order is inconsistent, its relative iteration order among
+ * siblings at the same nesting level (e.g. parallel branches) is consistent and that is good enough for typical use
+ * cases such as producing a tree).
+ * In particular, the following known issues are possible when using this scanner with multiple heads:
+ * <ul>
+ *   <li>Nodes may be visited more than once (mainly with nested parallelism, but also possible with bad timing while a parallel step is initializing heads for its branches).
+ *   <li>Nodes that should be visited may be skipped (only known to happen for nested parallel steps with less than two branches).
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScanner.java
```diff
@@ -51,8 +51,22 @@
...
 /**
- * Scanner that will scan down all forks when we hit parallel blocks before continuing, but generally runs in linear order
- * <p>Think of it as the opposite of {@link DepthFirstScanner}.
+ * Scanner that will scan down all forks when we hit parallel blocks before continuing (as opposed to {@link DepthFirstScanner}), but generally runs in linear order.
+ * <p><strong>Warning</strong> This scanner has various issues when iterating over incomplete builds, or more generally
+ * when multiple heads are passed to {@link #setup}, especially if the build contains nested parallelism.
+ * Consider using {@link DepthFirstScanner} instead in those cases to avoid these issues, employing its iteration order
+ * as needed to sort results (although its global iteration order is inconsistent, its relative iteration order among
+ * siblings at the same nesting level (e.g. parallel branches) is consistent and that is good enough for typical use
+ * cases such as producing a tree).
```
dwnusbaum, line 60: Note though that `DepthFirstScanner` has inconsistent behavior with regard to parallel branches: it visits them (and the current heads) first-to-last instead of last-to-first as would be natural given the overall end-to-start iteration order (I think for backwards compatibility with `FlowGraphWalker`). This means that although the relative order among siblings is meaningful, that order is opposite between parallel branches and other kinds of nodes at the same level, so a special case is necessary. This is easy to work around, but perhaps we should add a `DepthFirstScanner(boolean useNaturalOrder)` constructor that would avoid this inconsistency.

jglick (approved):

src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -1022,9 +1029,74 @@ public void inProgressParallelInParallel() throws Exception {
         SemaphoreStep.waitForStart("innerA/1", b);
         ForkScanner scanner = new ForkScanner();
         sanityTestIterationAndVisiter(b.getExecution().getCurrentHeads());
+        assertBranchOrder(b.getExecution(), "innerB", "innerA", "outerC", "outerB", "outerA");
```
jglick, line 1032: Flaky: https://github.com/jenkinsci/workflow-api-plugin/runs/16683329018

PR #310, opened 2023-09-11
Do not test branch ordering with ForkScanner for incomplete build with nested parallelism

dwnusbaum:
See https://github.com/jenkinsci/workflow-api-plugin/pull/306#discussion_r1321887545.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
@@ -63,7 +63,7 @@ import java.util.stream.StreamSupport;
 
 import org.junit.Ignore;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.equalTo;
 
 /**
  * Tests for internals of ForkScanner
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/graphanalysis/ForkScannerTest.java
```diff
@@ -1092,11 +1093,12 @@ public void parallelBranchOrdering() throws Exception {
...
-        List<String> branches = StreamSupport.stream(scanner.spliterator(), false)
+        String[] branches = StreamSupport.stream(scanner.spliterator(), false)
                 .map(n -> n.getPersistentAction(ThreadNameAction.class))
                 .filter(Objects::nonNull)
                 .map(ThreadNameAction::getThreadName)
-                .collect(Collectors.toList());
-        assertThat(branches, contains(expectedBranchNames));
+                .collect(Collectors.toList())
+                .toArray(String[]::new);
+        assertThat(branches, equalTo(expectedBranchNames));
```
dwnusbaum, line 1102: If it does fail again, this allows us to see the full order in the test failure message.

jglick: Do you mean to use `arrayEqualTo` or similar? I do not think `equalTo` will work on arrays generally.

dwnusbaum: There is a special case for arrays, see https://github.com/hamcrest/JavaHamcrest/blob/e4b9afdc2624bbebef3c7c739e6074ecd8a6c06e/hamcrest/src/main/java/org/hamcrest/core/IsEqual.java#L36-L38, which ends up being delegated to by matchers like `arrayContaining` (`arrayEqualTo` is not a thing as far as I can see). I can switch to `arrayContaining` though if you prefer.

Maybe you are thinking of `junit`'s `Assert.assertEquals` vs `Assert.assertArrayEquals`?

jglick (approved):

PR #309, opened 2023-09-11
[JENKINS-71970] Potentially solve memory leak

rsandell:
### Testing done

```[tasklist]
### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

PR #311, opened 2023-09-13
[JENKINS-71970] Memory leak involving `BufferedBuildListener`

jglick:
[JENKINS-71970](https://issues.jenkins.io/browse/JENKINS-71970), amending #309.

Reproduced by running `mvn hpi:run` with

```diff
diff --git pom.xml pom.xml
index c4482a2..646fe12 100644
--- pom.xml
+++ pom.xml
@@ -63,7 +63,7 @@
     </pluginRepositories>
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
-        <jenkins.version>2.361.4</jenkins.version>
+        <jenkins.version>2.414.1</jenkins.version>
         <no-test-jar>false</no-test-jar>
         <useBeta>true</useBeta>
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
@@ -72,8 +72,8 @@
         <dependencies>
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
-                <artifactId>bom-2.361.x</artifactId>
-                <version>2102.v854b_fec19c92</version>
+                <artifactId>bom-2.414.x</artifactId>
+                <version>2423.vce598171d115</version>
                 <scope>import</scope>
                 <type>pom</type>
             </dependency>
```

plus updating `workflow-support` from UC and installing `mock-slave`, creating a mock slave, and running

```groovy
node('static') {
    int i = 0
    while (true) {
        sh "echo ${i}"
        i++
    }
}
```

then

```bash
watch 'jmap -histo:live …
```

or

```bash
watch 'jmap -histo:live … | fgrep BufferedBuildListener'
```

There were multiple issues. First of all, removing the listener on `Channel.onClosed` is not enough (and may in fact be useless) when there are lots of different listeners being created while the channel is open. The listener needed to be associated with the stream (`BufferedBuildListener`), not its `Replacement` which is intended to be transient. Closing the stream (if that ever happens) should also remove the listener. And collecting the stream (which happens routinely) should remove the listener.

With the fix, I was able to run thousands of `sh` steps in a row on the same open agent channel, and memory did not increase significantly (in particular the `byte[]` usage would go up to ~10Mb but then drop again), nor did any classes from this package stay in the top screenful.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
@@ -47,6 +47,8 @@ import org.jenkinsci.remoting.SerializableOnlyOverRemoting;
  */
 final class BufferedBuildListener extends OutputStreamTaskListener.Default implements BuildListener, Closeable, SerializableOnlyOverRemoting {
 
+    private static final long serialVersionUID = 1;
+
     private static final Logger LOGGER = Logger.getLogger(BufferedBuildListener.class.getName());
 
     private static final Cleaner cleaner = Cleaner.create(new NamingThreadFactory(new DaemonThreadFactory(), BufferedBuildListener.class.getName() + ".cleaner"));
--- a/src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
@@ -29,12 +29,15 @@ import hudson.model.BuildListener;
 import hudson.remoting.Channel;
 import hudson.remoting.ChannelClosedException;
 import hudson.remoting.RemoteOutputStream;
+import hudson.util.DaemonThreadFactory;
+import hudson.util.NamingThreadFactory;
 import hudson.util.StreamTaskListener;
 import java.io.Closeable;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/log/BufferedBuildListener.java
```diff
@@ -46,10 +49,23 @@ final class BufferedBuildListener extends OutputStreamTaskListener.Default imple
 
     private static final Logger LOGGER = Logger.getLogger(BufferedBuildListener.class.getName());
 
+    private static final Cleaner cleaner = Cleaner.create(new NamingThreadFactory(new DaemonThreadFactory(), BufferedBuildListener.class.getName() + ".cleaner"));
+
     private final OutputStream out;
+    private final Channel channel;
+    private final Listener listener;
```
dwnusbaum, line 56: SpotBugs does not like these unserializable fields given that the class is `Serializable` although it does not matter in our case. I think they can be transient though:

```suggestion
    private transient final Channel channel;
    private transient final Listener listener;
```

dwnusbaum (approved):

car-roll (approved):

rsandell (approved):

PR #321, opened 2024-03-21
Take out a file lock on the `log` file

jglick:
Just experimenting at this point. Does not seem to regress the test added in #296.


jglick:
Appears to be harmless, but unnecessary.

PR #327, opened 2024-05-16
Fix origin identification for unserializable exceptions

dwnusbaum:
For exceptions which were not serializable, we previously only added `ErrorId` to the `ProxyException` created for serialization purposes. The original exception that was thrown up through the Pipeline's execution would have no `ErrorId`, so each block/step that failed due to the exception would get a fresh `ErrorId`.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
@@ -42,7 +42,7 @@ import org.apache.commons.io.output.NullOutputStream;
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
 import org.jenkinsci.plugins.workflow.graph.AtomNode;
 import org.jenkinsci.plugins.workflow.graph.BlockEndNode;
-import org.jenkinsci.plugins.workflow.graphanalysis.ForkScanner;
+import org.jenkinsci.plugins.workflow.graphanalysis.DepthFirstScanner;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.Beta;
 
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/actions/ErrorActionTest.java
```diff
@@ -245,4 +247,15 @@ public static class X extends Exception {
         assertNotNull(new ErrorAction(unserializable));
         assertNotNull(new ErrorAction(cyclic));
     }
+
+    @Test public void findOriginOfRethrownUnserializableException() throws Throwable {
```
dwnusbaum, line 251: Fails without the changes in `src/main`.

src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
```diff
@@ -140,7 +145,7 @@ public String getUrlName() {
      */
     public static @CheckForNull FlowNode findOrigin(@NonNull Throwable error, @NonNull FlowExecution execution) {
         FlowNode candidate = null;
-        for (FlowNode n : new ForkScanner().allNodes(execution)) {
+        for (FlowNode n : new DepthFirstScanner().allNodes(execution)) {
```
dwnusbaum, line 148: Unrelated, but there is no reason to use `ForkScanner` here and it has various potential issues if used on an incomplete build.

jglick (approved):
Looks right.

jglick:
…but does not suffice according to #328. Digging into it.

src/main/java/org/jenkinsci/plugins/workflow/actions/ErrorAction.java
dwnusbaum: Actually this is not ok, but I guess we have no tests for the relevant case. We need to make sure that we only replace the candidate if the new node is a parent of the candidate node. I guess I'll just revert this for now.

dwnusbaum: https://github.com/jenkinsci/workflow-api-plugin/pull/327/commits/28d0acd535d3008120412de33819115953edf710

jglick:
Oops, you are making changes here but I am working on changes in the more comprehensive test downstream, so this is all going to clash.

dwnusbaum:
It's fine, we can close this.

PR #343, opened 2024-07-18
Small change to pipeline durability description

julieheard:
I have changed the wording on the durability settings to make sure users know that the settings for 'max survivability' and 'max durability' are actually the same thing. Previously to this change, the monitor displays the name of the enum instead of the description for it, making some users think there are 4 different options instead of 3.

I removed the (previously the only option) bit as could'nt see any more info on that in the documentation [here](https://www.jenkins.io/doc/book/pipeline/scaling-pipeline/#what-are-the-durability-settings) and wanted to just trim and make clearer.

Before:
image

After:
image

This change is only description changes on the UI, so nothing to test other than manual tests to see the different wording. I also have a draft docs PR [here](https://github.com/jenkins-infra/jenkins.io/pull/7410) to sync the wording from this PR.

### Submitter checklist


jglick (approved):
Thanks, this looks clearer indeed.

PR #347, opened 2024-08-06
Prevent `StepExecutionIterator` from leaking memory in cases where a single processed execution has a stuck CPS VM thread

dwnusbaum:
I recently saw a real world case (ref. SECO-3962) where a Pipeline build’s flow graph had a loop due to some type of corruption (I suspect a crash while the Pipeline was running lead to a `build.xml` whose `iota` and `head` were older than the most recent `FlowNode` in `workflow/` and its `StepExecution` in `program.dat`), which caused its `CpsVmExecutorService` to get stuck in an infinite loop on a single task. This caused all calls to `StepExecution.applyAll` to leak all live `WorkflowRun`s at the time of the call because `StepExecution`s for those builds were strongly reachable from the CPS VM thread for the stuck build via the `SingleLaneExecutorService` task queue for the `CpsVmExecutorService` and the internals of Guava’s `Futures.allAsList`.

<details>
<summary>Here is the leak path (produced via <code>MemoryAssert</code>)</summary>

```
[Ljava.lang.Thread;@52454457-[38]->
java.lang.Thread@6dbe3b4f-threadLocals->
java.lang.ThreadLocal$ThreadLocalMap@3b25709e-table->
[Ljava.lang.ThreadLocal$ThreadLocalMap$Entry;@32a29361-[13]->
java.lang.ThreadLocal$ThreadLocalMap$Entry@cdf3ee2-value->
// This CpsVmExecutorService is for the stuck build.
org.jenkinsci.plugins.workflow.cps.CpsVmExecutorService$$Lambda$775/0x000000030078f4d0@3b60efd8-arg$1->
org.jenkinsci.plugins.workflow.cps.CpsVmExecutorService@50b736a0-base->
hudson.remoting.SingleLaneExecutorService@72e09a66-tasks->
java.util.concurrent.LinkedBlockingQueue@541e333d-last->
java.util.concurrent.LinkedBlockingQueue$Node@696552a4-item->
java.util.concurrent.FutureTask@1343f77d-callable->
java.util.concurrent.Executors$RunnableAdapter@774ecd5a-task->
org.jenkinsci.plugins.workflow.cps.CpsVmExecutorService$$Lambda$772/0x000000030078afc8@6d14d82b-arg$3->
org.jenkinsci.plugins.workflow.cps.CpsFlowExecution$4$1@6dd1f638-this$1->
org.jenkinsci.plugins.workflow.cps.CpsFlowExecution$4@2cbc5084-val$callback->
org.jenkinsci.plugins.workflow.cps.CpsFlowExecution$5@2b4abaaa-val$r->
// These references through futures are from the usage of `Futures.allAsList` in `FlowExecutionList$StepExecutionIteratorImpl.applyAll`
com.google.common.util.concurrent.SettableFuture@371f2f2a-listeners->
com.google.common.util.concurrent.AbstractFuture$Listener@7effd655-task->
com.google.common.util.concurrent.AggregateFuture$$Lambda$815/0x00000003007eb4b8@8f2f390-arg$1->
com.google.common.util.concurrent.CollectionFuture$ListFuture@752edffc-values->
java.util.ArrayList@741fc2e0-elementData->
[Ljava.lang.Object;@460cb18f-[1]->
com.google.common.util.concurrent.CollectionFuture$Present@14a432cd-value->
java.util.Collections$UnmodifiableRandomAccessList@72240090-list->
java.util.ArrayList@53dece06-elementData->
[Ljava.lang.Object;@67e46851-[0]->
// This step and its run are from a completed build that is _not_ the stuck build, which causes the memory leak.
org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep$Execution@5664556b-context->
org.jenkinsci.plugins.workflow.cps.CpsStepContext@2cd3b80f-executionRef->
org.jenkinsci.plugins.workflow.job.WorkflowRun$Owner@4d626ff-run->
org.jenkinsci.plugins.workflow.job.WorkflowRun@43a80df5}
```
</details>

There are a lot of things that could be improved here, but I am not sure where our effort is best spent given that I have not previously seen corruption cause an infinite loop like this:

1. **This PR**: Change the result type of `StepExecutionIteratorImpl.applyAll` to `ListenableFuture<List<Void>>` instead of `ListenableFuture<List<StepExecution>>`. The result type in the API is `ListenableFuture<?>`, and the return value is only intended to be used to track progress and cancel the operation if needed, so this change should be compatible. This change means the combined future drops its references to each build's `StepExecution`s as soon as they have been processed, rather than having to hold a reference to them until _all_ build's `StepExecution`s have been processed.
2. We could add infinite loop detection to all Pipeline flow graph iteration constructs, e.g. `LinearBlockHoppingScanner`, `LinearScanner`, `DepthFirstScanner`, `ForkScanner`, `StandardGraphLookupView.bruteForScanForEnd`, etc., and throw an exception whenever any infinite loop is encountered.
    * https://github.com/jenkinsci/workflow-api-plugin/pull/349 investigates this further.
4. We could add monitoring to `CpsVmExecutorService` to warn when a build’s `SingleLaneExecutorService` seems to be having problems, e.g. the queue size is only increasing, the executing task has been live for more than 15 minutes, tasks are starting more than 15 minutes after they were submitted, etc.
    * This could be reported in support bundles, an admin monitor, `metrics`, etc.
    * https://github.com/jenkinsci/workflow-cps-plugin/pull/916 explores this idea further.
5. We could add an admin monitor and/or support bundle components that report when a build has been live for more than (for example) 7 days, so that admins are aware that something might be wrong and they can try to kill the build (at the time I investigated the problematic build, it was around 3.5 months old).
6. Similarly, we could change Pipeline resumption logic to not attempt to resume any build started more than (for example) 14 days ago (at the time I investigated the build, it had tried to resume 16 times over the 3.5 month period).

### Submitter checklist


src/test/resources/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest/stepExecutionIteratorDoesNotLeakBuildsWhenOneIsStuck/jobs/test0/builds/1/workflow/7.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version='1.1' encoding='UTF-8'?>
+<Tag plugin="workflow-support@907.v6713a_ed8a_573">
+  <node class="cps.n.StepEndNode" plugin="workflow-cps@3894.3896.vca_2c931e7935">
+    <parentIds>
+      <string>6</string>
+    </parentIds>
+    <id>7</id>
+    <startId>8</startId> <!-- Manually edited to simulate corruption observed in the real world. -->
```
dwnusbaum, line 8: In the real build that I saw, this pointed to the `lock` step from `lockable-resources`, and there were no other nodes past that `StepStartNode`. I am still quite curious as to how we got a `startId` that was newer than `id` though whenever things got corrupted for the first time.

src/test/resources/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest/stepExecutionIteratorDoesNotLeakBuildsWhenOneIsStuck/jobs/test0/builds/1/build.xml
```diff
@@ -0,0 +1,46 @@
...
+      </entry>
+      <entry>
+        <string>saveProgram</string>
+        <long>57936125</long>
+      </entry>
+    </timings>
+    <internalCalls class="sorted-set"/>
+    <sandbox>true</sandbox>
+    <iota>5</iota>
+    <head>1:5</head>
```
dwnusbaum, line 39: These are notably older than the latest nodes in `workflow/`. In the real build that I saw, there was also a partially-written `atomic<numbers>.tmp` file for the build, indicative of some kind of uncontrolled Jenkins shutdown.

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
```diff
@@ -160,6 +165,62 @@ public class FlowExecutionListTest {
...
+            var b = p.getBuildByNumber(1);
+            // TODO: Create and run a new build, call StepExecutionIterator.applyAll while it is running, and assert
+            // that we do not leak a hard reference to it through the task queue for the CpsVmExecutorService for the
+            // stuck build and the futures inside of StepExecutionIteratorImpl.applyAll.
+            // TODO: Ideally, we would detect the issue with the stuck build and find some way to kill it.
+            r.waitForCompletion(b);
+        });
+    }
+
+    @Ignore("MemoryAssert is unable to clear the SoftReference from AbstractLazyLoadRunMap. Heap dumps show a strong reference from the test execution thread as a local variable, which is not captured by MemoryAssert. Also needs to be moved to a class with no BuildWatcher.")
```
dwnusbaum, line 194: I was hoping to get this test to work since I'm not really sure if there is any other way to test the fix. It isn't totally clear to me why `MemoryAssert` isn't working. I think the strong reference from a local variable in the test thread that I saw might just be transient, for example I saw it referenced while the thread was at [this line](https://github.com/jenkinsci/jenkins-test-harness/blob/3585fb3db90fff5374405e4c1149863825147959/src/main/java/org/jvnet/hudson/test/MemoryAssert.java#L171). I am not sure though. I also tried configuring `BuildReference` to use weak references and saw the same behavior. Maybe I need to explicitly null out the local variable in the test.

dwnusbaum: > Maybe I need to explicitly null out the local variable in the test.

🤦 that seems to fix it

dwnusbaum: [b2c93a4](https://github.com/jenkinsci/workflow-api-plugin/pull/347/commits/b2c93a4ac1a2f8d957f440a4704924204c1ad8ef)

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
```diff
@@ -160,6 +165,58 @@ public class FlowExecutionListTest {
         });
     }
 
+    @Ignore("Build never completes due to infinite loop")
+    @LocalData
+    @Test public void stepExecutionIteratorDoesNotLeakBuildsWhenOneIsStuckInfiniteLoop() throws Throwable {
```
dwnusbaum, line 170: I am not sure yet if I want to do anything about infinite loops directly. If so, I will split this into a new PR and test various forms of infinite loops, and if not, I will just delete this test since the other one is simpler and tests the memory leak.

dwnusbaum: Moved into #349 and removed here in [7100611](https://github.com/jenkinsci/workflow-api-plugin/pull/347/commits/71006111e7df33fbcc0cde1e8a81569127953492).

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
```diff
@@ -160,6 +163,32 @@ public class FlowExecutionListTest {
...
+            // Make FlowExecutionList$StepExecutionIteratorImpl.applyAll submit a task to the CpsVmExecutorService
+            // for stuck #1 that will never complete, so the resulting future will never complete.
+            StepExecution.applyAll(e -> null);
+            // Let notStuckBuild complete and check that it can be GC'd.
+            SemaphoreStep.success("wait/1", null);
+            r.waitForCompletion(notStuckBuild);
+            notStuckBuild = null; // Clear out the local variable in this thread.
+            Jenkins.get().getQueue().clearLeftItems(); // We don't want to wait 5 minutes.
+            MemoryAssert.assertGC(notStuckBuildRef, true);
+            // TODO: Test cleanup hangs for 1 minute in CpsFlowExecution.suspendAll because the checkpoint task can't run.
```
dwnusbaum, line 188: I'm not sure if there is a good way to interrupt the VM thread here so the test can complete quickly. We can probably make the Pipeline call a static method in the test class that checks for us to set some flag here before continuing or something like that, or create a custom synchronous step that does something similar.

dwnusbaum: Improved in [2de670d](https://github.com/jenkinsci/workflow-api-plugin/pull/347/commits/2de670dda3e6bdfa207e6c2b5bbcb2b56129a239) so that the test only takes 5ish seconds.

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
```diff
@@ -160,6 +164,34 @@ public class FlowExecutionListTest {
...
+            // Make FlowExecutionList$StepExecutionIteratorImpl.applyAll submit a task to the CpsVmExecutorService
+            // for stuck #1 that will never complete, so the resulting future will never complete.
+            StepExecution.applyAll(e -> null);
+            // Let notStuckBuild complete and clean up all references.
+            SemaphoreStep.success("wait/1", null);
+            r.waitForCompletion(notStuckBuild);
+            notStuckBuild = null; // Clear out the local variable in this thread.
+            Jenkins.get().getQueue().clearLeftItems(); // Otherwise we'd have to wait 5 minutes for the cache to be cleared.
+            // Make sure that the reference can be GC'd.
+            MemoryAssert.assertGC(notStuckBuildRef, true);
```
dwnusbaum, line 188: Revert the fix and this will fail. The PR description shows the path to the strong reference.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -253,24 +253,25 @@ public ListenableFuture<?> apply(final Function<StepExecution, Void> f) {
...
                         }
                     }
-
-                    @Override
-                    public void onFailure(@NonNull Throwable t) {
-                        LOGGER.log(Level.WARNING, null, t);
-                    }
+                    return null;
+                }, MoreExecutors.directExecutor());
+                ListenableFuture<Void> resultsWithWarningsLogged = Futures.catching(results, Throwable.class, t -> {
```
jglick, line 270: Yikes. Project Loom would be very welcome in code like this.

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
```diff
@@ -198,14 +230,59 @@ public String getFunctionName() {
...
+            blocked.put(id, State.UNBLOCKED);
+        }
+
+        private enum State {
+            NOT_STARTED,
+            BLOCKED,
+            UNBLOCKED,
+        }
+
+        @TestExtension public static class DescriptorImpl extends StepDescriptor {
```
jglick, line 276:
```suggestion
        @TestExtension("stepExecutionIteratorDoesNotLeakBuildsWhenOneIsStuck") public static class DescriptorImpl extends StepDescriptor {
```
?

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
```diff
@@ -160,6 +164,34 @@ public class FlowExecutionListTest {
...
+            var notStuck = r.createProject(WorkflowJob.class, "not-stuck");
+            notStuck.setDefinition(new CpsFlowDefinition("semaphore 'wait'", true));
+            var notStuckBuild = notStuck.scheduleBuild2(0).waitForStart();
+            SemaphoreStep.waitForStart("wait/1", notStuckBuild);
+            WeakReference<Object> notStuckBuildRef = new WeakReference<>(notStuckBuild);
+            // Create a Pipeline that runs a long-lived task on its CpsVmExecutorService, causing it to get stuck.
+            var stuck = r.createProject(WorkflowJob.class, "stuck");
+            stuck.setDefinition(new CpsFlowDefinition("blockSynchronously 'stuck'", false));
+            var stuckBuild = stuck.scheduleBuild2(0).waitForStart();
+            await().atMost(5, TimeUnit.SECONDS).until(() -> SynchronousBlockingStep.isStarted("stuck"));
```
jglick, line 178: Or `waitForMessage` would probably suffice.

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
```diff
@@ -160,6 +164,34 @@ public class FlowExecutionListTest {
...
+            // Make FlowExecutionList$StepExecutionIteratorImpl.applyAll submit a task to the CpsVmExecutorService
+            // for stuck #1 that will never complete, so the resulting future will never complete.
+            StepExecution.applyAll(e -> null);
```
jglick, line 181: Should this return value be kept in a local variable?

jglick (approved):

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
dwnusbaum: It's not necessary for the leak, if that's what you mean. Even if the `StepExecution.applyAll` caller completely ignores the return value, the stuck CPS VM thread holds a strong reference to the aggregate future because the `SingleLaneExecutorService`'s task queue has a reference to the `SettableFuture` for the `getCurrentExecutions` call, and then that future references the aggregate future via a listener just due to the implementation of `Futures.allAsList`.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
dwnusbaum: Looking more carefully through Guava's Javadoc I think I can use `FluentFuture` to simplify this.

dwnusbaum: Actually `FluentFuture` doesn't really make things any clearer, so I will leave it.

jglick (approved):

PR #349, opened 2024-08-08
Avoid infinite loops due to corrupted flow graphs in some cases and improve resumption error handling

dwnusbaum:
Related to #347. This PR tries to improve these main issues:
1. Infinite loops in `StandardGraphLookupView.bruteForceScanForEnclosingBlock` and `LinearBlockHoppingScanner` are now detected and throw an exception. This does not fix all possible infinite loops with flow graph iteration APIs, but these are the only two cases I saw where the infinite loop happens internally rather than the caller actually iterating over nodes infinitely. Loops in `StandardGraphLookupView.bruteForceScanForEnclosingBlock` via `PlaceholderTask.getAffinityKey` brick the build queue, and loops in `LinearBlockHoppingScanner` brick the  `CpsVmExecutorService` for the build in question.
2. When we try to resume a Pipeline but run into something that prevents steps from resuming, we should do what we can to ensure the build does not hang forever waiting for steps to resume.

### Submitter checklist


src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -362,20 +362,31 @@ private static final class ParallelResumer {
...
+                    scanner.setup(n);
+                    for (FlowNode parent : scanner) {
+                        if (parent != n && nodes.containsKey(parent)) {
+                            enclosing.put(n, parent);
+                            break;
+                        }
                     }
                 }
+            } catch (Exception e) {
+                // TODO: Should we instead just try to resume steps with no respect to topological order?
```
dwnusbaum, line 384: Not sure if this would be better or worse than killing all of the steps. If you end up here, something is pretty wrong with your flow graph. The main causes I can think of are if a `BlockEndNode.startId` points to a node that either doesn't exist, is not a `BlockStartNode`, or (the new case I just saw) points to a `BlockStartNode` with a newer ID than the end node, creating a cycle in the graph.

The question though is whether flow graph corruption necessarily means that trying to resume the steps is pointless, or if there is a decent chance that if we fall back to synchronous resumption of each step execution in order, the build might be able to complete successfully.

jglick: Ah I was just looking for #221 the other day as a reference, because its original motivating use case is obsolete as of https://github.com/jenkinsci/ssh-agent-plugin/pull/152. Of course there might still be other steps which legitimately need context from an enclosing step in their `onResume`, but I am not aware of any offhand. So I think it would be reasonable to fall back to resuming all steps in random order. I would not expect the build to complete successfully, but it might be able to e.g. release external resources more gracefully.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
```diff
@@ -362,20 +362,31 @@ private static final class ParallelResumer {
                         nodes.put(n, se);
                     } else {
                         LOGGER.warning(() -> "Could not find FlowNode for " + se + " so it will not be resumed");
+                        // TODO: Should we call se.getContext().onFailure()?
```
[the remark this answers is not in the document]
dwnusbaum: Well, it's kind of interesting. Specifically in the case of a missing or corrupt `FlowNode`, there are two cases that I see:
1. If `FlowExecution.heads` contains the bad node, then everything gets handled in `CpsFlowExecution.onLoad`, we never read `program.dat`, we create placeholder nodes, the Pipeline fails, great.
2. If `FlowExecution.heads` does not contain the bad node, then the Pipeline loads, we attempt to resume it, various warnings get logged, and it hangs forever. The problem is that [this call in CpsStepContext](https://github.com/jenkinsci/workflow-cps-plugin/blob/a6f67a8c907b4f3699bd54b46271e0ecd3caccc4/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java#L368) throws right away, so the `CpsThread` never resumes with the outcome, and the Pipeline hangs. It seems like [this log message should at least be bumped to `WARNING`](https://github.com/jenkinsci/workflow-cps-plugin/blob/a6f67a8c907b4f3699bd54b46271e0ecd3caccc4/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsStepContext.java#L438) because any errors there are likely to be the proximate cause of a build hanging forever, but I wonder if we should also attempt to call `CpsFlowExecution.croak` or `CpsVmExecutorService.reportProblem` in that catch block.

So long story short, it seems like there is no point trying to do anything special in these cases in this plugin.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
dwnusbaum: [c6b4fa9](https://github.com/jenkinsci/workflow-api-plugin/pull/349/commits/c6b4fa940e2289e124f88a1ffb929528b4ef202d) adjusts things so we resume all steps as long as they have a `FlowNode`. https://github.com/jenkinsci/workflow-api-plugin/pull/349#discussion_r1712183413 discusses the cases where the `FlowNode` is missing and/or corrupted.

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionList.java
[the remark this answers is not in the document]
dwnusbaum: https://github.com/jenkinsci/workflow-cps-plugin/pull/916 updates the relevant log message to `WARNING` so that we have an idea if the problematic situation is ever occurring in practice.

jglick (approved):

PR #370, opened 2025-01-07
`FlowExecutionList.Storage`

jglick:
Subsumes #368 & #369. Supersedes #312. Extension point used by CloudBees CI in HA mode to implement alternate behavior.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionOwner.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionOwner.java
@@ -100,6 +100,25 @@ public abstract class FlowExecutionOwner implements Serializable {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Vlatombe (approved):

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionOwner.java
```diff
@@ -99,6 +101,25 @@ public String getUrlOfExecution() throws IOException {
         return getUrl()+"execution/";
     }
 
+    /**
+     * The {@link Run#getExternalizableId}, if this owner is indeed a {@link Run}.
```
jglick, line 105: Called by the CloudBees CI implementation.

dwnusbaum (approved):

PR #371, opened 2025-01-10
In case of a large FlowExecutionList, logging everything to warning level is too verbose

Vlatombe:
Only log the details if there is actually `FINE` level enabled.

### Submitter checklist
- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did


dwnusbaum (approved):

jglick:
Actually in #370 I am just unconditionally disabling the complete list, as I do not suppose it is useful, and in fact potentially has side effects.

PR #383, opened 2025-03-03
`FlowExecution#getAuthentication2`

basil:
jenkinsci/jenkins#4848

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
@@ -47,10 +47,10 @@ import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import jenkins.model.Jenkins;
 import jenkins.model.queue.AsynchronousExecution;
-import org.acegisecurity.Authentication;
 import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
+import org.springframework.security.core.Authentication;
 
 /**
  * State of a currently executing workflow.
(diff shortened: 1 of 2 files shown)


jglick:
Looks right. Is there a planned impl in https://github.com/jenkinsci/workflow-cps-plugin/blob/d58c465ea71a1d94b1ad827b885b1a5193c9adac/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java#L1566-L1582 https://github.com/jenkinsci/workflow-cps-plugin/blob/d58c465ea71a1d94b1ad827b885b1a5193c9adac/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsVmExecutorService.java#L172 ? (see https://github.com/jenkinsci/pipeline-plugin/commit/9abf40ae442e650ff9a30c9a8e5973ced853f1d1)

basil:
@jglick Yes, such an implementation is planned but not currently assigned to any particular person.

jglick:
That is fine but this should probably stay on hold until the downstream PR has been prepared, as it is technically introducing a new API.

basil:
> That is fine but this should probably stay on hold until the downstream PR has been prepared

https://github.com/jenkinsci/jenkins/pull/4848 was not held back by the lack of downstream PRs (cf. https://github.com/jenkinsci/reverse-proxy-auth-plugin/pull/153#issuecomment-2471726084), so I see no reason why this PR should stay on hold.

jglick:
There were _some_; this method is both implemented and called in only one place (AFAIK), `workflow-cps`, so it only makes sense to evolve it in the context of those two sites.

(Actually `workflow-cps` could define and call `getAuthentication2` even before it is defined in `FlowExecution`; it would only be the `@Override` annotation that would depend on this.)

basil:
bq. so it only makes sense to evolve it in the context of those two sites

Hence the plan to update those two call sites eventually. Is there a particular reason that has to be done as a prerequisite to merging this PR?

jglick:
Yes: matching changes should be prepared in advance where feasible. Filed https://github.com/jenkinsci/workflow-cps-plugin/pull/1002 and will merge this if it passes tests.

jglick (approved):

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
```diff
@@ -248,10 +248,27 @@ public boolean blocksRestart() {
...
+                org.acegisecurity.Authentication.class,
+                getClass(),
+                "getAuthentication");
+    }
 
+    /**
+     * @deprecated use {@link #getAuthentication2()}
+     */
+    @Deprecated
+    public /* abstract */ @NonNull org.acegisecurity.Authentication getAuthentication() {
```
cyrille-leclerc, line 265: This commit causes:
```
java.lang.IllegalArgumentException: The specified derived class (org.jenkinsci.plugins.workflow.cps.CpsFlowExecution) does not derive from the specified base class (org.springframework.security.core.Authentication).
	at hudson.Util.isOverridden(Util.java:1536)
	at hudson.Util.ifOverridden(Util.java:1564)
	at PluginClassLoader for workflow-api//org.jenkinsci.plugins.workflow.flow.FlowExecution.getAuthentication(FlowExecution.java:266)
	at PluginClassLoader for opentelemetry//io.jenkins.plugins.opentelemetry.job.MonitoringPipelineListener.onAtomicStep(MonitoringPipelineListener.java:209)
```

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
```diff
@@ -248,10 +248,27 @@ public boolean blocksRestart() {
...
+        return Util.ifOverridden(
+                () -> org.acegisecurity.Authentication.fromSpring(getAuthentication2()),
+                Authentication.class,
+                getClass(),
```
cyrille-leclerc, line 269: `getClass()` will return values like `org.jenkinsci.plugins.workflow.cps.CpsFlowExecution` which do NOT derive from `org.springframework.security.core.Authentication`

src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
cyrille-leclerc: @basil  @jglick can you please check https://issues.jenkins.io/browse/JENKINS-75532

basil: ⇨ #391

PR #387, opened 2025-03-28
Delete `GCFlushedOutputStream`

dwnusbaum:
See https://github.com/jenkinsci/workflow-api-plugin/pull/83#discussion_r2018964003. Alternative to https://github.com/jenkinsci/workflow-api-plugin/pull/388.

### Testing done

Existing tests here pass for me locally, but if we want to go through with this we need to run it through the PCT.

### Submitter checklist


jglick:
Prefer #388 but I only have a dim memory of what this was (supposed to be) solving. Perhaps remote `TaskListener`s which never get `flush`ed before the callable exits?

PR #388, opened 2025-03-28
Make `GCFlushedOutputStream` work as intended by switching to `Cleaner`

dwnusbaum:
See https://github.com/jenkinsci/workflow-api-plugin/pull/83#discussion_r2018964003.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/log/GCFlushedOutputStream.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/log/GCFlushedOutputStream.java
@@ -28,18 +28,17 @@ import java.io.BufferedOutputStream;
 import java.io.FilterOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
-import java.lang.ref.PhantomReference;
-import java.lang.ref.ReferenceQueue;
-import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.remoting.ChannelClosedException;
+import hudson.remoting.DaemonThreadFactory;
+import hudson.remoting.NamingThreadFactory;
 import java.io.EOFException;
+import java.lang.ref.Cleaner;
 import java.nio.channels.ClosedChannelException;
 import java.util.stream.Stream;
-import jenkins.util.Timer;
 
 /**
  * A stream which will be flushed before garbage collection.
@@ -48,10 +47,12 @@ import jenkins.util.Timer;
 final class GCFlushedOutputStream extends FilterOutputStream {
     
     private static final Logger LOGGER = Logger.getLogger(GCFlushedOutputStream.class.getName());
+    private static final Cleaner CLEANER = Cleaner.create(
+            new NamingThreadFactory(new DaemonThreadFactory(), GCFlushedOutputStream.class.getName() + ".CLEANER"));
 
     GCFlushedOutputStream(OutputStream out) {
         super(out);
-        FlushRef.register(this, out);
+        CLEANER.register(this, new CleanerTask(out));
     }
 
     @Override public void write(@NonNull byte[] b, int off, int len) throws IOException {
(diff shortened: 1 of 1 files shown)


jglick:
> run it through the PCT

and `pipeline-cloudwatch-logs` (need an AWS account to run and some vars like `AWS_PROFILE`, `AWS_REGION`, `AWS_CHAINED_ROLE`, `CLOUDWATCH_LOG_GROUP_NAME`)

dwnusbaum:
I finally had a bit of time to test this. As far as I can see, `pipeline-cloudwatch-logs` does not care either way, because it does not use `BufferedBuildListener`, `GCFlushedOutputStream`, or `DelayBufferedOutputStream`. I did run its tests though and they passed with this PR as well as #387. Tests in `workflow-durable-task-step` also passed with both PRs.

The only way I could figure out to make `LogStorageTestBase#remoting` require `GCFlushedOutputStream` to pass with this PR was as follows:
* Replace all uses of `DelayBufferedOutputStream` with `BufferedOutputStream`
    * `FileLogStorageTest#remoting` was still passing
* Remove the call to `flush` in `LogStorageTestBase$RemotePrint`
    * `FileLogStorageTest#remoting` was failing consistently because the remote log output was never received by the controller
* Adjust `LogStorageTestBase$GC` to call `System#gc` repeatedly in a loop for around 5 seconds so that the cleanup task would run (I tried to investigate why this was needed, but acquiring a heap dump consistently triggered the cleanup task immediately. Calling `System#gc` once and sleeping for a long time did not consistently trigger cleanup)
    * `FileLogStorageTest#remoting` was passing, but flaky, because the listeners in the two `RemotePrint` calls were being flushed in random order
*  Add a call to `GC` after the first `RemotePrint` call, to ensure that it gets flushed before the second `RemotePrint` call
    * `FileLogStorageTest#remoting` was passing consistently

I then went back and tested the same test changes without this PR.
* Initially, `FileLogStorageTest#remoting` was flaky, because the listeners in the two `RemotePrint` calls were being flushed in random order
* I added a 10 second sleep call after the two `RemotePrint` calls (to give the timer task in `FlushRef` time to run
    * `FileLogStorageTest#remoting` was now passing consistently
* I then removed all calls to `GC`, so that the test no longer called `System#gc` at all
   * `FileLogStorageTest#remoting` was still passing consistently (!)
* I then deleted the 10 second sleep after the two `RemotePrint` calls, and instead added a 10 second sleep inside of the two `RemotePrint` calls, at the end of which I printed a log message that referenced the `listener`, to ensure that it could not be GC'd in that time frame
  * `FileLogStorageTest#remoting` was still passing consistently, and the log messages showed that `FlushRef` was running the cleanup task while the stream was still reachable  (!) 

So, here are my key takeaways:

* My suspicions in https://github.com/jenkinsci/workflow-api-plugin/pull/83#discussion_r2018964003 seem to be correct: The current behavior of `GCFlushedOutputStream` is a 1-time flush within 10 seconds of the the stream's creation, regardless of the stream's reachability. I suspect the reason that #83 fixed [JENKINS-54566](https://issues.jenkins-ci.org/browse/JENKINS-54566) was just that it deleted the problematic `finalize` override.
* I can create a test that specifically requires `GCFlushedOutputStream` for `FileLogStorage` by using `DelayBufferedOutputStream.Tuning` to disable the auto-flushing and adding an option to `RemotePrint` to not flush after writing, but it will be a pretty artificial test
    * `pipeline-cloudwatch-logs` would pass this test despite having no equivalent to `GCFlushedOutputStream`, since the only buffer that plugin uses is a `LineTransformationOutputStream`, so we'd have to write a line with no line ending to see a difference in behavior, but that starts to feel _extremely_ artificial
    * `cloudbees-pipeline-explorer` should pass this test due to reuse of `BufferedBuildListener`
    * IDK about `opentelemetry`
*  It is not clear to me whether there is a real-world case where `GCFlushedOutputStream` actually matters. Given that we haven't run into any known problems since #83, I am still somewhat inclined to just delete this class.

src/test/java/org/jenkinsci/plugins/workflow/log/LogStorageTestBase.java
```diff
@@ -202,22 +273,32 @@ protected Map<String, Level> agentLoggers() {
...
+                }
+                case GcFlushOnly -> {
+                    listener = null;
+                    // Sleeping and calling `System.gc` more than once seem to be necessary for Cleaner to run reliably.
+                    for (int i = 0; i < 3; i++) {
+                        System.gc();
+                        Thread.sleep(1000);
+                    }
+                }
+            }
```
dwnusbaum, line 301: Note that in real-world scenarios, where we do not nicely wait after writing each individual line, there is no guarantee of ordering with `GCFlushedOutputStream` or `DelayBufferedOutputStream`, and so you can end up with lines in unexpected orders.

src/test/java/org/jenkinsci/plugins/workflow/log/LogStorageTestBase.java
```diff
@@ -171,9 +186,66 @@ protected static void close(TaskListener listener) throws Exception {
...
+            DelayBufferedOutputStream.Tuning.DEFAULT.maxRecurrencePeriod = TimeUnit.HOURS.toMillis(1);
+        }
+        if (variant.disableGcFlush) {
+            GCFlushedOutputStream.DISABLED = true;
+        }
         LogStorage ls = createStorage();
         TaskListener overall = ls.overallListener();
+        if (variant != RemotingTestVariant.Regular) {
+            assumeThat("Skipping BufferedBuildListener-specific tests because listener is " + overall,
+                    overall, instanceOf(BufferedBuildListener.class));
```
dwnusbaum, line 247: This is unnecessary, i.e. the new tests pass against `pipeline-cloudwatch-logs`, but it seemed somewhat pointless to run tests that configure `BufferedBuildListener` against unrelated listener implementations.

If desired, we could instead move the new tests down into `FileLogStorageTest` add make the tests here more generic. For example, we'd just have one test that checks that if you write a line and do not flush, the line gets written if you wait for a few seconds. `pipeline-cloudwatch-logs` would still pass this test without waiting because it doesn't buffer more than a single line. `FileLogStorage` would only pass after a few seconds due to `DelayBufferedOutputStream`.

jglick: > if you write a line and do not flush, the line gets written if you wait for a few seconds

Fine, I guess, but no one should rely on that; remote code printing to a `TaskListener` is expected to flush explicitly.

PR #406, opened 2025-06-11
Allow multiple LogStorage with primary and secondaries

jgreffe:
https://github.com/user-attachments/assets/63560719-2765-490e-b3e8-f1ac03a6a05b

### Submitter checklist
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed


src/main/java/org/jenkinsci/plugins/workflow/log/OutputStreamTaskListener.java
```diff
@@ -63,7 +65,7 @@ public interface OutputStreamTaskListener extends TaskListener {
             return ((OutputStreamTaskListener) listener).getOutputStream();
         }
         PrintStream ps = listener.getLogger();
-        if (ps.getClass() != PrintStream.class) {
+        if (ps.getClass() != PrintStream.class && ps.getClass() != TeePrintStream.class) {
```
dwnusbaum, line 68: The changes here don't seem right. I think instead we should handle `TeeTaskListener` and/or `TeePrintStream` directly and have it return a `TeeOutputStream` where possible, using this method recursively as needed.

src/test/java/org/jenkinsci/plugins/workflow/log/tee/TeeLogStorageTestBase.java
```diff
@@ -0,0 +1,38 @@
...
+import org.jenkinsci.plugins.workflow.log.LogStorage;
+import org.jenkinsci.plugins.workflow.log.LogStorageTestBase;
+import org.junit.Before;
+import org.junit.ClassRule;
+import org.junit.rules.TemporaryFolder;
+
+/**
+ * Foundation for compliance tests of {@link TeeLogStorage} implementations.
+ */
+public abstract class TeeLogStorageTestBase extends LogStorageTestBase {
```
dwnusbaum, line 14: I know I recommended this, but now that I see how the API works, I think this approach is not useful in the way I thought it would be. I was thinking that each implementation would implement `TeeLogStorage`, but really they just implement `TeeLogStorageFactory`, and there is just one `TeeLogStorage`. I think you can roll this directly into `TeeLogStorageTest`.

I think a key test here is to create a mock `LogStorage` implementation that implements `writeReplace` and has special logic on the agent side (see for example https://github.com/jenkinsci/pipeline-cloudwatch-logs-plugin/blob/aadcfae6e78de1c96ba564c013a61b2eecc43d90/src/main/java/io/jenkins/plugins/pipeline_cloudwatch_logs/CloudWatchSender.java#L99). Then write a test to make sure that `TeeLogStorage` handles this case correctly, regardless of whether the implementation that does special stuff on agents is a primary or secondary.

dwnusbaum:
Nice! I've spent a bit of time looking at this and thinking about possible improvements/simplifications. Here are some initial thoughts.

As context for any other readers, `opentelemetry` has already implemented a custom variant of this kind of idea, where you can effectively choose `FileLogStorage` as the "primary" and have writes be mirrored to OTel. We would like whatever we do to allow that custom logic to be deleted from the OTel plugin in favor of this new system. For CloudBees Pipeline Explorer (CPE) at CloudBees we want support for a mode where CPE is the "primary" and writes are optionally mirrored to OTel.

Some general thoughts about this implementation:
* Adding `getPrimary` and `getSecondaries` to `LogStorage` itself seems undesirable. For CPE we need the former since the CPE implementation has custom read methods, but maybe it would be better to just have CPE handle `TeeLogStorage` directly and only have the methods in that class?
    * (I thought about having `TeeLogStorage` be an abstract class that all of the implementations would implement to avoid this kind of problem, but that gets a bit messy at instance construction time.) 
* The UI data binding for the config seems unusual, e.g. the use of `Describable` and `Descriptor` implies that the factories are configurable, but the factories are currently all just `@Extension` with configuration handled in separate objects. Maybe we should change the new interface to `public abstract class TeeLogStorageFactory extends Descriptor<TeeLogStorageFactory> implements LogStorage, Describable<TeeLogStorageFactory>` so that there is just one `@Extension` for each impl? Or instead of having the config fields be `TeeLogStorageFactory` instances, we could have the fields be class names instead, and use `f:select` as the form control, to make it clearer that the configuration must happen elsewhere? If we do that, maybe we don't need `TeeLogStorageFactory`, and we can just add a new `LogStorageFactory.getDisplayName` method with a default implementation that returns `getClass().getSimpleName()`, so that we don't need to change any existing callers?
* I don't know if we really need to support a list of secondaries. I think just supporting a single secondary would be enough to start with since it covers the existing use cases.
* I think we should introduce a factor for `FileLogStorage` to support the existing OTel use case.

As a side note, `LogStorageFactory` is `@Restricted(Beta.class)`, and only has 4ish known implementations (`opentelemetry` and `pipeline-cloudwatch-logs` going by https://www.jenkins.io/doc/developer/extensions/workflow-api/#logstoragefactory, CloudBees Pipeline Explorer (CPE) as a proprietary implementation, and then `FileLogStorage` as the default implementation even though it has no factory). So I think we can consider API changes that require us to update all of the existing implementations if it makes things any better/simpler in the long run, as long as it can be handled in a way that is backwards compatible for existing users' configurations (e.g. we don't want make changes that force existing OTel/CPE users to reconfigure things to preserve functionality).

Here are some things about the existing `LogStorage` and `LogStorageFactory` APIs that might be worth changing if we do make any bigger changes:
* The fall-through behavior trying multiple `LogStorageFactory` extensions does not really seem to make sense with the existing implementations. All of them expect to either be fully disabled or fully enabled, with the caveat that OTel and CPE have logic to ensure that builds which used `FileLogStorage` before the plugins were configured continue to use `FileLogStorage` even after the plugins are configured.
* All of the factories apart from `FileLogStorage` are configurable, the configuration is just done separately from the factory. Maybe we should introduce a `GlobalConfiguration` for Pipeline logging in general, and change `LogStorageFactory` to be `Describable`, and then all of the logging-related configuration would happen in one place, and users could explicitly select which one they want to use rather than having `ordinal` pick the winner. Doing this without breaking users's configurations may be tricky though.

src/test/java/org/jenkinsci/plugins/workflow/log/tee/TeeLogStorageTestBase.java
jgreffe: Maybe I misunderstand, but the current `TeeLogStorageTestBase` extends `LogStorageTestBase` which does [remoting test](https://github.com/jgreffe/workflow-api-plugin/blob/7993673de9ace7fb74228fd52af033eb2fa5ee18/src/test/java/org/jenkinsci/plugins/workflow/log/LogStorageTestBase.java#L188).
Doesn't it cover the case already?

dwnusbaum: I'm not sure that it covers the case in question. I'll take a closer look though.

jglick:
> just have CPE handle `TeeLogStorage` directly

If possible, that makes sense to me, assuming there is no known use case in OSS Jenkins.

jglick:
> we can consider API changes that require us to update all of the existing implementations if it makes things any better/simpler in the long run

Indeed. I am the maintainer of `pipeline-cloudwatch-logs` so happy to merge any required PRs that pass existing tests (a bit tricky to run since you need an AWS account and they are very slow).

src/main/java/org/jenkinsci/plugins/workflow/log/OutputStreamTaskListener.java
```diff
@@ -73,7 +75,9 @@ public interface OutputStreamTaskListener extends TaskListener {
                 listener.getClass() == StreamTaskListener.class ||
                 listener.getClass() == LogTaskListener.class ||
                 listener.getClass() == StreamBuildListener.class ||
-                listener.getClass() == BuildListenerAdapter.class;
+                listener.getClass() == BuildListenerAdapter.class ||
+                listener.getClass() == TeeTaskListener.class;
```
jglick, line 79: This does not make any sense as written. The point of this list is to enumerate classes which are defined in Jenkins core and thus cannot be fixed to be implemented in a way that uses an interface defined in this plugin. That obviously does not apply to `TeeTaskListener` which is implemented in this plugin.

src/main/java/org/jenkinsci/plugins/workflow/configuration/TeeLogStorageFactoryConfiguration.java
```diff
@@ -0,0 +1,95 @@
...
+    //        List<TeeLogStorageFactory> result = new ArrayList<>();
+    //        List<TeeLogStorageFactory> all = ExtensionList.lookup(TeeLogStorageFactory.class);
+    //        for(String factory : factories) {
+    //            all.stream().filter(f -> f.getId().equals(factory)).findFirst().ifPresent(result::add);
+    //        }
+    //        return result;
+    //    }
+
+    @Extension
+    public static class TeeLogStorgeFactoryConfigurationFilter extends DescriptorVisibilityFilter {
```
jglick, line 85:
```suggestion
    public static class TeeLogStorageFactoryConfigurationFilter extends DescriptorVisibilityFilter {
```

src/main/java/org/jenkinsci/plugins/workflow/configuration/TeeLogStorageFactoryConfiguration.java
```diff
@@ -0,0 +1,95 @@
...
+        this.enabled = enabled;
+    }
+
+    @Override
+    public boolean configure(StaplerRequest2 req, JSONObject json) throws FormException {
+        // We have to null out providers before data binding to allow all providers to be deleted in the config UI.
+        // We use a BulkChange to avoid double saves in other cases.
+        try (BulkChange bc = new BulkChange(this)) {
+            factories = null;
+            req.bindJSON(this, json);
```
jglick, line 59: IIRC you should just be calling the super method.

src/test/java/org/jenkinsci/plugins/workflow/log/tee/TeeLogStorageTestBase.java
dwnusbaum: FTR the case I was interested in was for a log storage like this, where the stream gets serialized over remoting.

<details>
<summary>Example LogStorage that has different behavior for logs written from agents</summary>

```

public static class RemoteUppercasingFileLogStorage implements LogStorage {
    private final File log;
    private final OutputStream out;

    public RemoteUppercasingFileLogStorage(File log) throws IOException {
        this.log = log;
        this.out = new FileOutputStream(log, true);
    }

    @Override
    public BuildListener overallListener() throws IOException, InterruptedException {
        return new MyListener(new Writer(null));
    }

    @Override
    public TaskListener nodeListener(FlowNode node) throws IOException, InterruptedException {
        // TODO: Does not actually handle step logs differently.
        return new MyListener(new Writer(node.getId()));
    }

    private static final class MyListener extends OutputStreamTaskListener.Default implements BuildListener, Closeable {
        private static final long serialVersionUID = 1;
        private final OutputStream out;
        
        public MyListener(OutputStream out) {
            this.out = out;
        }

        @Override
        public OutputStream getOutputStream() {
            return out;
        }

        @Override
        public void close() throws IOException {
            out.close();
        }

        private Object writeReplace() throws IOException {
            return new MyListener(new UppercaseWriter(out));
        }
    }

    private final class Writer extends OutputStream implements SerializableOnlyOverRemoting {
        private final String node;

        public Writer(String node) {
            this.node = node;
        }

        @Override
        public void write(int b) throws IOException {
            out.write(b);
        }

        @Override
        public void write(byte[] b, int off, int len) throws IOException {
            out.write(b, off, len);
        }

        @Override public void flush() throws IOException {
            out.flush();
        }

        @Override public void close() throws IOException {
            if (node == null) {
                out.close();
            }
        }
    }

    private static final class UppercaseWriter extends LineTransformationOutputStream implements SerializableOnlyOverRemoting {
        private static final long serialVersionUID = 1L;
        private final RemoteOutputStream out;
        public UppercaseWriter(OutputStream out) {
            this.out = new RemoteOutputStream(out);
        }

        @Override
        protected void eol(byte[] b, int len) throws IOException {
            String line = new String(b, 0, len, StandardCharsets.UTF_8);
            var uppercaseLine = line.toUpperCase(Locale.ENGLISH);
            var uppercaseBytes = uppercaseLine.getBytes(StandardCharsets.UTF_8);
            out.write(uppercaseBytes);
        }

        @Override public void flush() throws IOException {
            out.flush();
        }

        @Override public void close() throws IOException {
            out.close();
        }
    }

    @Override
    public AnnotatedLargeText<FlowExecutionOwner.Executable> overallLog(FlowExecutionOwner.Executable build, boolean complete) {
        return new AnnotatedLargeText(log, StandardCharsets.UTF_8, complete, build);
    }

    @Override
    public AnnotatedLargeText<FlowNode> stepLog(FlowNode node, boolean complete) {
        // TODO: Does not actually handle step logs differently.
        return new AnnotatedLargeText(log, StandardCharsets.UTF_8, complete, node);
    }
}
```
</details>

What I wanted to make sure of was that `TeeBuildListener` would not interfere with `writeReplace` being called on inner listeners when sent over the remoting channel. Everything seems to work fine though.

src/main/java/org/jenkinsci/plugins/workflow/configuration/TeeLogStorageFactoryWrapper.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import net.sf.json.JSONObject;
+import org.jenkinsci.plugins.workflow.log.tee.TeeLogStorageFactory;
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.StaplerRequest2;
+
+/**
+ * Wrapper of {@link org.jenkinsci.plugins.workflow.log.tee.TeeLogStorageFactory} for UI selection.
+ * It avoids having to declare a {@code @DataBoundConstructor} and {@code Describable} on implementations of TeeLogStorageFactory.
```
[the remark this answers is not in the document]
jgreffe: Yes, I still wanted to check if using an `f:hetero-list` without `Describable` for each `TeeLogStorageFactory` implementation was feasible. I liked the `f:hetero-list` UI, with a button allowing to add filtered factories and drag'n'drop them to change the order.

Anyway, I guess I'll change the PR to have only one primary and one secondary in the UI, but still allow in the code to define multiple secondaries (?)

src/main/java/org/jenkinsci/plugins/workflow/flow/GlobalDefaultFlowDurabilityLevel.java
```diff
@@ -103,7 +103,11 @@ public Permission getRequiredGlobalConfigPagePermission() {
     public static FlowDurabilityHint getDefaultDurabilityHint() {
         Jenkins j = Jenkins.getInstanceOrNull();
         if (j != null) {
-            FlowDurabilityHint hint = j.getExtensionList(DescriptorImpl.class).get(0).durabilityHint;
+            List<DescriptorImpl> descriptor = j.getExtensionList(DescriptorImpl.class);
+            if (descriptor.isEmpty()) {
```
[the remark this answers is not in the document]
jgreffe: Simply reverted, see https://github.com/jenkinsci/workflow-api-plugin/pull/406/commits/d226c01650584a0bb88ac7250181a60f79aa4825

src/main/java/org/jenkinsci/plugins/workflow/log/OutputStreamTaskListener.java
```diff
@@ -63,7 +65,7 @@ public interface OutputStreamTaskListener extends TaskListener {
             return ((OutputStreamTaskListener) listener).getOutputStream();
         }
         PrintStream ps = listener.getLogger();
-        if (ps.getClass() != PrintStream.class) {
+        if (ps.getClass() != PrintStream.class && ps.getClass() != TeePrintStream.class) {
             Logger.getLogger(OutputStreamTaskListener.class.getName()).warning(() -> "Unexpected PrintStream subclass " + ps.getClass().getName() + " which might override write(…); error handling is degraded unless OutputStreamTaskListener is used: " + listener.getClass().getName());
             return ps;
         }
```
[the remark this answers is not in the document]
jgreffe: Simply reverted, see https://github.com/jenkinsci/workflow-api-plugin/pull/406/commits/d226c01650584a0bb88ac7250181a60f79aa4825

src/test/java/org/jenkinsci/plugins/workflow/log/tee/TeeLogStorageTestBase.java
jgreffe: Added this case to cover the specific serialization over remoting: https://github.com/jenkinsci/workflow-api-plugin/pull/406/commits/44bd04affcf8dd263e1c8ea0e50c68249e15dab2

src/main/java/org/jenkinsci/plugins/workflow/configuration/TeeLogStorageFactoryWrapper.java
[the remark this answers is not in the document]
jgreffe: Please see https://github.com/jenkinsci/workflow-api-plugin/pull/406/commits/c22740e07b74ae770047e17dec1eb08b7467fa1b

jgreffe:
PoC phase is done, let's close this PR.

PR #414, opened 2025-07-11
Replace `docker-fixtures` with `testcontainers`

strangelookingnerd:
Replace `docker-fixtures` with `testcontainers`  (see https://github.com/jenkinsci/docker-fixtures/pull/122#issuecomment-3046351724)

This change removes already deprecated API methods from the test jar.

### Testing done

`mvn clean verfiy`

### Submitter checklist

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
@@ -58,7 +58,6 @@ import hudson.util.StreamTaskListener;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
-import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Collections;
@@ -74,14 +73,14 @@ import jenkins.security.MasterToSlaveCallable;
 import jenkins.util.VirtualFile;
 import org.apache.commons.io.IOUtils;
 import org.jenkinsci.plugins.workflow.flow.StashManager;
-import org.jenkinsci.test.acceptance.docker.Docker;
-import org.jenkinsci.test.acceptance.docker.DockerImage;
-import org.jenkinsci.test.acceptance.docker.fixtures.JavaContainer;
 import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
 import org.jvnet.hudson.test.JenkinsRule;
 import org.jvnet.hudson.test.LoggerRule;
+import org.testcontainers.DockerClientFactory;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.images.builder.ImageFromDockerfile;
 
 /**
  * {@link #artifactArchiveAndDelete} and variants allow an implementation of {@link ArtifactManager} plus {@link VirtualFile} to be run through a standard gantlet of tests.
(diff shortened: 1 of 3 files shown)


basil (changes requested):
This PR is introduced technical debt, as the `Dockerfile` is duplicated between https://github.com/jenkinsci/workflow-api-plugin/pull/414 and https://github.com/jenkinsci/docker-fixtures/pull/122, which creates a maintenance burden. See [this page](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) for more information about why this is the case. I recommend refactoring the common images that used to be present in `docker-fixtures` into a new Jenkins repository for TestContainers, something like https://github.com/InformaticsMatters/docker-keycloak.

jglick:
I have thought of that but have mixed feelings. On the one hand, there would be less duplication. On the other hand, it makes it harder to develop and maintain tests because the repo is not self-contained: if you wish to adjust even a minor aspect of the Docker environment, you have to open a pull request on an external repo, spend time thinking about whether it is backward compatible, find someone to merge it, wait for it to be published, and only then proceed with your code changes.

Another benefit of a published image is that tests would also run faster on fresh CI machines since they would only need to download an image, not build one in steps. This can be solved by using `--cache-from` / `--cache-to` but requires additional infrastructure on the CI server for authentication, which may be impractical in @jenkinsci where plugin developers are not really trusted (every repo would need its own cache location and corresponding write credentials). Note that we already pay this cost with `docker-fixtures`, so switching to Testcontainers with `ImageFromDockerfile` does not make anything worse.

src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
```diff
@@ -119,14 +115,13 @@ private static void wrapInContainer(@NonNull JenkinsRule r, @CheckForNull Artifa
...
         try {
             DumbSlave agent;
-            if (image != null) {
-                runningContainer = image.start(JavaContainer.class).start();
+            if (container != null) {
+                container.start();
                 StandardUsernameCredentials creds = new UsernamePasswordCredentialsImpl(CredentialsScope.SYSTEM, "test", "desc", "test", "test");
                 CredentialsProvider.lookupStores(Jenkins.get()).iterator().next().addCredentials(Domain.global(), creds);
-                agent = new DumbSlave("test-agent", "/home/test/slave", new SSHLauncher(runningContainer.ipBound(22), runningContainer.port(22), "test"));
+                agent = new DumbSlave("test-agent", "/home/test/slave", new SSHLauncher(container.getHost(), container.getMappedPort(22), "test"));
```
jglick, line 124: I actually recommend switching from this container to a standard one. https://hub.docker.com/r/jenkins/ssh-agent could probably be used, just passing a predefined `JENKINS_AGENT_SSH_PUBKEY`. It may be easier to switch to an inbound https://hub.docker.com/r/jenkins/agent though. https://github.com/jenkinsci/workflow-cps-plugin/blob/b2a1d7b29c6fdd0161dbb1c7586d579d49bbe1e2/plugin/src/test/java/org/jenkinsci/plugins/workflow/cps/SerialFormTest.java#L80-L85 works (though the requirement to `exposeHostPorts` is admittedly ugly); note that that test is more complex than you actually need for most purposes specifically because it is testing upgrade from `@LocalData`. I swear I wrote a simpler version somewhere recently, but I cannot find it now.

jglick: For context, there are three common reasons to use containers in a Jenkins test:
* To run some random service, like LDAP.
* To run an agent with specific tools installed that it is expected to call, like a particular version of an SCM.
* To run an agent but where the test does not care anything about the launcher or the details of the container image or even whether the agent runs in a container at all—merely that it _not_ run in the same OS namespace as the controller (and test code), otherwise you could just use `JenkinsRule.createAgent`.

This test falls into the last category. The idea is to catch mistakes in remoted callables which might be presuming that a path valid on the controller is also valid on an agent or vice-versa. So we can use any inbound or outbound agent image from the past few years (just needs Java 17+ obviously) and it would be fine.

jglick: It is an option of course, but we would anyway want to throw out any such images sooner or later and just use the standard `jenkins/{,ssh-}agent` images for these cases of the third type. Since the existing `docker-fixtures` release is working fine as it is, I do not see any urgency in an intermediate fix: we may as well write tests the straightforward way to begin with. I can check feasibility in this PR next week.

src/test/java/org/jenkinsci/plugins/workflow/ArtifactManagerTest.java
```diff
@@ -90,21 +89,18 @@ public class ArtifactManagerTest {
...
 
-    /**
-     * @deprecated Not actually used externally.
-     */
-    @Deprecated
-    public static @CheckForNull DockerImage prepareImage() throws Exception {
-        Docker docker = new Docker();
-        if (!Functions.isWindows() && docker.isAvailable()) { // TODO: Windows agents on ci.jenkins.io have Docker, but cannot build the image.
-            return docker.build(JavaContainer.class);
+    public static @CheckForNull GenericContainer<?> prepareImage() {
```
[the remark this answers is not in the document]
jglick: I see, the comment and deprecation were in fact wrong.

PR #423, opened 2025-08-09
Avoid including unrelated content in step log after parse errors

das7pad:
While working on another patch, I noticed this subtle bug in the error handling of the per-step log reading. Let me know if you want me to create a Jira ticket first.

Previously, when not resetting the position on Long parse error or non-positive index change, the position of the next transition was mistaken for the end of the current section and the entire content since the last transition was emitted.

### Testing done

The added tests are failing without this patch: The logs of step 1 contained the "between1\n" line -> "1\nbetween1\n". The atEnd tests would contain the "after" part.

### Submitter checklist

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/log/FileLogStorageTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/log/FileLogStorageTest.java
@@ -28,6 +28,8 @@ import static org.junit.Assert.assertTrue;
 
 import hudson.model.TaskListener;
 import java.io.File;
+import java.nio.charset.StandardCharsets;
+import org.apache.commons.io.FileUtils;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/log/FileLogStorageTest.java
```diff
@@ -55,6 +57,48 @@ public class FileLogStorageTest extends LogStorageTestBase {
         assertOverallLog(0, lines("stuff"), true);
     }
 
+    @Test public void corruptIndex() throws Exception {
+        FileUtils.writeStringToFile(log, "before\n1\nbetween1\n2\nbetween2\n3\nafter", StandardCharsets.UTF_8);
+        FileUtils.writeStringToFile(new File(log + "-index"), "7 1\n?\n18 2\n20\n29 3\n31", StandardCharsets.UTF_8);
```
jglick, line 62: Multi-line strings might be more legible here, though you would need to careful about placement of newlines and indentation.

jglick (approved):
Not sure what you would need to do for this code to be active, but sounds right. Only dimly recall this.

dwnusbaum (approved):

PR #437, opened 2025-09-30
Disallow soft references in `MemoryAssert`-related tests in `FlowExecutionListTest`

dwnusbaum:
While looking for places to test https://github.com/jenkinsci/jenkins-test-harness/pull/1029, I noticed that this class was allowing soft references while using `MemoryAssert.assertGC`, so I tracked them down and cleaned them out so the test can run with soft references disallowed.

### Testing done

See updated tests.

### Submitter checklist

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
@@ -47,6 +47,7 @@ import java.util.Objects;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.LogRecord;
@@ -203,6 +204,8 @@ public class FlowExecutionListTest {
     }
 
     @Test public void stepExecutionIteratorDoesNotLeakBuildsWhenCpsVmIsStuck() throws Throwable {
+        // Make sure ForkJoinPool is not initialized on a thread where its inheritedAccessControlContext would point to a CpsGroovyShell$CleanGroovyClassLoader
+        ForkJoinPool.commonPool().execute(() -> {});
         sessions.then(r -> {
             var notStuck = r.createProject(WorkflowJob.class, "not-stuck");
             notStuck.setDefinition(new CpsFlowDefinition("semaphore 'wait'", true));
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
```diff
@@ -203,6 +204,8 @@ public class FlowExecutionListTest {
     }
 
     @Test public void stepExecutionIteratorDoesNotLeakBuildsWhenCpsVmIsStuck() throws Throwable {
+        // Make sure ForkJoinPool is not initialized on a thread where its inheritedAccessControlContext would point to a CpsGroovyShell$CleanGroovyClassLoader
+        ForkJoinPool.commonPool().execute(() -> {});
```
dwnusbaum, line 208: It's a bit interesting that I keep running into these cases where first-time initialization of various objects happening in the context of a Pipeline build causes problems. In practice it's probably ok, and at worst they would only cause 1-build leaks, but I wonder how many such leaks exist in practice.

Compare https://github.com/jenkinsci/pipeline-groovy-lib-plugin/pull/199#discussion_r2377121804.

jglick (approved):

src/test/java/org/jenkinsci/plugins/workflow/flow/FlowExecutionListTest.java
dwnusbaum: Maybe, but this is the common pool, which seems to be unconfigurable, and the reference path was not via `Thread.contextClassLoader` as is typical (`AppClassLoader` in this case), but `Thread.inheritedAccessControlContext`, which holds a reference to `CpsGroovyShell$CleanGroovyClassLoader` via a `ProtectionDomain.classloader`, constructed I think via `AccessController.getContext()`.

So whatever the exact problem is, it will go away once we pick up [JEP 486](https://openjdk.org/jeps/486), which deletes `Thread.inheritedAccessControlContext`.

FWIW also I think the critical usage path of the common pool is via the default executor in `Caffeine` caches, specifically two caches in `SimpleXStreamFlowNodeStorage` and one in `EnumeratingWhitelist`. We could perhaps set their executors to `Computer.threadPoolForRemoting` or some manually-constructed thread pool to avoid the issue in the meantime.

dwnusbaum: The notable thing in these cases is that we are not obviously using any async-related methods or types with the Caffeine caches. We only use `weakKeys`, `strongValues`, and `LoadingCache`, so the fact that an async task is being submitted to the common pool is not immediately obvious. Perhaps it has to do with the internal maintenance referred to in [the Javadoc](https://www.javadoc.io/static/com.github.ben-manes.caffeine/caffeine/3.2.2/com.github.benmanes.caffeine/com/github/benmanes/caffeine/cache/Caffeine.html#executor(java.util.concurrent.Executor)).

PR #528, opened 2026-08-24
Add FlowNodeConsoleUrlProvider interface for console URLs

das7pad:
- Part of https://github.com/jenkinsci/pipeline-graph-view-plugin/issues/1266

This interface provides a method to retrieve console URLs for specific FlowNodes in Jenkins, allowing for richer visualizations in the Pipeline Graph View.

### Testing done

Implement `public String getConsoleUrl(Run<?, ?> run, FlowNode node)` in `PipelineConsoleViewUrlProvider` and adopt that.

See logs URLs on GitHub checks point at the step.

### Submitter checklist


dwnusbaum (changes requested):
@das7pad see https://github.com/jenkinsci/pipeline-graph-view-plugin/issues/1266#issuecomment-5396372599, `ConsoleUrlProvider` is for internal Jenkins-to-Jenkins links, the case you care about is an external link, so it should be a new API method in https://github.com/jenkinsci/display-url-api-plugin instead.

The general idea makes sense though and I think it will be helpful

